PR #364, opened 2021-07-06
feat: Jest resolver that supports out-of-tree platforms

tido64:


packages/jest-resolver/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "lint": "rnx-kit-scripts lint",
+    "test": "rnx-kit-scripts test"
+  },
+  "dependencies": {
+    "pkg-dir": "^5.0.0",
+    "resolve": "^1.20.0"
+  },
+  "peerDependencies": {
+    "@react-native-community/cli": ">=4.10",
+    "react-native": ">=0.63"
```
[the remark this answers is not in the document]
NickGerleman:
```suggestion
    "react-native": ">=0.63 || <0.0.0"
```

(to support nightlies without warnings)

packages/jest-resolver/src/index.ts
```diff
@@ -0,0 +1,121 @@
...
+function initialize(): [string[] | undefined, Opts["pathFilter"]] {
+  const targetPlatform = getTargetPlatform();
+  if (!targetPlatform) {
+    return [undefined, (_pkg, _path, relativePath) => relativePath];
+  }
+
+  const [platformName, platformPath] = targetPlatform;
+  const reactNativePath = path.dirname(
+    require.resolve("react-native/package.json")
+  );
```
NickGerleman, line 95: nit: this could resolve the wrong version in multi-rn repos. Consider adding the root dir to `require.resolve` options to search for the RN dependency used by the pacakge running the tests.

packages/jest-resolver/src/index.ts
```diff
@@ -0,0 +1,140 @@
+import loadConfig from "@react-native-community/cli/build/tools/config";
```
afoxman, line 1: We have `loadMetroConfig` in metro-service to wrap this and give it proper types.

packages/jest-resolver/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  // TODO: Should probably read Jest config
+  const { moduleFileExtensions } = defaults;
+  if (!moduleFileExtensions || moduleFileExtensions.length === 0) {
+    return undefined;
+  }
+
+  return [
+    ...moduleFileExtensions.map((ext) => `.${targetPlatform}.${ext}`),
```
afoxman, line 25: For `win` you'll need to include expansion to win32 and windows, too.

packages/jest-resolver/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+    return [names[0], rootDir];
+  }
+
+  console.warn("No platforms found; picking a random one");
+  return [undefined, undefined];
+}
+
+function getTargetPlatform(): PlatformPath {
+  // TODO: Figure out the mechanism for providing a target platform.
+  const targetPlatform = process.env["RN_TARGET_PLATFORM"];
```
afoxman, line 76: This is why I didn't add Jest platform support early on :) There is some code that works in sdx-platform which you should look at -- Coates wrote it. They support it via some build tasks which run jest per-platform and specify on the command line (I think).

afoxman: I realize command-line args are beyond the scope of this preset/plugin, but you may need to provide a larger solution (e.g. multiple packages in rnx-kit that must work together). Or, some form of extensibility for external packages to provide a function which returns the current platform.

tido64: I've filed #425 for further investigation.

packages/jest-resolver/src/index.ts
tido64: Calling `@react-native-community/cli/build/tools/config` is equivalent to invoking `react-native config` on the command line. Unless I missed something, I don't think `loadMetroConfig` does the same thing?

afoxman: ooooh. CLI config, not metro config. Sorry for the noise.

afoxman (approved):

PR #422, opened 2021-07-21
fix(metro-service): support Metro <0.63

tido64:
### Description

This enables `metro-service` to work in FluentUI React Native (FURN), where Metro is still at 0.59.

### Test plan

1. Build `metro-service` locally
2. Clone FURN: `yarn clone https://github.com/microsoft/fluentui-react-native.git`
3. Bump `@rnx-kit/cli` to 0.5.36 in FURN
4. Replace `FURN/node_modules/@rnx-kit/metro-service` with `rnx-kit/packages/metro-service`
5. Build FURN: `yarn build`
6. Bundle: `yarn bundle`

### DIFF
--- a/packages/metro-service/src/config.ts
+++ b/packages/metro-service/src/config.ts
@@ -48,6 +48,15 @@ function reactNativePlatformResolver(platformImplementations: {
   };
 }
 
+function getAsyncRequireModulePath(): string | undefined {
+  try {
+    // `metro-runtime` was introduced in 0.63
+    return require.resolve("metro-runtime/src/modules/asyncRequire");
+  } catch (_) {
+    return require.resolve("metro/src/lib/bundle-modules/asyncRequire");
+  }
+}
+
 function getDefaultConfig(cliConfig: CLIConfig): InputConfigT {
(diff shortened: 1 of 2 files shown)


packages/metro-service/src/config.ts
```diff
@@ -48,6 +48,15 @@ function reactNativePlatformResolver(platformImplementations: {
   };
 }
 
+function getAsyncRequireModulePath(): string | undefined {
+  try {
+    // `metro-runtime` was introduced in 0.63
+    return require.resolve("metro-runtime/src/modules/asyncRequire");
+  } catch (_) {
+    return require.resolve("metro/src/lib/bundle-modules/asyncRequire");
```
tido64, line 56: @afoxman: Even though this should work all the way down to 0.59, I don't know if we should change `package.json` to advertise that. What do you think?

afoxman: I don't want to advertise broad support for old versions of Metro. If we had to, down the road, we'd need a test suite that exercises the metro-service package in full, and then run that as we wind back metro package versions.

afoxman (approved):

PR #443, opened 2021-07-30
fix(third-party-notices): add JSON output format

tido64:
### Description

Adds JSON output format to `third-party-notices`.

### Test plan

Tests were added.

### DIFF
--- a/change/@rnx-kit-third-party-notices-2576042b-fc6d-49f0-94ed-2dbb1eeb2a53.json
+++ b/change/@rnx-kit-third-party-notices-2576042b-fc6d-49f0-94ed-2dbb1eeb2a53.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "minor",
   "comment": "Add JSON output format",
   "packageName": "@rnx-kit/third-party-notices",
   "email": "4123478+tido64@users.noreply.github.com",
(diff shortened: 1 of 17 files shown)


packages/cli/react-native.config.js
```diff
@@ -148,6 +148,12 @@ module.exports = {
           name: "--source-map-file <file>",
           description: "The sourceMap file to generate licence contents for.",
         },
+        {
+          name: "--json",
+          description: "Output license information as a JSON",
+          default: true,
```
afoxman, line 154: IIRC the default output is plain text... if I'm wrong, ignore this comment. 

Otherwise, this seems dangerous, as you're changing the defaults and callers probably won't know that their TPN file has changed formats as it tends to get generated and packaged/nuget-ed, rather than checked in.
```suggestion
          default: false,
```

packages/third-party-notices/src/cli.ts
```diff
@@ -24,6 +21,11 @@ function getArgs(): IWriteThirdPartyNoticesOptions {
       describe: "The sourceMap file to generate license contents for.",
       require: true,
     },
+    json: {
+      type: "boolean",
+      describe: "Output license information as a JSON",
+      default: false,
```
afoxman, line 27: So this disagrees with the default from our CLI. That seems wrong and confusing to people who might use this.

change/@rnx-kit-third-party-notices-2576042b-fc6d-49f0-94ed-2dbb1eeb2a53.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
afoxman, line 2: API is changing, but is backward compatible. This should be a minor bump.

packages/cli/react-native.config.js
tido64: Good catch! I made a copy/pasta mistake 😛

afoxman (approved):

PR #444, opened 2021-07-30
Update Metro and TypeScript service packages

afoxman:
### Description

Update Metro and TypeScript service packages, in preparation for adding a Metro server CLI command.

`metro-service`
* `loadMetroConfig()`: formalize overrides, add more of them
* add `startServer()` for launching a Metro server
* and `createTerminal()` as a factory for creating a Metro terminal and terminal-reporter, which are used to report messages while the server is running

`typecript-service`
* add an optional `write()` function to the diagnostic code, giving the caller control over how diagnostic messages are written

### Test plan

Run all repo tests and generate test-app bundle. Manually verify as part of a larger set of changes.


packages/typescript-service/src/diagnostics.ts
```diff
@@ -23,7 +25,12 @@ export function createDiagnosticWriter(): DiagnosticWriter {
   }
 
   function print(diagnostic: ts.Diagnostic | ts.Diagnostic[]) {
-    ts.sys.write(format(diagnostic));
+    const message = format(diagnostic);
+    if (write) {
+      write(message);
+    } else {
+      ts.sys.write(message);
+    }
```
tido64, line 33: Can we move the if-check outside the function so we don't have to check everytime we want to print?

packages/metro-service/src/config.ts
```diff
@@ -127,22 +127,43 @@ function getDefaultConfig(cliConfig: CLIConfig): InputConfigT {
...
-    config?: string;
-    maxWorkers?: number;
-    resetCache?: boolean;
-    reporter?: Reporter;
-  }
+  overrides: MetroConfigOverrides
 ): Promise<ConfigT> {
   const defaultConfig = getDefaultConfig(cliConfig);
+
+  //  apply overrides that loadConfig() doesn't do for us
```
tido64, line 157: Do we need to upstream this?

packages/metro-service/src/terminal.ts
```diff
@@ -0,0 +1,37 @@
...
+    // require('./reporter');
+    return require(path.resolve(customReporterPath));
+  }
+}
+
+export function createTerminal(customReporterPath?: string): MetroTerminal {
+  const terminal = new Terminal(process.stdout);
+
+  const TerminalReporterClass = getTerminalReporterClass(customReporterPath);
+  const reporter = new TerminalReporterClass(terminal);
```
tido64, line 34: Can we reuse the instance in [`defaults`](https://github.com/facebook/metro/blob/master/packages/metro-config/src/defaults/index.js#L140) if no custom reporter is set?

packages/metro-service/src/terminal.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (e.code !== "MODULE_NOT_FOUND") {
+      throw e;
+    }
+    // If that doesn't work, then we next try relative to the cwd, eg:
+    // require('./reporter');
+    return require(path.resolve(customReporterPath));
+  }
+}
+
+export function createTerminal(customReporterPath?: string): MetroTerminal {
```
tido64, line 30: Is `createTerminal` being used anywhere? I can't find it in this PR.

packages/metro-service/src/config.ts
afoxman: There are only a few overrides here, and it doesn't seem worth the trouble of ushering it through the whole big process.

packages/metro-service/src/terminal.ts
afoxman: That's a clever optimization. I'd rather the caller be a bit more explicit, only creating a custom terminal if needed (eg. customReporterPath === undefined).

Your next comment notes that the call site isn't in this PR. It's in one of my other queued up PRs. In that PR, I will update the caller to only create a new terminal when a custom reporter path is set.

afoxman:
@JasonVMo can you finish this PR review while Tommy is out? I've addressed all of his comments.

packages/@types/metro-core/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "private": "true",
+  "name": "@types/metro-core",
```
JasonVMo, line 3: I wonder if it is worth starting something on definitely typed for this? Not a blocker here but it seems useful (and like there will be other types to add over time)

JasonVMo (approved):

packages/@types/metro-core/package.json
afoxman: Tommy and I talked about it, and it's a long-term goal. Filed #473 to keep track of it.

PR #446, opened 2021-07-30
Refactoring and fixes in rnx-kit/cli

afoxman:
### Description

Refactor the code behind the `rnx-bundle` command. 
- Move kit config and metro config to separate files
- Add specific overrides to kit and metro config functions, rather than using the big `CLIBundleOptions` type

Updated failure cases
- Fail when no target platform is given on the command-line or in kit config
- Stop validating Metro config. This was meant to help users realize when they were touching fields that we will be touching, but we override so much at this point, it isn't workable. We can't always tell when a prop is set by default in Metro or by the user.

TypeScript support -- Metro hook function:
- Track one TS project per platform
- fix: incorrectly using modified list to run `delete` ops on the TS project

Cleanup
- Remove command-line props which have no effect.
- Change @rnx-kit dependencies to semver ranges.

### Test plan

Build and test across the repo. Bundle test app. Inspect in the debugger.


packages/cli/package.json
```diff
@@ -19,29 +19,32 @@
...
     "@rnx-kit/metro-serializer-esbuild": "^0.0.11",
-    "@rnx-kit/metro-service": "1.0.5",
-    "@rnx-kit/third-party-notices": "1.0.3",
-    "@rnx-kit/typescript-service": "1.1.1",
-    "chalk": "^4.1.0"
+    "@rnx-kit/metro-service": "^1.0.5",
+    "@rnx-kit/third-party-notices": "^1.0.3",
+    "@rnx-kit/typescript-service": "^1.1.1",
+    "chalk": "^4.1.0",
+    "readline": "^1.3.0"
```
tido64, line 32: Where are we using `readline`?

packages/cli/src/bundle.ts
```diff
@@ -203,12 +121,13 @@ export async function rnxBundle(
       const configFileName = tsservice.findProject(entryPath, "tsconfig.json");
       if (!configFileName) {
         console.warn(
-          "skipping TypeScript validation -- cannot find tsconfig.json for entry file %o",
+          chalk.yellow(
+            "skipping TypeScript validation -- cannot find tsconfig.json for entry file %o"
+          ),
           entryPath
         );
```
tido64, line 128: Should we use `@rnx-kit/console` for printing warnings? It doesn't have to be in this PR, of course.

packages/cli/package.json
afoxman: It's in the next PR. Sorry, I tried to keep them each completely separate, but I missed this bit. It's part of the CLI server code.

packages/cli/src/bundle.ts
afoxman: fixed

JasonVMo (approved):

PR #447, opened 2021-07-30
Add `rnx-start` command for running a Metro bundle server

afoxman:
### Description

Add `rnx-start` command for running a Metro bundle server. Update documentation. Add supporting kit config.

### Test plan

Repo build, test and bundle (to make sure it still works). Manual run on the dev server and inspection in the debugger. Verified using test-app on iOS.


packages/cli/README.md
```diff
@@ -6,6 +6,7 @@
 Command-line interface for working with `kit` packages in your repo.
 
 - [Bundle a kit](#Bundle-a-Kit)
+- [Start a Bundle server](#Start-a-Bundle-Server)
```
tido64, line 9:
```suggestion
- [Start a bundle server](#Start-a-Bundle-Server)
```

packages/cli/src/start.ts
```diff
@@ -0,0 +1,217 @@
...
+): Promise<void> {
+  const serverConfig = getKitServerConfig(cliOptions);
+  if (!serverConfig) {
+    return Promise.resolve();
+  }
+
+  // interactive mode requires raw access to stdin
+  let interactive = cliOptions.interactive;
+  if (interactive) {
+    interactive = Boolean(process.stdin.setRawMode);
```
tido64, line 52: If we're checking whether a text terminal is attached, we should probably use `isTTY` instead:
```suggestion
    interactive = Boolean(process.stdin.isTTY);
```

packages/config/src/serverConfig.ts
```diff
@@ -0,0 +1,28 @@
...
+  assetPlugins?: string[];
+
+  /**
+   * Additional source-file extensions to include when generating bundles. Comma-separated
+   * list, excluding the leading dot.
+   */
+  sourceExts?: string[];
+};
+
+export type ServerConfig = ServerParameters;
```
tido64, line 28: Why are these separate from `BundlerRuntimeParameters`? Does Metro also split them this way?

JasonVMo (approved):

PR #455, opened 2021-08-03
feat(dep-check): add command for setting react-native version

tido64:
### Description

Adds a command for easily setting `reactNativeVersion` and `reactNativeDevVersion` consistently across a repo.

### Test plan

1. Prepare: 
   ```sh
   yarn
   cd packages/test-app
   yarn build --dependencies

   # If you're not on Windows, you need to make the script executable:
   chmod +x ../dep-check/lib/cli.js
   ```
2. Test the prompt: `yarn rnx-dep-check --set-version`
   - On the first question, at least one version must be picked
   - When several versions are picked, the second prompt should only include the selection
   - If only one version was picked, there is no second prompt
3. Test providing values:
   - `yarn rnx-dep-check --set-version 0.63`: should set both `reactNativeVersion` and `reactNativeDevVersion` to 0.63, and also change the `react-native` version to `^0.63.2`
   - `yarn rnx-dep-check --set-version 0.64,0.63`: should set `reactNativeVersion: "^0.63 || ^0.64"` and `reactNativeDevVersion: ^0.64`
4. Repeat steps 2 and 3 using `yarn react-native rnx-dep-check` instead

### DIFF
--- a/packages/cli/react-native.config.js
+++ b/packages/cli/react-native.config.js
@@ -5,7 +5,7 @@ const {
   parseBoolean,
   rnxBundle,
   rnxStart,
-  rnxDepCheck,
+  rnxDepCheckCommand,
   rnxTestCommand,
   rnxWriteThirdPartyNotices,
 } = require("./lib/index");
@@ -181,36 +181,7 @@ module.exports = {
         },
       ],
     },
-    {
-      name: "rnx-dep-check",
-      description: "Dependency checker for React Native apps",
-      func: rnxDepCheck,
-      options: [
-        {
-          name: "--custom-profiles [module]",
-          description:
-            "Path to custom profiles. This can be a path to a JSON file, a `.js` file, or a module name.",
-        },
-        {
-          name: "--exclude-packages [packages]",
-          description:
-            "Comma-separated list of package names to exclude from inspection.",
-        },
-        {
-          name: "--init [app|library]",
-          description: "Writes an initial kit config",
-        },
-        {
-          name: "--vigilant [versions]",
-          description:
-            "Inspects packages regardless of whether they've been configured. Specify a comma-separated list of profile versions to compare against, e.g. `0.63,0.64`. The first number specifies the target version.",
-        },
-        {
-          name: "--write",
-          description: "Writes all changes to the specified `package.json`",
-        },
-      ],
-    },
+    rnxDepCheckCommand,
     rnxTestCommand,
     {
       name: "rnx-write-third-party-notices",
--- a/packages/dep-check/src/cli.ts
+++ b/packages/dep-check/src/cli.ts
@@ -107,9 +107,9 @@ async function makeCommand(args: Args): Promise<Command | undefined> {
     return makeInitializeCommand(init);
   }
 
-  // When `--set-version` is without a value, `versions` is an empty string if
+  // When `--set-version` is without a value, `setVersion` is an empty string if
   // invoked directly. When invoked via `@react-native-community/cli`,
-  // `versions` is `true` instead.
+  // `setVersion` is `true` instead.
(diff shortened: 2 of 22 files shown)


packages/cli/react-native.config.js
```diff
@@ -103,36 +103,7 @@ module.exports = {
         },
       ],
     },
-    {
-      name: "rnx-dep-check",
```
afoxman, line 107: The name `dep-check` is getting stretched awfully thin at this point... maybe a more general name since it can do so much to handle dependencies?  Or several commands?

packages/dep-check/src/cli.ts
```diff
@@ -107,6 +107,13 @@ function makeCommand(args: Args): Command | undefined {
     return makeInitializeCommand(init);
   }
 
+  // When `--set-version` is without a value, `versions` is an empty string if
```
afoxman, line 110:
```suggestion
  // When `--set-version` is without a value, `setVersion` is an empty string if
```

packages/dep-check/src/cli.ts
```diff
@@ -107,6 +107,13 @@ function makeCommand(args: Args): Command | undefined {
...
+  // invoked directly. When invoked via `@react-native-community/cli`,
+  // `versions` is `true` instead.
```
afoxman, line 112:
```suggestion
  // `setVersion` is `true` instead.
```

packages/dep-check/test/setVersion.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      "rnx-kit": {
+        ...mockManifest["rnx-kit"],
+        reactNativeVersion: "^0.63.0 || ^0.64.0",
+        reactNativeDevVersion: "^0.64.0",
+      },
+    });
+  });
+
+  test("prompts the user if no version is specified", async () => {
+    const mockManifest = {
```
afoxman, line 66: This pre-amble / setup seems a lot like what is in the previous test. Can they be consolidated and shared?

afoxman: Same with any test(s) below

packages/cli/react-native.config.js
kelset: that's a good point, I've opened an issue about this https://github.com/microsoft/rnx-kit/issues/484

kelset:
I cloned it locally and I just want to say that the CLI experience of `yarn rnx-dep-check --set-version` is SO GOOD. I love it.

packages/dep-check/src/cli.ts
```diff
@@ -173,6 +183,12 @@ if (require.main === module) {
         choices: ["app", "library"],
         conflicts: ["vigilant"],
       },
+      "set-version": {
+        description:
+          "Sets `reactNativeVersion` and `reactNativeDevVersion` for any configured package. The value should be a comma-separated list of `react-native` versions to set. The first number specifies the development version. Example: `0.64,0.63`",
```
kelset, line 188:
```suggestion
          "Sets `reactNativeVersion` and `reactNativeDevVersion` for any configured package. Has an interactive menu if value is not specified. The value should be a comma-separated list of `react-native` versions to set (first number specifies the development version). Example: `0.64,0.63`",
```

kelset (approved):
overall I agree with Adam's feedbacks; aside from that, looks solid.

Personally, I would have split some of the refactoring(s) (ex. the helper and json files) to a separate PR but 👍

packages/cli/react-native.config.js
tido64: Thanks for starting the conversation, @kelset. I will resolve this for now since we don't want to change the name before RNEU.

PR #475, opened 2021-08-09
Adding esbuild plugin for mapping lib/ imports to src/

sverrejoh:
This is similar to babel-plugin-import-path-remapper.

Having a monorepo of local packages that compiles from typescript source code in
`src/` to distributed javascript in `lib/`, this plugin rewrites all imports
from pointing to the generated artifacts to point directly to the source. This
makes it possible to consume local packages without first having to compile
them.


packages/esbuild-plugin-import-path-remapper/README.md
```diff
@@ -0,0 +1,26 @@
...
+For example, to remap all paths under the `@rnx-kit` scope:
+
+```js
+// esbuild.js
+require("esbuild")
+  .build({
+    entryPoints: ["app.tsx"],
+    bundle: true,
+    outfile: "out.js",
+    plugins: [ImportPathRemapperPlugin("@rnx-kit")],
```
afoxman, line 23: Are you missing a require statement to define `ImportPathRemapperPlugin`?

afoxman (commented):
The code looks good to me, but I haven't done any work on esbuild. Tommy should review once he gets back from vacation.

packages/esbuild-plugin-import-path-remapper/README.md
sverrejoh: Yes, you’re right. I should add that. Thank you!

sverrejoh: I added the require like you suggested.

packages/esbuild-plugin-import-path-remapper/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "peerDependencies": {
+    "esbuild": "^0.12.19"
+  },
+  "dependencies": {
+    "esbuild": "^0.12.19"
+  },
+  "devDependencies": {
+    "prettier": "^2.0.0"
+  },
```
tido64, line 30: I don't think you need `esbuild` in `dependencies` or `peerDependencies` since you only need typings from it:
```suggestion
  "devDependencies": {
    "esbuild": "^0.12.19"
  },
```

packages/esbuild-plugin-import-path-remapper/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+};
+
+const findMainSourceFile = (
+  sourcePath: string,
+  resolveDir: string
+): string | undefined => {
+  const { main } = require(require.resolve(`${sourcePath}/package.json`, {
+    paths: [resolveDir],
+  }));
```
tido64, line 17: Should we be using `resolvePath` here?

```suggestion
  const packageJsonPath = resolvePath(`${sourcePath}/package.json`, resolveDir);
  if (!packageJsonPath) {
    return undefined;
  }

  const { main } = require(packageJsonPath);
```

packages/esbuild-plugin-import-path-remapper/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  if (!main) {
+    return;
+  }
+
+  const remappedPath = `${sourcePath}/${main.replace(
+    /^(?:\.\/)?lib\/(.*)\.js/,
```
tido64, line 24: Optionally match `x` as well:
```suggestion
    /^(?:\.\/)?lib\/(.*)\.jsx?/,
```

packages/esbuild-plugin-import-path-remapper/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  return undefined;
+};
+
+const resolvePathToLib = (
+  path: string,
+  resolveDir: string,
+  resolveExtensions: string[] = []
+): string | undefined => {
+  const postFixList = [
+    "",
```
tido64, line 44: Shouldn't the file without extension be last?

packages/esbuild-plugin-import-path-remapper/src/index.ts
```diff
@@ -0,0 +1,88 @@
...
+    const resolved = resolvePath(`${path}${postFix}`, resolveDir);
+    if (resolved) {
+      return resolved;
+    }
+  }
+
+  return undefined;
+};
+
+const ImportPathRemapperPlugin = (packagePrefix: string): Plugin => ({
```
tido64, line 59: Can we turn this prefix into a regex or a tester function? Kinda limited if we require packages to be scoped or prefixed.

packages/esbuild-plugin-import-path-remapper/test/esbuild-plugin-import-path-remapper.test.js
```diff
@@ -0,0 +1,112 @@
+// @ts-check
+"use strict";
+
+describe("@rnx-kit/esbuild-plugin-import-path-remapper", () => {
+  const esbuild = require("esbuild");
+  const ImportPathRemapperPlugin = require("../lib/index");
```
tido64, line 6: You should be able to consume TypeScript with `@rnx-kit/jest-preset`.
```suggestion
  const ImportPathRemapperPlugin = require("../src/index");
```

packages/esbuild-plugin-import-path-remapper/test/input/.gitignore
```diff
@@ -0,0 +1 @@
+!node_modules
```
tido64, line 1: If you move your fixtures under a `__fixtures__` folder, you won't need this file. The `.gitignore` at repo root has a pattern for it.

tido64 (commented):
Thanks for upstreaming this. I left some minor comments.

JasonVMo:
Thanks for submitting this change Sverre! One thing to think about (and this might be a follow up PR) is whether we can share code between this package and the babel-plugin-import-path-remapper. Particularly once we start having a lot of regex expressions and handling edge cases it would be nice to only fix it in one place.

packages/esbuild-plugin-import-path-remapper/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@rnx-kit/esbuild-plugin-import-path-remapper",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
sverrejoh: I don't really have an opinion on the version number, I just went with the defaults. Does 1.0.0 make sense to you? If so I can add -alpha. Or start somewhere lower like 0.5.0 😊

packages/esbuild-plugin-import-path-remapper/src/index.ts
```diff
@@ -0,0 +1,88 @@
+import { Plugin } from "esbuild";
```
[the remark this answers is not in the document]
sverrejoh: I agree! Is there a flag in typescript to enforce this? `--isolatedModules` takes some corner cases, but it's not general I think. 

We should probably use `isolatedModules` anyway as it makes sure the code is consumable with babel.

packages/esbuild-plugin-import-path-remapper/src/index.ts
sverrejoh: `resolveDir` is esbuild's name for this (see where I get it from). Dir is more specific (path can be file or directory), and it aligns with the app, so doesn't it make sense to keep it?

packages/esbuild-plugin-import-path-remapper/src/index.ts
sverrejoh: I'm going to pretend I didn't copy/paste this from your babel code 😊. I should refactor this out to a shared package like Jason suggests.

packages/esbuild-plugin-import-path-remapper/src/index.ts
tido64: I meant `resolvePath` the function 😄 
https://github.com/microsoft/rnx-kit/blob/eb430eb9eddac23ceba428379b56e9a79a6f18bd/packages/esbuild-plugin-import-path-remapper/src/index.ts#L3

packages/esbuild-plugin-import-path-remapper/package.json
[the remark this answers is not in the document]
tido64: I'm fine with 1.0.0. I don't think starting at a lower number makes sense here since it's not a complicated plugin. And we probably don't want to make it more complicated anyway.

packages/esbuild-plugin-import-path-remapper/src/index.ts
sverrejoh: Node actually looks for the file without the extension first, and then try them: https://nodejs.org/api/modules.html#modules_file_modules

I missed `dir/index` though, so I'll add that.

Another thing is esbuild has a list of extensions that is used if you don't set `resolveExtensions`. There's no way of accessing this so I have to hard code it and set it as default.

packages/esbuild-plugin-import-path-remapper/src/index.ts
sverrejoh: Right, I see your point. The babel plugin is more general.

The esbuild API is limited to filtering out import paths by (go compatible) regex', for performance reasons. They do it like this to not have to delegate to javascript more than necessary.

What do you think if I change it from `packagePrefix` to a `packageFilter` regex and use it as the beginning of the `onResolve` filter regex?

packages/esbuild-plugin-import-path-remapper/package.json
sverrejoh: Depcheck complains though.

packages/esbuild-plugin-import-path-remapper/src/index.ts
sverrejoh: Hehe, sorry. Yes, you're probably right ;)

sverrejoh:
> Thanks for submitting this change Sverre! One thing to think about (and this might be a follow up PR) is whether we can share code between this package and the babel-plugin-import-path-remapper. Particularly once we start having a lot of regex expressions and handling edge cases it would be nice to only fix it in one place.

Yes, I'll think about that! There are some functions copied over that should be shared too. I started looking at creating a similar plugin for `swc`, which is a rust based bundler that have some other advantages compared to `esbuild`. Its API is more like babel so we could probably share even more.

Should I make a new package under `packages/` in this repo with the shared bits? I'll have to think about it, but I'll try to come up with some names that makes sense.

packages/esbuild-plugin-import-path-remapper/src/index.ts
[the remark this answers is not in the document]
tido64: @dzearing made a codemod for it: https://github.com/dzearing/transform-typed-imports ❤️ 

I might take a look at how we can integrate it into CI.

packages/esbuild-plugin-import-path-remapper/src/index.ts
tido64: Wouldn't that break platform extensions? It sounds like we would pick `SomeModule.ts` over `SomeModule.ios.ts`. Or am I misunderstanding what this is used for?

change/@rnx-kit-esbuild-plugin-import-path-remapper-2c6b6104-670f-491a-8c2c-631d578d4b9e.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
tido64, line 2: This should be minor if you want to promote `1.0.0-alpha` to `1.0.0`

packages/esbuild-plugin-import-path-remapper/src/index.ts
sverrejoh: It’ll will choose `SomeModule` (Which never exists) over `SomeModule.ts`, `SomeModule.ios.ts`. The bare .ts will still be at the end.

packages/esbuild-plugin-import-path-remapper/package.json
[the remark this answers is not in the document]
sverrejoh: Changed it to alpha.

packages/esbuild-plugin-import-path-remapper/src/index.ts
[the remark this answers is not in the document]
sverrejoh: Fixed it <3

packages/esbuild-plugin-import-path-remapper/src/index.ts
sverrejoh: I changed it to take a package name filter regular expression, `packageNameFilter`. I named it filter to be in line with the esbuild API.

sverrejoh:
Added an issue about refactoring the shared bits to a common package: https://github.com/microsoft/rnx-kit/issues/518

packages/esbuild-plugin-import-path-remapper/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@rnx-kit/esbuild-plugin-import-path-remapper",
+  "version": "1.0.0-alpha",
+  "description": "Esbuild plugin for remapping 'lib/' imports to 'src/'",
+  "homepage": "https://github.com/microsoft/rnx-kit/tree/main/packages/esbuild-plugin-import-path-remapper#readme",
+  "license": "MIT",
+  "files": [
+    "src/*"
+  ],
+  "main": "src/index.js",
```
tido64, line 10: Can you verify that these fields are still correct? Shouldn't we be publishing the lib folder?

sverrejoh: you're right!

tido64 (approved):

PR #488, opened 2021-08-16
ci: fix publishing pipeline

tido64:
### Description

Beachball is trying to bump @rnx-kit/cli to a version that already exists. cli depends on console@^1.0.1 but latest is 1.0.2.

### Test plan

Beachball doesn't have a dry-run flag but you can run `yarn publish:beachball` without a token and it would do most of the work up until the actual push to npm.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "@react-native-community/cli-server-api": "^5.0.1",
     "@rnx-kit/config": "^0.4.0",
-    "@rnx-kit/console": "^1.0.1",
+    "@rnx-kit/console": "^1.0.2",
     "@rnx-kit/dep-check": "^1.5.21",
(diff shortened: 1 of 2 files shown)


packages/cli/package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@react-native-community/cli-server-api": "^5.0.1",
     "@rnx-kit/config": "^0.4.0",
-    "@rnx-kit/console": "^1.0.1",
+    "@rnx-kit/console": "^1.0.2",
```
tido64, line 24: @afoxman: Is it really necessary to bump dependents when we publish new versions? It looks like console got bumped twice even though nothing has changed. It seems excessive to me.

tido64: I also submitted a PR to make this easier to debug in the future: https://github.com/microsoft/beachball/pull/581

kelset (approved):

PR #493, opened 2021-08-17
feat(dep-check): allow looser requirements checks

### LINKED ISSUE #491: dep-check: limitation on supported versions by libraries
When trying to upgrade an app to RN64 that is configured with `rnx-kit/dep-check`, I realised that because one of its dependencies also uses `rnx-kit/dep-check` and it declared support only up to RN63, dep-check fails.

This is because we treat "unsupported version by library" as an error level alert. We should consider changing it to warning so that we don't necessarily block an app from using an "outdated" library.

tido64:
### Description

Allow apps to depend on a newer version of React Native than their dependencies declare support for via the `--loose` flag.

Resolves #491.

### Test plan

Tests were added.

### DIFF
--- a/packages/dep-check/src/dependencies.ts
+++ b/packages/dep-check/src/dependencies.ts
@@ -16,10 +16,11 @@ import {
   profilesSatisfying,
 } from "./profiles";
 import type {
+  CheckOptions,
   PackageManifest,
   Profile,
   ProfileVersion,
-  ResolverOptions,
+  TestOverrides,
 } from "./types";
 
 type Requirements = Required<
@@ -83,7 +84,8 @@ export function getRequirements(
   targetManifest: PackageManifest,
   projectRoot: string,
   customProfiles: string | undefined,
-  options?: ResolverOptions
+  { loose }: Pick<CheckOptions, "loose">,
+  testOverrides?: TestOverrides
 ): Requirements {
   let profileVersions = getProfileVersionsFor(targetReactNativeVersion);
   if (profileVersions.length === 0) {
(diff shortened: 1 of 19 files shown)


packages/dep-check/src/dependencies.ts
```diff
@@ -104,39 +106,53 @@ export function getRequirements(
...
+        const fullTrace = [
+          message,
+          ...trace.map(({ module, reactNativeVersion, profiles }) => {
+            const satisfiedVersions = profiles.join(", ");
+            return `    [${satisfiedVersions}] satisfies '${module}' because it supports '${reactNativeVersion}'`;
+          }),
+        ].join("\n");
+        error(fullTrace);
+        if (!loose) {
+          throw new Error(message);
```
kelset, line 144: I think overall the PR is fine, but what I would change here is that when it's "loose" I would still have a console.warn to be thrown - as it is currently when loose it won't show any error at all, right?

tido64: This is the line printing to the console:
https://github.com/microsoft/rnx-kit/blob/b4f030e9df31f247852775bbb35411f7ca5967e0/packages/dep-check/src/dependencies.ts#L142
The throw is just to terminate the process. I could change it to a `warn` if that's preferred.

kelset: I see - maybe then put the `error(fullTrace);` in the if statement, and in the else put a `        warn(fullTrace);`? something like that. I think both scenarios should report the issue, but only the loose flow should be able to proceed

kelset (approved):
:shipit:

PR #496, opened 2021-08-17
ci: test on two latest LTS versions of Node

tido64:
### Description

Make sure we are compatible with the two latest LTS versions of Node.

Also, ignore if labeler fails. It won't label things correctly if PRs come from a forked repo.

### Test plan

CI should pass.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -37,7 +37,7 @@ jobs:
     name: "Build"
     strategy:
       matrix:
-        node-version: [14, 15]
+        node-version: [14, 16]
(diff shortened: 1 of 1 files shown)


JasonVMo (approved):

JasonVMo:
Just for clarification, are the two latest LTS builds 14 and 16? Or should it be 15 and 16?

tido64:
> Just for clarification, are the two latest LTS builds 14 and 16? Or should it be 15 and 16?

According to Node's release schedule, every other release is eventually LTS: https://nodejs.org/en/about/releases/. v16 will be LTS later this year. v18 will be the next LTS.

PR #500, opened 2021-08-19
fix(dep-check): fix error messages accumulating when gathering requirements

tido64:
### Description

Error messages are being accumulated and printed with every dependency that cannot be satisfied with current set of profiles. Moving the print/throw outside the loop fixes the issue.

### Test plan

Tests should pass.


kelset (approved):
🤪

PR #508, opened 2021-08-23
CLI refactoring for testability - part 2

afoxman:
### Description

In part 2, I'm adding tests for the bundle/kit-config code (refactored without changes in part 1). No user-facing code is being changed.

Part 2 has its PR base branch set to part 1. This matches how I made my branches, and shows you only the changes in part 2. *Please approve this once your review is complete. Don't wait for part 1 to be done/committed/merged before approving.*

### Test plan

Run the new test suite as part of builds and CI.


packages/cli/test/bundle/kit-config.test.ts
```diff
@@ -0,0 +1,128 @@
+import "jest-extended";
+import { BundleDefinitionWithRequiredParameters } from "../../../config/lib";
```
tido64, line 2:
```suggestion
import type { BundleDefinitionWithRequiredParameters } from "@rnx-kit/config";
```

tido64 (changes requested):

packages/cli/test/bundle/kit-config.test.ts
afoxman: The dang automation... I should have caught this, but it happens so many times. I need to just turn this off.

afoxman: I found the setting and disabled it. and fixed this.

tido64 (approved):

PR #519, opened 2021-08-23
Fix bundling

afoxman:
### Description

Fix bundling target bug.

### Test plan

Manually run bundling to ensure all platforms in kit-config are bundled.


tido64 (approved):

tido64:
Btw, how did you discover this? Can we add something to the CI to catch it earlier?

PR #520, opened 2021-08-23
Fix type-checking in Metro for windows dev machines

afoxman:
### Description

There were a few spots where path-normalization was missing in the Metro+TS code. This led to failures resolving modules, and incorrect error-reporting.

### Test plan

Manual inspection in the debugger and running bundling on all platforms.

### DIFF
--- a/packages/cli/src/metro-ts-resolver.ts
+++ b/packages/cli/src/metro-ts-resolver.ts
@@ -6,6 +6,7 @@ import {
   getMangledPackageName,
   isFileModuleRef,
   isPackageModuleRef,
+  normalizePath,
   PackageModuleRef,
   parseModuleRef,
   readPackage,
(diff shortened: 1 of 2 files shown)


packages/cli/src/metro-ts-resolver.ts
```diff
@@ -312,21 +313,25 @@ export class SourceFiles {
...
   }
 
   public set(fileName: string, metroModules: Map<string, Dependency>): void {
+    const normalized = normalizePath(fileName);
+
     // extract each Metro dependency
     const modules: ModuleMap = {};
     metroModules.forEach((value: Dependency, key: string) => {
-      modules[key] = value.absolutePath;
+      modules[key] = normalizePath(value.absolutePath);
```
tido64, line 326: Can we add tests to catch this?

tido64 (approved):

packages/cli/src/metro-ts-resolver.ts
afoxman: The solution is to have CI fail bundling when there are type-checking errors. I just haven't gotten there yet. I'll file a bug.

PR #522, opened 2021-08-24
CLI refactoring for testability - part 4

afoxman:
### Description

Part 4 is all about moving to a `KitBundleConfig` object which is a combination of a platform field and the platform-overridden kit bundle definition. I've made a function which creates an array of `KitBundleConfig` objects. They are used in the main bundle code as the basis for generating each bundle. I added a set of tests to protect the new function.

As part of this change, I had to rename the "apply overrides" method and arg type in code and tests.

### Test plan

New tests pass.

CI passes, and bundling continues to work without issue.

### DIFF
--- a/packages/cli/src/bundle/overrides.ts
+++ b/packages/cli/src/bundle/overrides.ts
@@ -1,8 +1,8 @@
-import type { BundleDefinitionWithRequiredParameters } from "@rnx-kit/config";
 import type { BundleArgs } from "@rnx-kit/metro-service";
 import { pickValues } from "@rnx-kit/tools-language/properties";
+import type { KitBundleConfig } from "../bundle/types";
 
-export type BundleDefinitionOverrides = {
+export type KitBundleConfigOverrides = {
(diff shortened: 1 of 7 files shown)


packages/cli/src/bundle/overrides.ts
```diff
@@ -1,8 +1,8 @@
-import type { BundleDefinitionWithRequiredParameters } from "@rnx-kit/config";
 import type { BundleArgs } from "@rnx-kit/metro-service";
 import { pickValues } from "@rnx-kit/tools-language/properties";
+import type { KitBundleConfig } from "../bundle/types";
```
tido64, line 3:
```suggestion
import type { KitBundleConfig } from "./types";
```

tido64 (approved):

packages/cli/src/bundle/overrides.ts
afoxman: Fixed in the next PR.

PR #523, opened 2021-08-24
chore: bump react-native to 0.65

tido64:
### Description

Bumps `react-native` to 0.65.

### Test plan

1. Build `test-app`:
   ```
   yarn
   cd packages/test-app
   yarn build --dependencies
   ```
2. Build and launch the Android app: `yarn android`
3. Build and launch the iOS app:
   ```
   yarn ios
   ```
4. Build and launch the Windows app:
   ```
   yarn install-windows-test-app --use-nuget
   yarn windows
   ```
5. Test that bundling still works: `yarn bundle`
6. Test that bundling with esbuild still works: `yarn bundle+esbuild`

| Platform | Home | App |
| :- | - | -|
| Android | Screenshot_1629798286 | Screenshot_1629798288 |
| iOS | image |  image |
| Windows | | |

### Known issues

- Android: The background image is not displayed (#528)
- Windows: Windows test app doesn't boot app and just displays a blank screen (#529)

Both of these issues are present on `main` today, so they are not regressions.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -19,7 +19,6 @@
     "test": "rnx-kit-scripts test"
   },
   "dependencies": {
-    "@react-native-community/cli-server-api": "^5.0.1",
     "@rnx-kit/config": "^0.4.1",
     "@rnx-kit/console": "^1.0.2",
     "@rnx-kit/dep-check": "^1.7.2",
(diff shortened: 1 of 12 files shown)


afoxman (approved):

packages/cli/package.json
```diff
@@ -19,7 +19,6 @@
     "test": "rnx-kit-scripts test"
   },
   "dependencies": {
-    "@react-native-community/cli-server-api": "^5.0.1",
```
tido64, line 22: @afoxman: I had to move this down to `peerDependencies` to avoid forcing 5.x on our users. Let me know if this doesn't look right.

afoxman: @tido64 it should work, it just complicates the story for our consumers... i think. Our consumers will all pull in react-native which in turn pulls in @rnc/cli (including the server api package for `start` command). Can the transitive dependency on the rnc/cli package be used to meet this peerDependency for our customers?

tido64: I don't think it does, and it will probably be a problem for package managers that are strict with `peerDependencies`. We can't have loose version numbers in `dependencies` either since Yarn will always resolve it to the latest possible. One thing we _can_ do is to mark this dependency as optional, and add proper error messages when it is missing.

tido64: Let me know what you think: https://github.com/microsoft/rnx-kit/pull/541

PR #537, opened 2021-08-24
CLI refactoring for testability - part 6

afoxman:
### Description

In part 6, I've combined kit bundle configuration with command-line-only params (dev, minify) to produce a final bundle config. This is what will drive each bundling operation.

To support this new type, I've added utility functions which "extend" an object from type `T` to type `TExtended extends T`.

### Test plan

Add new tests to cover the "extend" functions.

Run existing tests and bundling to ensure things continue working as expected.

### DIFF
--- a/packages/cli/src/bundle/types.ts
+++ b/packages/cli/src/bundle/types.ts
@@ -6,5 +6,23 @@ import type { AllPlatforms } from "@rnx-kit/tools-react-native/platform";
 
 export type KitBundleConfig = BundleParameters &
   BundleRequiredParameters & {
+    /**
+     * Target platform for the bundle
+     */
     platform: AllPlatforms;
   };
+
+export type BundleConfig = KitBundleConfig & {
+  /**
+   * Choose whether or not this will be a "developer" bundle. The alternative is a "production" bundle.
+   *
+   * When `true`, warnings are enabled, and the bundle is not minified by default.
+   * When `false`, Warnings are disabled and the bundle is minified by default.
+   */
+  dev: boolean;
+
+  /**
+   * Optionally choose whether or not the bundle is minified. When not set, minification is controlled by the `dev` property.
+   */
+  minify?: boolean;
+};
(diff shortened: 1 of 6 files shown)


packages/tools-language/test/properties.test.ts
```diff
@@ -90,4 +95,61 @@ describe("Language > Props", () => {
...
+  });
+
+  test("extendObjectArray adds the extended props to each object in the array", () => {
+    const copy = [{ ...baseObj }];
+    const extended = extendObjectArray<BaseType, ExtendedType>(
+      copy,
+      extendedProps
+    );
+    expect(extended).toBeArrayOfSize(1);
+    expect(extended).toIncludeSameMembers([
```
tido64, line 132: Isn't `toEqual` sufficient here?
```suggestion
    expect(extended).toEqual([
```

packages/tools-language/test/properties.test.ts
```diff
@@ -90,4 +95,61 @@ describe("Language > Props", () => {
...
+    );
+  });
+
+  test("extendObject makes no changes when extended props are optional and are not given", () => {
+    const copy = { ...baseObj };
+    expect(extendObject<BaseType, ExtendedType>(copy, {})).toEqual(baseObj);
+  });
+
+  test("extendObjectArray adds the extended props to each object in the array", () => {
+    const copy = [{ ...baseObj }];
```
tido64, line 126: Can you also test with array that contains more elements?

packages/cli/src/bundle/types.ts
```diff
@@ -6,5 +6,23 @@ import type { AllPlatforms } from "@rnx-kit/tools-react-native/platform";
...
+     */
     platform: AllPlatforms;
   };
+
+export type BundleConfig = KitBundleConfig & {
+  /**
+   * Choose whether or not this will be a "developer" bundle. The alternative is a "production" bundle.
+   *
+   * When `true`, warnings are enabled, and the bundle is not minified by default.
+   * When `false`, Warnings are disabled and the bundle is minified by default.
```
tido64, line 20:
```suggestion
   * When `false`, warnings are disabled and the bundle is minified by default.
```

tido64: Should we also mention that when `dev` is true, optimizations like constant folding are also disabled?

afoxman: Sure. I got this text from the bundle command-line description, and I'm sure there's a lot more that is different between the two.

tido64 (approved):

PR #539, opened 2021-08-25
CLI refactoring for testability - part 8

afoxman:
### Description

In part 8, I've added a `createDirectory` function to the Node tools package. I used it in the metro bundling code to replace calls to `fs.mkdirSync(...)`.

### Test plan

New tests for `createDirectory` pass. Existing build, test, bundle CI loop passes.


packages/tools-node/test/fs.test.ts
```diff
@@ -28,4 +42,19 @@ describe("Node > FS", () => {
...
+    createDirectory(p);
+    expect(fs.existsSync(p)).toBeTrue();
+  });
+
+  test("createDirectory() creates a parent directory", () => {
+    const parent = path.join(testTempDir, "parentdir");
+    const p = path.join(parent, "testdir");
+    expect(fs.existsSync(parent)).toBeFalse();
+    createDirectory(p);
+    expect(fs.existsSync(parent)).toBeTrue();
```
tido64, line 58: Can we also verify that its mode/permissions are correctly set?

tido64 (approved):

packages/tools-node/test/fs.test.ts
afoxman: I'll give it a shot. Who knows how well it will translate on windows :)

tido64: Making it conditional on unix filesystems is probably fine :)

tido64 (approved):

packages/tools-node/test/fs.test.ts
afoxman: I found common ground that I think will work -- i feed in rwx|rx|rx and on macos i get that back plus a higher bit (which i ignore). On windows I get back rw|r|r which is good enough of a test, so I've made that the lowest common denominator here. I test for rw|r|r as well as "isDirectory". I'm going assume the exec bits will be there if rw|r|r is there.

afoxman: At some point I'm testing nodejs, and not the library code, so this seems good enough.

tido64: It doesn't check that we didn't accidentally set `rw|rw|rw` though. We want to make sure that only the owner has write permissions.

PR #540, opened 2021-08-25
CLI refactoring for testability - part 9

afoxman:
### Description

In part 9, I separated out the code which converts from a `BundleConfig` to Metro's `BundleArgs`. This is the type used when invoking Metro's `bundle()` API. I also added tests for the conversion function.

### Test plan

New conversion tests pass. Existing build, test, and bundle CI loop continues to pass.


packages/cli/src/bundle/metro.ts
```diff
@@ -10,40 +10,20 @@ import type { TSProjectInfo } from "../types";
...
-    bundlePrefix,
-    bundleEncoding,
-    sourceMapSourceRootPath,
-    detectCyclicDependencies,
-    detectDuplicateDependencies,
-    typescriptValidation,
-    experimental_treeShake,
-  } = bundleConfig;
+): MetroBundleArgs {
+  const { platform, distPath } = bundleConfig;
```
tido64, line 21: We should extract everything we need here once, then reference them further down:
```suggestion
  const {
    assetsPath: assetsDest,
    bundleEncoding,
    bundlePrefix,
    entryPath: entryFile,
    dev,
    distPath,
    platform,
    sourceMapSourceRootPath: sourcemapSourcesRoot,
  } = bundleConfig;
```

packages/cli/src/bundle/metro.ts
```diff
@@ -58,17 +38,44 @@ export async function metroBundle(
...
+ * Run the Metro bundler.
+ *
+ * @param tsservice TypeScript service to use for type-checking (when enabled)
+ * @param metroConfig Metro configuration
+ * @param bundleConfig Bundle configuration
+ */
+export async function metroBundle(
+  tsservice: Service,
+  metroConfig: ConfigT,
+  bundleConfig: BundleConfig
```
tido64, line 64: Nit: We can also extract everything we need from `BundleConfig` here:
```suggestion
  {
    detectCyclicDependencies,
    detectDuplicateDependencies,
    dev,
    entryPath,
    experimental_treeShake,
    minify
    platform,
    typescriptValidation,
  }: BundleConfig
```

packages/cli/test/bundle/metro.test.ts
```diff
@@ -0,0 +1,116 @@
+import type { AllPlatforms } from "@rnx-kit/tools-react-native/platform";
+import "jest-extended";
```
tido64, line 2: Is this being used?

packages/cli/src/bundle/metro.ts
afoxman: I had that but then went back to only extracting what I used more than once. My thinking is that I don't need storage space for the single-use params and I only want to read them once. I realize I am making a lot of assumptions about how the JSVM works, so if I've gone wrong, let me know... And then there's the calculus of source-code size and how much will be minified. I don't have enough exp with that to know how it will impact this, either.

How do you see one choice vs. the other when considering storage space and number of reads? Are they real concerns?

tido64: I should probably have prefixed this with "Nit" 😛 

I think it's mostly subjective at this point. I like having everything destructured once so we don't make separate lookups. Any savings (if any) would be insignificant. I had a look at the output, and it seems like the statement is preserved, so I am _assuming_ that the JSVM can make some optimizations under the hood if we used destructuring. Minified, I think this syntax will be smaller since you don't have to keep referencing `bundleConfig`.

packages/cli/test/bundle/metro.test.ts
```diff
@@ -0,0 +1,115 @@
+import type { AllPlatforms } from "@rnx-kit/tools-react-native/platform";
+import path from "path";
+import { createMetroBundleArgs } from "../../src/bundle/metro";
+import { BundleConfig } from "../../src/bundle/types";
```
tido64, line 4:
```suggestion
import type { BundleConfig } from "../../src/bundle/types";
```

tido64 (approved):

PR #553, opened 2021-08-27
chore: upgrade to TypeScript 4.4

tido64:
### Description

As of 4.4, the parameter in the `catch` clause is `unknown` by default:
https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-beta/#use-unknown-catch-variables

### Test plan

All existing code should build.

### DIFF
--- a/packages/tools-language/test/properties.test.ts
+++ b/packages/tools-language/test/properties.test.ts
@@ -196,4 +196,12 @@ describe("Language > Props > hasProperty", () => {
     expect(hasProperty(testObj, "message")).toBe(true);
     expect(hasProperty(testObj, "stack")).toBe(false);
   });
+
+  test("returns true for objects with inherited property", () => {
+    const testProto = { message: "code", code: 0 };
+    const testObj = Object.create(testProto);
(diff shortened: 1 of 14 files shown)


packages/tools-language/test/properties.test.ts
```diff
@@ -166,10 +173,27 @@ describe("Language > Props", () => {
...
+
+describe("Language > Props > hasProperty", () => {
+  test("returns false for primitives", () => {
+    expect(hasProperty("string", "0")).toBe(false);
+    expect(hasProperty(0, "0")).toBe(false);
+    expect(hasProperty(true, "0")).toBe(false);
+    expect(hasProperty(null, "0")).toBe(false);
+    expect(hasProperty(undefined, "0")).toBe(false);
+  });
+
```
afoxman, line 192: Since you're using `in`, consider adding a test to show that inherited props are also detected.

```typescript
const c = new String("green");
expect(hasProperty(c , "length")).toBe(true);
```

afoxman (approved):

PR #565, opened 2021-08-31
Set Metro package roots separately from Metro watch folders

### LINKED ISSUE #536: We should provide a way to have metro to not watch all the roots
Metro by default will watch all the files within all the roots.  This means that generally metro ends up watching all the files within node_modules. 

The haste-map built internally by metro currently only has a single flag to control if the entire map is watched.  Really this needs to be expanded to have a config to say which folders should be watched, rather than always watching all the roots.
https://github.com/facebook/metro/blob/96fb6e904e1660b37f4d1f5353ca1e5477c4afbf/packages/metro/src/node-haste/DependencyGraph.js#L165

afoxman:
### Description

When Metro runs, it gets the list of package root paths from the config property `watchFolders`. This means all packages, in-repo and external, must be watched for changes. In large mono-repos, this causes problems when thousands of external packages are used. File-watcher limits are hit, and a lot of CPU, memory and disk resources are used.

This PR **PATCHES** Metro. It adds support for a new `roots` Metro config property to explicitly separate package roots from watch folders. It is the patch-equivalent of this Metro PR: https://github.com/facebook/metro/pull/701

This PR also makes use of the patch.

1. It exposes two new APIs in `@rnx-kit/metro-config` which give the developer the data needed to set `roots` and `watchFolders` together.
2. It updates the `test-app` package to use the new `metro-config` APIs, and adds warning-suppression because Metro complains that `roots` is not a valid property.

When the Metro PR is merged and published, we will upgrade our packages to make use of it and the remove the patching and test-app changes.

Fixes #536.

### Test plan

Add tests for new APIs. Verify bundling of all platforms in the debugger. Ensure that `roots` and `watchFolders` are set correctly, and that bundling still works.


packages/metro-config/src/index.js
```diff
@@ -5,17 +5,19 @@
  * @typedef {import("metro-config").ExtraTransformOptions} ExtraTransformOptions;
  * @typedef {import("metro-config").InputConfigT} InputConfigT;
  * @typedef {import("type-fest").PartialDeep<InputConfigT>} MetroConfig;
+ * @typedef {{ externalPackages: string[]; internalPackages: string[] }} RepoPackageInfo;
```
tido64, line 8: Do you anticipate that we will have more than one root some time in the future? Otherwise, I think we can change this type to make more sense:
```suggestion
 * @typedef {{ root: string; packages: string[] }} WorkspaceInfo;
```

packages/metro-config/src/index.js
```diff
@@ -27,28 +29,61 @@ function defaultWatchFolders(projectRoot) {
...
     }
 
     // In a monorepo, in particular when using Yarn workspaces, packages are
-    // symlinked in the root `node_modules` folder so it needs to be watched.
-    return [
-      path.join(root, "node_modules"),
-      ...packages.map((pkg) => path.dirname(pkg)),
-    ];
+    // symlinked in the root `node_modules` folder so it needs to be included.
+    info.externalPackages.push(path.join(root, "node_modules"));
```
tido64, line 47: Can we move these lines above the `getAllPackageJsonFiles()` line? This piece of information shouldn't be gated behind how many packages there are in the workspace.

packages/metro-config/src/index.js
```diff
@@ -27,28 +29,61 @@ function defaultWatchFolders(projectRoot) {
...
+}
+
+/**
+ * A minimum list of package root folders.
+ * @param {string | undefined} projectRoot
+ * @returns {string[]}
+ */
+function defaultRoots(projectRoot) {
+  const info = getRepoPackageInfo(projectRoot);
+  return [...info.externalPackages, ...info.internalPackages];
```
tido64, line 84: The roots should already contain the packages:
```suggestion
  const info = getRepoPackageInfo(projectRoot);
  return info.externalPackages;
```

packages/metro-config/src/index.js
```diff
@@ -27,28 +29,61 @@ function defaultWatchFolders(projectRoot) {
...
+
+/**
+ * A minimum list of in-repo folders that should be watched by Metro.
+ * @param {string | undefined} projectRoot
+ * @returns {string[]}
+ */
+function defaultWatchFoldersInRepo(projectRoot) {
+  const info = getRepoPackageInfo(projectRoot);
+  return info.internalPackages;
+}
```
tido64, line 75: I wonder if it's possible to detect which version of CLI and/or Metro we're dealing with and return the correct value here. That would eliminate the need for a separate `defaultWatchFoldersInRepo()`. I think it would only confuse users if we have two. This would also allow us to add `roots` if it's supported. The less the user has to do, the less there is for things to go wrong 😄

packages/metro-config/src/index.js
tido64: Also, I think we can simplify `getDefaultRoots()` if we dropped `node_modules`. I forgot to mention this in the comment below 😛

tido64: Or if removing `node_modules` is a bad idea, then discard that comment 😄

packages/metro-config/src/index.js
afoxman: Andrew tells me that with pnpm, you end up with thousands of package roots. Something to do with the way the symlinks and the backing store is laid out, I think.

packages/metro-config/src/index.js
afoxman: I could move it, but I like setting the output all in one place at the end, in case I need to return early (as I do in a few places). I want it to be either fully filled out or not set at all.

As for removing node_modules from this function, it needs to be there to preserve the original behavior of `defaultWatchFolders`. This is already part of the public API and can't change.

packages/metro-config/src/index.js
afoxman: That's a good idea. Since the Metro PR is still being reviewed and may need more work, I can't add anything like that right now (no version number). Instead, I've added it to the "cleanup" issue -- #566.

packages/metro-config/src/index.js
afoxman: They don't. `externalPackages` is stuff that's not in-repo, `internalPackages` is only in-repo stuff.

packages/metro-config/src/index.js
tido64: Do you plan to address #566 before or after merging this PR? I'm hesitant to adding new methods that we take away in the next version.

afoxman:
Back to draft. Looks like a different patch is needed, deeper in jest-haste-map.

packages/metro-config/src/index.js
afoxman: It would be after. I expect us to be using patches until the Metro/jest-haste-map PRs are merged, published, and incorporated into RN/CLI (and those are published). Then we can start to peel things back.

These APIs could stay. There's not reason to deprecate them. Clients can stop using them once the upstream PRs are in and we have sensible defaults in place in this package. But keeping them around won't hurt.

afoxman:
We're not going forward with this PR.

PR #568, opened 2021-09-03
test(dep-check): refactor tests for incoming meta packages feature

tido64:
### Description

The meta packages feature shouldn't break existing tests but we have to make some changes to the tests so they can handle both normal and meta packages.

### Test plan

Tests should pass.

### DIFF
--- a/packages/dep-check/package.json
+++ b/packages/dep-check/package.json
@@ -45,7 +45,6 @@
     "@types/semver": "^7.0.0",
     "@types/yargs": "^16.0.0",
     "markdown-table": "^2.0.0",
-    "prettier": "^2.0.0",
     "rnx-kit-scripts": "*",
     "typescript": "^4.0.0"
   },
@@ -54,6 +53,9 @@
   },
   "jest": {
     "preset": "@rnx-kit/jest-preset",
+    "collectCoverageFrom": [
+      "src/**"
+    ],
     "roots": [
       "test"
     ],
--- a/packages/dep-check/test/capabilities.test.ts
+++ b/packages/dep-check/test/capabilities.test.ts
@@ -4,6 +4,7 @@ import { getProfilesFor } from "../src/profiles";
 import profile_0_62 from "../src/profiles/profile-0.62";
 import profile_0_63 from "../src/profiles/profile-0.63";
 import profile_0_64 from "../src/profiles/profile-0.64";
+import { mockResolver } from "./helpers";
 
 describe("capabilitiesFor()", () => {
   test("returns `undefined` when react-native is not a dependency", () => {
@@ -86,14 +87,14 @@ describe("resolveCapabilities()", () => {
 
   test("dedupes packages", () => {
     const packages = resolveCapabilities(
-      ["core-android", "core-ios", "test-app"],
+      ["core", "core", "test-app"],
       [profile_0_64]
     );
 
-    const { name } = profile_0_64["core-ios"];
+    const { name } = profile_0_64["core"];
     const { name: testAppName } = profile_0_64["test-app"];
     expect(packages).toEqual({
-      [name]: [profile_0_64["core-ios"]],
+      [name]: [profile_0_64["core"]],
       [testAppName]: [profile_0_64["test-app"]],
     });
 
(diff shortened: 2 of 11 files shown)


packages/dep-check/package.json
```diff
@@ -45,7 +45,6 @@
     "@types/semver": "^7.0.0",
     "@types/yargs": "^16.0.0",
     "markdown-table": "^2.0.0",
-    "prettier": "^2.0.0",
```
kelset, line 48: no need for prettier? is it because it's defined at root level?

packages/dep-check/test/capabilities.test.ts
```diff
@@ -86,14 +87,14 @@ describe("resolveCapabilities()", () => {
 
   test("dedupes packages", () => {
     const packages = resolveCapabilities(
-      ["core-android", "core-ios", "test-app"],
+      ["core", "core", "test-app"],
```
kelset, line 90:
```suggestion
      ["core", "test-app"],
```

do you need "core" twice here?

tido64: Yes, this tests that `core` is deduped. `core`, `core-android` and `core-ios` all point to the same package, but I had in mind to change the latter two to be meta.

packages/dep-check/package.json
tido64: We use the formatter that comes with `rnx-kit-scripts` (which is Prettier). So we don't need to declare it here.

packages/dep-check/package.json
```diff
@@ -54,6 +53,9 @@
   },
   "jest": {
     "preset": "@rnx-kit/jest-preset",
+    "collectCoverageFrom": [
```
afoxman, line 56: Can coverage be included in our preset so every package gets analyzed? I'd like to one day turn on coverage-minimums...

packages/dep-check/test/manifest.test.ts
```diff
@@ -91,11 +99,23 @@ describe("updateDependencies()", () => {
...
-      react: `${profile_0_63["react"].version} || ${profile_0_64["react"].version}`,
-      "react-native": `${profile_0_63["core-ios"].version} || ${profile_0_64["core-ios"].version}`,
-      "react-native-macos": `${profile_0_63["core-macos"].version} || ${profile_0_64["core-macos"].version}`,
+      react: `${packageVersion(profile_0_63, "react")} || ${packageVersion(
+        profile_0_64,
+        "react"
+      )}`,
+      "react-native": `${packageVersion(
+        profile_0_63,
+        "core"
```
afoxman, line 108: "core" used to be "core-ios" -- is this a typo? I saw similar substitutions elsewhere in this PR so I wasn't sure.

afoxman: There are a few more instance of this substitution in the tests below.

afoxman: Is "core" going to be a meta package, meaning it will be mapped to core-ios or core-android (or the whatever they represent) at runtime, depending on the selected platform?

packages/dep-check/package.json
tido64: It is enabled by default. But if you have non-test files in the test folder, they will be counted towards total coverage. I added a test helper file in the test folder.

packages/dep-check/test/manifest.test.ts
tido64: No, this is correct for now. The mapping should be straightforward:

| Capability | Description |
| :- | :- |
| core | react-native core |
| core-android | react-native bits that target Android (would be react-native-android if it existed) |
| core-ios | react-native bits that target iOS (would be react-native-ios if it existed) |
| core-macos | react-native-macos |
| core-windows | react-native-windows |

The introduction of `core` was due to `react-native-windows` having a dependency on `react-native`. Previously, you would have to add `core-windows` and one of `core-android` or `core-ios` just to pull in `react-native`. We added `core` as an alias to `react-native` so it doesn't feel as awkward. With meta capabilities implemented, `core-windows` will just take a dependency on `core`, and users no longer have to add `core` manually. The idea is to make `core-ios` and `core-android` meta capabilities that point to `core` since both are included in react-native right now. When, or if, they get extracted, they would point to `react-native-ios` + `react-native`, or equivalent.

afoxman (approved):

PR #577, opened 2021-09-03
feat(dep-check): add support for meta packages

### LINKED ISSUE #256: dep-check: Should we add support for "meta" packages?
> Can a capability resolve to more than one package? It seems reasonable that something like networking might resolve to an http client, a websocket client, and an ssl/tls library.
> 
> I know the default resolver doesn't support this. Maybe it should.

-- see discussion: https://github.com/microsoft/rnx-kit/pull/249#discussion_r637071797

tido64:
### Description

Adds support for dependencies and meta packages:
- Dependencies: This allows us to declare that `core-windows` has a dependency on `core` as `react-native-windows` has a peer dependency on `react-native`.
- Meta packages: These are packages that don't provide anything by itself, but can reference other capabilities. For instance, you can define a capability, `core/all`, to pull in all platform capabilities, i.e. `[core-android, core-ios, core-macos, core-windows]`.

Resolves #256.

### Test plan

Tests were added.

### DIFF
--- a/packages/dep-check/README.md
+++ b/packages/dep-check/README.md
@@ -227,8 +227,8 @@ that `core-windows` depends on `core`:
 }
 ```
 
-You can also create "meta" capabilities that don't resolve to a package, but to
-a list of capabilities instead:
+You can also create capabilities that don't resolve to a package, but to a list
+of capabilities instead:
 
 ```js
 {
--- a/packages/dep-check/README.md
+++ b/packages/dep-check/README.md
@@ -213,7 +213,7 @@ is a capability that resolves to `react-native`:
 }
 ```
 
-A capability can depend on other capabilies. For example, we can ensure that
+A capability can depend on other capabilities. For example, we can ensure that
 `react-native` gets installed along with `react-native-windows` by declaring
 that `core-windows` depends on `core`:
 
--- a/packages/dep-check/README.md
+++ b/packages/dep-check/README.md
@@ -114,12 +114,11 @@ your `package.json`.
 
 ## Capabilities
 
-To add new capabilities, first add it to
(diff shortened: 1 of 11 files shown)


kelset:
A couple things:
1) I'm not super clear on what you mean by meta here - could you expand the PR body to explain better?
2) similarly, worth modifying docs to add an explanation about this
3) it sounds like a fairly important "new feature", maybe worth considering setting it as a new major? Or too much hassle to then update all the places in which we use dep-check?

tido64:
> 1. I'm not super clear on what you mean by meta here - could you expand the PR body to explain better?

Good point. Updated.

> 2. similarly, worth modifying docs to add an explanation about this

The README is missing a whole bunch of stuff actually. I'm adding it now.

> 3. it sounds like a fairly important "new feature", maybe worth considering setting it as a new major? Or too much hassle to then update all the places in which we use dep-check?

I don't think this warrants a major bump. No breaking changes are introduced.

packages/dep-check/README.md
```diff
@@ -161,6 +160,90 @@ For an example, have a look at how the
...
+{
+  "core": {
+    name: "react-native",
+    version: "0.0.0",
+  },
+}
+```
+
+A capability can depend on other capabilies. For example, we can ensure that
+`react-native` gets installed along `react-native-windows` by declaring that
```
afoxman, line 217:
```suggestion
`react-native` gets installed along with `react-native-windows` by declaring that
```

packages/dep-check/test/capabilities.test.ts
```diff
@@ -151,4 +151,91 @@ describe("resolveCapabilities()", () => {
...
+        "0.64": {
+          connor: {
+            name: "#meta",
+            capabilities: ["core", "reese"],
+          },
+          reese: {
+            name: "#meta",
+            capabilities: ["t-800"],
+          },
+          "t-800": {
```
afoxman, line 221: This circular reference as a hint of "I'll be back" in it...

afoxman (approved):

PR #578, opened 2021-09-03
fix(dep-check): remove `reactNativeDevVersion` if `kitType` is `app`

tido64:
### Description

If `reactNativeDevVersion` is found during upgrading profile version, we should remove it to avoid confusion.

### Test plan

Tests were added.

### DIFF
--- a/packages/dep-check/src/setVersion.ts
+++ b/packages/dep-check/src/setVersion.ts
@@ -77,7 +77,11 @@ export async function makeSetVersionCommand(
     }
 
     rnxKitConfig.reactNativeVersion = supportedVersions;
-    rnxKitConfig.reactNativeDevVersion = targetVersion;
+    if (rnxKitConfig.kitType === "app") {
+      delete rnxKitConfig.reactNativeDevVersion;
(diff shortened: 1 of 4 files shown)


kelset (approved):

packages/dep-check/src/setVersion.ts
```diff
@@ -77,7 +77,11 @@ export async function makeSetVersionCommand(
     }
 
     rnxKitConfig.reactNativeVersion = supportedVersions;
-    rnxKitConfig.reactNativeDevVersion = targetVersion;
+    if (rnxKitConfig.kitType === "app") {
+      delete rnxKitConfig.reactNativeDevVersion;
```
afoxman, line 81: This seems like a good place for a message -- like the user did something wrong here, and we're handling it, but we should teach them to stop doing it.

Or if this isn't a user issue, then we're compensating here for something else that we control. Should we fix that something else directly?

tido64: The warning was introduced in an earlier PR: #557

PR #586, opened 2021-09-07
fix(metro-config): wrong module resolved due to ignoring symlinks

### LINKED ISSUE #585: metro-resolver-symlinks throws error because of excludeExtraCopiesOf
### What happened?

Made a brand new react native project with the react native typescript template. Installed dependencies with pnpm and added some rnx-kit stuff.

Added the symlink resolver to metro.config.js.

First error I was getting was like 
`MetroSymlinksResolver is not a function.`
This was from the example in the readme which I copy pasted. I had to use `const MetroSymlinksResolver = require('@rnx-kit/metro-resolver-symlinks').default` instead.

After that I started getting `react could not be found.`

I followed the exception down to around here:
```
 * @param {string} name
 * @param {string=} projectRoot
 * @returns {string | undefined}
 */
function resolveModule(name, projectRoot) {
  const { findPackageDependencyDir } = require("@rnx-kit/tools-node/package");
  return findPackageDependencyDir(
    { name },
    { startDir: projectRoot, allowSymlinks: false }
  );
}
```
where allowSymlinks is hardcoded as false in @rnx-kit/metro-config.

### Affected Package

@rnx-kit/metro-resolver-symlinks

### Version

0.1.2

- [X] Android
- [X] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
Error: Command failed: wmic os get Caption
'wmic' is not recognized as an internal or external command,
operable program or batch file.

Well, something is messed up here.
```

### Steps to Reproduce

[Repro repo](https://github.com/icesuicune/rnxSymlinkErrorRepro)
pnpm install -g
cd MyApp
pnpm install
pnpx rnx-kit --config ./metro.config.js

The repo is basically the react native generated template, with bits taken from the /test app package in this project.

### Code of Conduct

tido64:
### Description

In a pnpm-managed repository, we cannot ignore symlinks. This change will allow symlinks, and resolve them to a real path.

Resolves #585.

### Test plan

Check out the repro in #585:

```
git clone https://github.com/icesuicune/rnxSymlinkErrorRepro.git
cd rnxSymlinkErrorRepro/MyApp
pnpm install
pnpm react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.ios.jsbundle --assets-dest dist
```

### DIFF
--- a/packages/metro-config/src/index.js
+++ b/packages/metro-config/src/index.js
@@ -62,10 +62,11 @@ function defaultWatchFolders(projectRoot) {
  */
 function resolveModule(name, projectRoot) {
   const { findPackageDependencyDir } = require("@rnx-kit/tools-node/package");
-  return findPackageDependencyDir(
+  const result = findPackageDependencyDir(
     { name },
-    { startDir: projectRoot, allowSymlinks: false }
+    { startDir: projectRoot, allowSymlinks: true }
   );
+  return result && require("fs").realpathSync(result);
 }
 
 /**
--- a/packages/metro-config/src/index.js
+++ b/packages/metro-config/src/index.js
@@ -52,9 +52,8 @@ function defaultWatchFolders(projectRoot) {
 /**
  * Returns the path to specified module; `undefined` if not found.
  *
- * Note that this function ignores symlinks. When creating rules to
- * exclude extra copies, Metro will be unable to resolve packages
- * because it doesn't resolve symlinks and all real paths are excluded.
+ * Note that this function resolves symlinks. This is necessary for setups that
+ * only have symlinks under `node_modules` (e.g. with pnpm).
  *
  * @param {string} name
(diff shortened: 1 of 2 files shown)


packages/metro-config/src/index.js
```diff
@@ -62,10 +62,11 @@ function defaultWatchFolders(projectRoot) {
  */
 function resolveModule(name, projectRoot) {
   const { findPackageDependencyDir } = require("@rnx-kit/tools-node/package");
-  return findPackageDependencyDir(
+  const result = findPackageDependencyDir(
     { name },
-    { startDir: projectRoot, allowSymlinks: false }
```
afoxman, line 67: Function comment should be updated to match the new behavior.

afoxman (approved):

PR #587, opened 2021-09-07
fix(metro-resolver-symlinks): fails when path is child of origin path

### LINKED ISSUE #585: metro-resolver-symlinks throws error because of excludeExtraCopiesOf

tido64:
### Description

`metro-resolver-symlinks` returns the correct path, but in the wrong format, when a path is a child of origin module path.

Resolves #585.

### Test plan

Check out the repro in #585. You may have to apply #586 first before reproing this:

```
git clone https://github.com/icesuicune/rnxSymlinkErrorRepro.git
cd rnxSymlinkErrorRepro/MyApp
pnpm install
pnpm react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.ios.jsbundle --assets-dest dist
```

### DIFF
--- a/change/@rnx-kit-metro-resolver-symlinks-63a399a7-93d8-4425-94af-bdab500a153a.json
+++ b/change/@rnx-kit-metro-resolver-symlinks-63a399a7-93d8-4425-94af-bdab500a153a.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Fix wrong resolved path when module is relative to current working directory",
+  "comment": "Fixed wrong resolved path when module is in a path within origin module path",
   "packageName": "@rnx-kit/metro-resolver-symlinks",
   "email": "4123478+tido64@users.noreply.github.com",
   "dependentChangeType": "patch"
--- a/change/@rnx-kit-metro-resolver-symlinks-63a399a7-93d8-4425-94af-bdab500a153a.json
+++ b/change/@rnx-kit-metro-resolver-symlinks-63a399a7-93d8-4425-94af-bdab500a153a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix wrong resolved path when module is relative to current working directory",
+  "packageName": "@rnx-kit/metro-resolver-symlinks",
+  "email": "4123478+tido64@users.noreply.github.com",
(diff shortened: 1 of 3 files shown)


change/@rnx-kit-metro-resolver-symlinks-63a399a7-93d8-4425-94af-bdab500a153a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix wrong resolved path when module is relative to current working directory",
```
tido64, line 3: Add tests and update this comment.

acoates-ms (approved):

afoxman (approved):

PR #607, opened 2021-09-13
fix(cli): cannot disable plugins when bundling/serving

### LINKED ISSUE #605: detectDuplicateDependencies and detectCyclicDependencies ignore false setting
So i am trying to get rnx-kit working on a pre-existing repo, and found that I couldn't seem to disable cyclic dependency detection and duplicate dependency detection. I have them configured in my package.json, and other config values seem to get picked up by the bundler. Even the ignoredModules array gets used properly. However, setting the values to false or removing the properties entirely from the rnx-kit section doesn't do anything.

Here is a chunk of my configuration:
```json
"rnx-kit": {
    "reactNativeVersion": "^0.63",
    "kitType": "app",
    "bundle": {
      "entryPath": "index.js",
      "distPath": "dist",
      "assetsPath": "dist",
      "detectCyclicDependencies": false,
      "bundlePrefix": "main",
      "typescriptValidation": false,
      "detectDuplicateDependencies": {
        "ignoredModules": [
          "react-is", "react-native-modal", "postcss-value-parser", "react-native-device-info", "domelementtype", "entities", "react-native-svg", "escape-string-regexp", "color-name", "fbjs", "eventemitter3", "tslib", "lodash-es", "reselect", "@react-navigation/core", "hoist-non-react-statics", "@react-navigation/native", "react-native-vector-icons", "color", "css-to-react-native"
        ]
      }
    },
    "server": {
      "detectDuplicateDependencies": {
        "ignoredModules": [
          "react-is"
        ]
      },
      "detectCyclicDependencies": false
    },
```

Been testing the config by just running rnx-bundle repeatedly. As you can see I had to add packages to the ignore array by hand, at which point I instead started getting cyclic dependency errors too, despite having disabled that.

I can spin up a repro repo if need be, but I took a quick look and found this:

https://github.com/microsoft/rnx-kit/blob/7200426250deb288fb38a3d8da91bb74a0e71437/packages/cli/src/metro-config.ts#L104-L113

So... am I misunderstanding something or are the plugins getting enabled as long as a boolean is provided? Regardless of whether it's true or false? Or is there another check further downstream on the value of the booleans?

tido64:
### Description

Plugins cannot be disabled because we only check whether `detectCyclicDependencies` and `detectDuplicateDependencies` are booleans.

Resolves #605.

### Test plan

Tests were added.

### DIFF
--- a/packages/cli/src/metro-config.ts
+++ b/packages/cli/src/metro-config.ts
@@ -102,15 +102,15 @@ export function customizeMetroConfig(
   const metroConfig = metroConfigReadonly as InputConfigT;
 
   const plugins: MetroPlugin[] = [];
-  if (typeof detectDuplicateDependencies === "boolean") {
-    plugins.push(DuplicateDependencies());
-  } else if (typeof detectDuplicateDependencies === "object") {
+  if (typeof detectDuplicateDependencies === "object") {
     plugins.push(DuplicateDependencies(detectDuplicateDependencies));
+  } else if (detectDuplicateDependencies !== false) {
+    plugins.push(DuplicateDependencies());
   }
-  if (typeof detectCyclicDependencies === "boolean") {
-    plugins.push(CyclicDependencies());
-  } else if (typeof detectCyclicDependencies === "object") {
+  if (typeof detectCyclicDependencies === "object") {
     plugins.push(CyclicDependencies(detectCyclicDependencies));
+  } else if (detectCyclicDependencies !== false) {
+    plugins.push(CyclicDependencies());
   }
 
   if (experimental_treeShake) {
--- a/change/@rnx-kit-cli-ee96f13f-0047-4cf8-8618-fb3f866285e9.json
+++ b/change/@rnx-kit-cli-ee96f13f-0047-4cf8-8618-fb3f866285e9.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix plugins cannot be disabled when bundling/serving",
+  "packageName": "@rnx-kit/cli",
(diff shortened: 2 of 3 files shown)


packages/cli/src/metro-config.ts
```diff
@@ -102,15 +102,15 @@ export function customizeMetroConfig(
   const metroConfig = metroConfigReadonly as InputConfigT;
 
   const plugins: MetroPlugin[] = [];
-  if (typeof detectDuplicateDependencies === "boolean") {
-    plugins.push(DuplicateDependencies());
-  } else if (typeof detectDuplicateDependencies === "object") {
+  if (typeof detectDuplicateDependencies === "object") {
     plugins.push(DuplicateDependencies(detectDuplicateDependencies));
+  } else if (detectDuplicateDependencies !== false) {
```
afoxman, line 107: What about null or undefined? Those won't be type "object" and they won't be false.

tido64: It looks like they are enabled by default here
https://github.com/microsoft/rnx-kit/blob/112ac9e9efdb88a5dcd8e87d8c2fe93d617d7d65/packages/config/src/getServerConfig.ts#L18-L19
and here
https://github.com/microsoft/rnx-kit/blob/f65f706c734841c4339ca1f825b9c9505cea8941/packages/config/src/getBundleDefinition.ts#L31-L32
So I assumed that it should be default on here as well. Should we inverse this here?

`typeof null` is actually `object` so we need to check for `null` as well…

afoxman: Good point. Forgot about the defaults which will override undefined. LGTM!

afoxman (approved):

PR #645, opened 2021-09-21
chore: enable `--isolatedModules` by default

tido64:
### Description

Enable isolated modules to ensure compatibility with transpilers that only operate on a single file at a time, e.g. Babel or esbuild.

See also https://www.typescriptlang.org/tsconfig#isolatedModules.

### Test plan

n/a


kelset (approved):
:shipit:

PR #666, opened 2021-09-24
feat(dep-check): add Metro related capabilities (7)

kelset:
### Description

Adds to the profiles the capabilities that are related to Metro that we use across rnx-kit (it's not all the ones defined in the [Metro monorepo](https://github.com/facebook/metro/tree/master/packages) but should cover most usages).

This is one of the few "core" capabilities that I could see missing still ;)

This PR will fail the dep-check slightly since it's targetting 0.65 still but some packages across rnx-kit are more "up to date" with the latest versions of metro packages.


packages/config/src/kitConfig.ts
```diff
@@ -20,6 +20,7 @@ export type Capability =
   | "html"
   | "lazy-index"
   | "masked-view"
+  | "metro-babel"
```
tido64, line 23: This should be `babel-preset-react-native` to follow Babel's naming convention. It's not tied to Metro, so we could potentially redirect to a different package in the future (e.g. [`@rnx-kit/babel-preset-react-native`](https://github.com/microsoft/rnx-kit/tree/main/packages/babel-preset-metro-react-native#readme)).

tido64:
> Dependencies are potentially misaligned – run `yarn rnx-dep-check` locally for more details
> 
> * https://github.com/microsoft/rnx-kit/blob/a2de7dffb14137e9cae1af0c98defb024c4ec20b/package.json#L46
> * https://github.com/microsoft/rnx-kit/blob/a2de7dffb14137e9cae1af0c98defb024c4ec20b/packages/babel-preset-metro-react-native/package.json#L27

You can ignore this. I'll take care of it.

packages/config/src/kitConfig.ts
kelset: It is tied to Metro in the sense that this comes directly from the Metro repo: https://github.com/facebook/metro/tree/master/packages/metro-react-native-babel-preset

and also I want to keep Metro in the name to underline how basically the version that this capability sets should also be reflected in any other metro-* dependencies someone might have (ex. `metro`).

Also, didn't want to have a long name. Alternatively, what about `metro-babel-preset`?

tido64: I'm not sure I follow. If we want metro related packages to be aligned, should we add a `metro` meta package instead? In general, capability names should be as generic when possible. In my opinion, if I just want a Babel preset for React Native, I wouldn't care where it came from. Since I'm using a specific profile, I would assume that the version declared is compatible with the other packages. From that perspective, I think `babel-preset-react-native` makes more sense.

kelset: ahhh gotcha I see what you mean. Do we already have any meta dependencies defined in the repo?

tido64: No, we currently only have a capability depending on another capability:

https://github.com/microsoft/rnx-kit/blob/49993ba6ee35872468166497cc60dc96339a2769/packages/dep-check/src/profiles/profile-0.64.ts#L20-L24

For a meta package, set `name` to `#meta` and drop `version`, e.g.:

```js
{
  "metro/all": {
    name: "#meta",
    capabilities: [
      "babel-preset-react-native",
      "metro",
    ],
  },
}
```

Alternatively:

```js
{
  "metro": {
    name: "metro",
    version: "^0.64.0", 
    capabilities: ["babel-preset-react-native"],
  },
}
```

packages/dep-check/src/profiles/profile-0.61.ts
```diff
@@ -75,6 +75,14 @@ const profile: Profile = {
...
     version: "^0.2.4",
   },
+  "babel-preset-react-native": {
+    name: "metro-react-native-babel-preset",
+    version: "^0.56.0",
+  },
+  metro: {
+    name: "metro",
+    version: "^0.56.0",
+  },
```
tido64, line 85: Can you add `devOnly: true` so these don't get propagated to the consuming apps?
```suggestion
  "babel-preset-react-native": {
    name: "metro-react-native-babel-preset",
    version: "^0.56.0",
    devOnly: true,
  },
  metro: {
    name: "metro",
    version: "^0.56.0",
    devOnly: true,
  },
```

kelset: sure, will do!

tido64 (approved):

tido64:
> Dependencies are potentially misaligned – run `yarn rnx-dep-check` locally for more details
> 
> * https://github.com/microsoft/rnx-kit/blob/467ba52e61230f1d25d645efbb7f12bdc0e05d07/package.json#L43-L49
> * https://github.com/microsoft/rnx-kit/blob/467ba52e61230f1d25d645efbb7f12bdc0e05d07/packages/metro-resolver-symlinks/package.json#L30
> * https://github.com/microsoft/rnx-kit/blob/467ba52e61230f1d25d645efbb7f12bdc0e05d07/packages/metro-serializer-esbuild/package.json#L36
> * https://github.com/microsoft/rnx-kit/blob/467ba52e61230f1d25d645efbb7f12bdc0e05d07/packages/metro-serializer/package.json#L31
> * https://github.com/microsoft/rnx-kit/blob/467ba52e61230f1d25d645efbb7f12bdc0e05d07/packages/metro-service/package.json#L47-L52

These should go away once we move to `react-native` 0.66. Which should happen any time soon now…

PR #671, opened 2021-09-27
feat(dep-check): tweaks to react-native capability, adding jest and add first meta capability to dep-check

kelset:
### Description

Now that I think I understand a bit better how the whole gist works, I'm tweaking a bit more the capabilities.
* adding react as "capability" to core so that we make sure to always have the right React/RN combo
* make sure also macos gets the right version of react
* adds a meta package for testing related things
	* and adds Jest as capability
	
And of course necessary clean up of tests and readme to align with this.


packages/dep-check/src/profiles/profile-0.64.ts
```diff
@@ -1,11 +1,14 @@
...
-
-const reactNative = {
-  name: "react-native",
-  version: "^0.64.2",
-};
+import profile63 from "./profile-0.63";
+import { reactNativeMeta } from "./profile-0.61";
 
 const profile: Profile = {
+  ...profile63,
```
kelset, line 6: I've also added this to 0.64 because all the other packages had it - so I felt like the loop of "depending on preceding version" was broken without this. Please let me know if there was any deliberate reason around this

packages/dep-check/src/profiles/profile-0.66.ts
```diff
@@ -64,10 +66,6 @@ const profile: Profile = {
     version: "^0.66.2",
     devOnly: true,
   },
-  react: {
-    name: "react",
-    version: "17.0.2",
-  },
```
kelset, line 70: this is not really needed since we get it from the 0.65 profile

packages/dep-check/src/profiles/profile-0.61.ts
```diff
@@ -1,11 +1,16 @@
-import type { Profile } from "../types";
+import type { Profile, MetaPackage } from "../types";
 
-const reactNative = {
-  name: "react-native",
-  version: "^0.61.5",
+export const reactNativeMeta: MetaPackage = {
+  name: "#meta",
+  capabilities: ["react-native", "react", "react-test-renderer"],
 };
```
kelset, line 6: potentially we could "isolate" this meta into its own file ex. `profiles/meta-packages.ts`, let me know what you think

tido64: Nah, I think this is fine.

packages/dep-check/src/profiles/profile-0.66.ts
tido64: I added it explicitly so it's clear that it's using the same version, and not because we forgot to add it for some reason 😛

packages/dep-check/src/profiles/profile-0.61.ts
```diff
@@ -1,11 +1,16 @@
-import type { Profile } from "../types";
+import type { Profile, MetaPackage } from "../types";
 
-const reactNative = {
-  name: "react-native",
-  version: "^0.61.5",
+export const reactNativeMeta: MetaPackage = {
+  name: "#meta",
+  capabilities: ["react-native", "react", "react-test-renderer"],
```
tido64, line 5: I don't agree with this. Testing should be separate from `core`. Not all packages need `react-test-renderer`, e.g. if you're maintaining a non-UI facing native module like `react-native-netinfo`. I also don't like that we're introducing a fourth way of installing `react-native` with its own capability. The thought behind `core` is that it would resolve to shared `react-native` code, and `react-native-*` would resolve to platform specific packages, i.e. like `react-native` vs `react-native-windows`. Today, both Android and iOS live in core so it's a bit muddled, but it's not unlikely that we could see them be pulled out of core as part of the Lean Core Effort.

I think having a separate testing capability, e.g. `core/testing`, that pulls in `jest`, `react-test-renderer`, etc. makes more sense. Is that something you've considered?

packages/dep-check/src/profiles/profile-0.66.ts
kelset: could I add this as an information as a comment at the top (where the ...profile-0.65 import lives) instead?

packages/dep-check/src/profiles/profile-0.61.ts
kelset: two points:
1) sure, no problem removing `react-test-renderer` and creating a separate meta package for testing
2) the only reason I added "react-native" as its own capability is that I need it for the meta package... maybe I just missed it, but is there a way to define it directly react-native and its version within the meta definition?

packages/dep-check/src/profiles/profile-0.66.ts
kelset: or if you prefer I can just have react/react-dom/react-test-renderer redefined as if they were a different verison

packages/dep-check/src/profiles/profile-0.61.ts
tido64: I don't think you need the testing capability to depend on `react-native`, but I would probably do something like:

```js
{
  "core/test": {
    name: "#meta",
    capabilities: ["core", "jest", "react-test-renderer"]
  }
}
```

Or if you want to be explicit about `react-native`:

```js
{
  "core/test": {
    name: "react-native",
    version: "^0.64.0",
    capabilities: ["jest", "react-test-renderer"],
    devOnly: true,
  }
}
```

packages/dep-check/src/profiles/profile-0.66.ts
tido64: I'd prefer to have them redefined to be explicit. It also keeps everything in the same space so you don't have to go to another profile to find out which version you're on.

packages/dep-check/src/profiles/profile-0.61.ts
kelset: allright, I think I've changed this to be consistent with your feedback now :)

packages/dep-check/README.md
```diff
@@ -135,13 +135,15 @@ resolve to:
 | base64                               | react-native-base64@^0.2.1                        | react-native-base64@^0.2.1                        | react-native-base64@^0.2.1                    | react-native-base64@^0.2.1                    | react-native-base64@^0.2.1                    |
 | checkbox                             | @react-native-community/checkbox@^0.5.8           | @react-native-community/checkbox@^0.5.8           | @react-native-community/checkbox@^0.5.7       | @react-native-community/checkbox@^0.5.7       | @react-native-community/checkbox@^0.5.7       |
 | clipboard                            | @react-native-clipboard/clipboard@^1.8.4          | @react-native-clipboard/clipboard@^1.7.3          | @react-native-community/clipboard@^1.5.1      | @react-native-community/clipboard@^1.5.1      | @react-native-community/clipboard@^1.5.1      |
+| core/testing                         | #meta@undefined                                   | #meta@undefined                                   | #meta@undefined                               | #meta@undefined                               | #meta@undefined                               |
```
kelset, line 138: looks like the update-readme script doesn't like meta packages 🤣 what should we do about it @tido64?

tido64: I fixed the script in #680

packages/config/src/kitConfig.ts
```diff
@@ -20,6 +23,7 @@ export type Capability =
   | "hooks"
   | "html"
   | "lazy-index"
+  | "jest"
```
tido64, line 26:
```suggestion
  | "html"
  | "jest"
  | "lazy-index"
```

tido64 (approved):

PR #682, opened 2021-09-29
chore(dep-check): bump a couple of RN versions to latest patch

kelset:
### Description

A couple of RN versions were not pointing to latest available, which in some scenarios might lead to baad behaviours when ex. the native side pick sup the version from the package.json but doesn't understand semver so it complains (ex. 0.63.2 v 0.63.4 that we are hitting in a project right now).

For some reason once again I can't seem to get the update-script to update the README, so @tido64 do you think you'd be able to do it? 😓


kelset:
I don't think we need this anymore

PR #730, opened 2021-10-15
feat(react-native-test-app-msal): MSAL module for react-native-test-app

tido64:
### Description

Extracted MSAL support module from our internal repo.

### Remaining work

- [ ] Android: JetPack Compose?
- [x] iOS
  - [x] ~~Clear out cookies from the WebView used for auth~~ We can wipe out the account, but not the cookies it looks like. Something to investigate for later.
  - [x] https://github.com/microsoft/react-native-test-app/pull/560
- [x] macOS: Since the switcher is written in SwiftUI, it just needs some adjustments/testing
- [ ] Windows: This will be a new implementation since we don't have one from before
- [ ] Documentation

### Test plan

| Account switcher | List of accounts from MSAL | Adding a new account |
|-|-|-|
| Simulator Screen Shot - iPhone 12 Pro - 2021-10-15 at 17 06 04 | Simulator Screen Shot - iPhone 12 Pro - 2021-10-15 at 17 07 09 | Simulator Screen Shot - iPhone 12 Pro - 2021-10-15 at 17 08 29 |


afoxman:
If this is a module for RNTA, should it be developed and published from that repo instead of rnx-kit?

packages/react-native-test-app-msal/ios/ReactTestApp-MSAL.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1 @@
+/* Dummy file so @react-native-community/cli recognizes this as an iOS package */
```
kelset, line 1: 🤣

kelset (approved):
I think it's ok to have this package live in rnx-kit since it fits the "range" of plugins and tools we create for RN experiences. It just so happens that this one is specifically for RNTA

packages/react-native-test-app-msal/ios/ReactTestApp-MSAL.xcodeproj/project.pbxproj
tido64: Yeah, it's on my wishlist to the cli team that they should look for `Podfile`.

PR #771, opened 2021-10-27
feat: Metro transform worker that uses swc under the hood

tido64:
### Description

Introducing an experimental Metro transformer that uses [swc](https://swc.rs/) instead of Babel for files that are not assets or written in Flow.

### Test plan

1. Configure Metro to use `@rnx-kit/metro-swc-worker`:
   ```diff
   diff --git a/packages/test-app/babel.config.js b/packages/test-app/babel.config.js
   index e1a82f0c..caa760d8 100644
   --- a/packages/test-app/babel.config.js
   +++ b/packages/test-app/babel.config.js
   @@ -1,3 +1,7 @@
    module.exports = {
   -  presets: ["@rnx-kit/babel-preset-metro-react-native"],
   +  presets: [
   +    ["@rnx-kit/babel-preset-metro-react-native", {
   +      disableImportExportTransform: true,
   +    }],
   +  ],
    };
   diff --git a/packages/test-app/metro.config.js b/packages/test-app/metro.config.js
   index 233a4937..6893aadd 100644
   --- a/packages/test-app/metro.config.js
   +++ b/packages/test-app/metro.config.js
   @@ -1,14 +1,22 @@
    const { exclusionList, makeMetroConfig } = require("@rnx-kit/metro-config");
    const MetroSymlinksResolver = require("@rnx-kit/metro-resolver-symlinks");
   +const {
   +  MetroSerializer,
   +  esbuildTransformerConfig,
   +} = require("@rnx-kit/metro-serializer-esbuild");

    // Metro will pick up react-native-macos/-windows mocks if we don't exclude them
    const blockList = exclusionList([/.*__fixtures__.*/]);

    module.exports = makeMetroConfig({
   -  projectRoot: __dirname + "/src",
      resolver: {
        blacklistRE: blockList,
        blockList,
        resolveRequest: MetroSymlinksResolver(),
      },
   +  serializer: {
   +    customSerializer: MetroSerializer([], { target: "es2015" }),
   +  },
   +  transformer: esbuildTransformerConfig,
   +  transformerPath: require.resolve("@rnx-kit/metro-swc-worker"),
    });
   ```
2. Bundle: `yarn react-native bundle --entry-file src/index.ts --platform ios --dev false --bundle-output dist/main.ios.jsbundle --sourcemap-output dist/main.jsbundle.map --assets-dest dist --reset-cache`

image


packages/metro-swc-worker/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export { getCacheKey } from "metro-transform-worker";
+
+const countLines = (() => {
+  const reLine = /^/gm;
+  return (str: string): number => str.match(reLine)?.length ?? 0;
+})();
+
+function isAsset(filename: string, options: JsTransformOptions): boolean {
+  return filename.endsWith(".json") || options.type === "asset";
```
afoxman, line 17: Should this include common multimedia file extensions, like mpg and jpg? There's a list of supported react-native asset types floating around in a regular expression somewhere in haul or react-native.

packages/metro-swc-worker/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  options
+) => {
+  if (config.minifierPath !== esbuildTransformerConfig.minifierPath) {
+    const { name } = require("../package.json");
+    throw new Error(
+      `'${name}' requires '@rnx-kit/metro-serializer-esbuild' and it doesn't look like it's correctly set up. Please verify your 'metro.config.js' and try again.`
+    );
+  }
+
+  if (isAsset(filename, options) || isFlow(filename, data)) {
```
afoxman, line 44: Might be less maintenance to say "if not TS/JS, send to metro" since those file extensions are finite and well-known.

afoxman (approved):

packages/metro-swc-worker/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "main": "lib/index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/rnx-kit",
+    "directory": "packages/metro-swc-worker"
+  },
+  "scripts": {
+    "build": "rnx-kit-scripts build",
+    "format": "rnx-kit-scripts format",
+    "lint": "rnx-kit-scripts lint"
```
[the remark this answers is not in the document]
tido64: There are currently no tests for this package. This would fail during build.

packages/metro-swc-worker/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+import { transform as swcTransform } from "@swc/core";
+import {
+  JsTransformOptions,
+  transform as metroTransform,
+} from "metro-transform-worker";
+import DependencyCollector from "./DependencyCollector";
+
+export { getCacheKey } from "metro-transform-worker";
+
+const countLines = (() => {
```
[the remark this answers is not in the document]
tido64: This one is a bit weird, but it has some state (a regex), similar to a static variable inside a function. I call it immediately a few lines down, and it returns a function that captures the state. This way it cannot be accessed from outside.

packages/metro-swc-worker/src/index.ts
tido64: These two checks are what Metro does internally. I think the multimedia would be `asset` types.

packages/metro-swc-worker/src/index.ts
tido64: It's not as simple since Flow also uses `.js` extension. You need to look for `@flow` or `@noflow`, but sometimes they are omitted as well. If you look at the implementation of `isFlow`, we just assume react-native is all Flow.

packages/metro-swc-worker/src/index.ts
afoxman: Got it... that makes sense.

packages/metro-swc-worker/src/index.ts
afoxman: I did see that... well, with the other comment resolved re: not need to maintain a huge list of extensions, this is looking like a good choice.

packages/metro-swc-worker/src/index.ts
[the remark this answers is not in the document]
afoxman: Oh yeah, i see it now. It's a factory function with embedded state to avoid recreating the regex on each call.

PR #784, opened 2021-10-30
Add new tools methods

afoxman:
### Description

Add new functions to `@rnx-kit/tools-language` and `@rnx-kit/tools-node`. These will be used in upcoming PRs. I've separated them to make each PR smaller and more focused.

### Test plan

New tests were added for the new functions.

### DIFF
--- a/packages/tools-language/test/array.test.ts
+++ b/packages/tools-language/test/array.test.ts
@@ -1,0 +1,95 @@
+import "jest-extended";
+import { toIndex, addRange } from "../src/array";
+
+describe("Language > Array > toIndex()", () => {
+  const testArray = ["a", "b", "c"];
+
+  test("index matches offset when offset is zero", () => {
+    expect(toIndex(testArray, 0)).toEqual(0);
+  });
+
+  test("index matches offset when offset is positive", () => {
+    expect(toIndex(testArray, 2)).toEqual(2);
+  });
+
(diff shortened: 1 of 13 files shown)


packages/tools-node/test/fs.test.ts
```diff
@@ -78,4 +84,39 @@ describe("Node > FS", () => {
     // check for rw|r|r rather than rwx|rx|rx since 'x' doesn't seem to come through on Windows
     expect(stats.mode & 0o644).toEqual(0o644);
   });
+
+  test("statSync() succeeds when given a valid path", () => {
+    const stats = statSync(fixtureDir);
+    expect(stats).not.toBeNil();
```
tido64, line 90: What's the difference between `toBeNil` and Jest's [`toBeNull`](https://jestjs.io/docs/using-matchers#truthiness)?

packages/tools-language/test/array.test.ts
```diff
@@ -0,0 +1,95 @@
+import "jest-extended";
```
tido64, line 1: Do we need this here? It doesn't look like you're using it.

packages/tools-node/test/fs.test.ts
afoxman: toBeNil checks for undefined and null.

tido64 (approved):

PR #794, opened 2021-11-01
New package: @rnx-kit/typescript-react-native-resolver

afoxman:
### Description

Add a new package `@rnx-kit/typescript-react-native-resolver` to provide react-native module resolution to TypeScript.

It implements NodeJs resolution and adds support for react-native platform extensions, as well as being able to resolve `react-native` references to out-of-tree platforms like `react-native-windows` or `react-native-macos`.

This PR includes all of the resolver code, a full suite of tests, and documentation.

My next PR will integrate this package with TypeScript using TS APIs to produce `rn-tsc`, a replacement for `tsc`.

### Test plan

CI tests pass, including the new tests added with this package.


packages/typescript-react-native-resolver/src/module.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  const { host } = context;
+
+  //
+  //  See if the module path has an extension that is in the list. If it
+  //  does, see if the path refers to a real file. Either way, return the
+  //  result and stop here.
+  //
+  const extension = getExtensionFromPath(modulePath);
+  if (extension && extensions.includes(extension)) {
```
tido64, line 91:
```suggestion
  if (extensions?.includes(extension)) {
```

packages/typescript-react-native-resolver/src/resolve.ts
```diff
@@ -0,0 +1,205 @@
...
+  //  No path was given. Try resolving the module using package.json
+  //  properties.
+  log.log("Reading package.json from directory %o.", packageDir);
+  const result = host.readFile(path.join(packageDir, "package.json"));
+  if (!result) {
+    throw new Error(
+      `Failed to read package.json from directory '${packageDir}'`
+    );
+  }
+  const { types, typings, main } = JSON.parse(result);
```
tido64, line 53: Is it possible to configure which fields should be considered for resolving the main file? In Metro, we would set [`resolverMainFields`](https://facebook.github.io/metro/docs/configuration/#resolvermainfields). For instances, packages that use Apollo would prefer `module` over `main`.

packages/typescript-react-native-resolver/src/react-native-package-name.ts
```diff
@@ -0,0 +1,55 @@
+import type { ResolverLog } from "./log";
+
+/**
+ * Get the name of an out-of-tree platform's react-native package.
+ *
+ * @param platform Platform
+ * @returns Name of the out-of-tree platform's react-native package, or `undefined` if it is in-tree or unknown.
+ */
+export function getReactNativePlatformPackageName(
```
tido64, line 9: We should get these values from `react-native config` instead. There are more platforms than listed here.

packages/typescript-react-native-resolver/src/resolve.ts
```diff
@@ -0,0 +1,205 @@
...
+      module = findModuleFile(context, packageDir, types, extensions);
+    } else if (typeof typings === "string" && typings.length > 0) {
+      log.log("Package has 'typings' field '%s'.", typings);
+      module = findModuleFile(context, packageDir, typings, extensions);
+    }
+  }
+  if (!module && typeof main === "string" && main.length > 0) {
+    log.log("Package has 'main' field '%s'.", main);
+    module = findModuleFile(context, packageDir, main, extensions);
+  }
```
tido64, line 69: If we need to look at more fields, I think something like this would be better:
```suggestion
  const manifest = JSON.parse(result);

  [
    // Only consult 'types' and 'typings' properties when looking for
    // type files (.d.ts).
    ...(extensions.includes(ts.Extension.Dts) ? ["types", "typings"] : []),
    "main",
  ].find((field) => {
    const modulePath = manifest[field];
    if (typeof modulePath === "string" && modulePath.length > 0) {
      log.log("Package has '%s' field '%s'.", field, modulePath);
      module = findModuleFile(context, packageDir, modulePath, extensions);
      return Boolean(module);
    }
    return false;
  });
```

packages/typescript-react-native-resolver/src/resolve.ts
```diff
@@ -0,0 +1,205 @@
...
+  while (true) {
+    const p = path.join(searchDir, suffixDir);
+    if (host.directoryExists(p)) {
+      return p;
+    }
+    if (searchDir === searchRoot) {
+      return undefined;
+    }
+    searchDir = path.dirname(searchDir);
+  }
```
tido64, line 108: We can avoid the linter warning by comparing against root in while condition:
```suggestion
  while (searchDir !== searchRoot) {
    const p = path.join(searchDir, suffixDir);
    if (host.directoryExists(p)) {
      return p;
    }
    searchDir = path.dirname(searchDir);
  }
  return undefined;
```

packages/typescript-react-native-resolver/src/resolve.ts
```diff
@@ -0,0 +1,205 @@
...
+  let module: ts.ResolvedModuleFull | undefined = undefined;
+
+  // Resolve the module to a file within the package
+  log.log(
+    "Searching for external package %o starting in %o.",
+    moduleRef.scope ? moduleRef.scope + "/" + moduleRef.name : moduleRef.name,
+    searchDir
+  );
+  const pkgDir = findPackageDependencyDir(context, moduleRef, searchDir);
+  // TODO: stopDir ==> workspace root? security & perf
```
tido64, line 144: Do we need to create an issue for this?

packages/typescript-react-native-resolver/test/extension.test.ts
```diff
@@ -0,0 +1,34 @@
+import ts from "typescript";
+import { hasExtension, getExtensionFromPath } from "../src/extension";
```
tido64, line 2: Can you also add file names that don't have an extension, e.g. `foo-ts`, `bar-jsx`.

packages/typescript-react-native-resolver/test/host.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  function testChangeCompilerHost(host: ts.CompilerHost): void {
+    changeCompilerHostToUseReactNativeResolver(
+      host,
+      {}, // compiler options
+      "ios",
+      ["native"],
+      false, // enable RN substitution?
+      false, // enable error-tracing?
+      undefined // trace file
+    );
```
tido64, line 26: If we have so many parameters that we need to annotate them, maybe it makes more sense to pass an object so we can use named fields?

tido64 (commented):
First round of comments. I might have to do a second round since there's so much… 😛

change/@rnx-kit-typescript-react-native-resolver-23ea8f2c-96b0-4cac-bb3a-ef6e42d52814.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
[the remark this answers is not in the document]
afoxman: It's a new package -- there's no previous version to measure against.

packages/typescript-react-native-resolver/src/extension.ts
```diff
@@ -0,0 +1,24 @@
...
+];
+
+export const ExtensionsTypeScript = [
+  ts.Extension.Ts,
+  ts.Extension.Tsx,
+  ts.Extension.Dts,
+];
+
+export function hasExtension(p: string, ext: ts.Extension): boolean {
+  return p.length > ext.length && p.endsWith(ext);
```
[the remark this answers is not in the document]
afoxman: ts.Extension does include the '.' prefix, so it wouldn't match "something-something-ts" to any extension. path.extname() can't be used because declaration files (enum Dts here) are ".d.ts" and it would miss the ".d".

packages/typescript-react-native-resolver/src/extension.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export function getExtensionFromPath(p: string): ts.Extension | undefined {
+  return Extensions.find((e) => hasExtension(p, e));
```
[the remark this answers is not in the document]
afoxman: I can't use path.extname() so there isn't really a way to extract.

In order to find out which extension is present, if any, I need to compare the end of the path with each known extension. See how Extension.Dts is first in the list? That's a good example of what makes this tricky -- you don't want to match .ts unless you are sure you aren't looking at .d.ts.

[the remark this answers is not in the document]
afoxman: I'm leaving this unresolved in case I've missed something and there's a clever way to do this...

packages/typescript-react-native-resolver/src/log.ts
```diff
@@ -0,0 +1,249 @@
...
+  // ignore JSON module failures when TypeScript is configured to ignore them
+  if (!options.resolveJsonModule) {
+    if (path.extname(moduleName).match(/\.json$/i)) {
+      return false;
+    }
+  }
+
+  // ignore resolver errors for multimedia files
+  const multimediaExts =
+    /\.(aac|aiff|bmp|caf|gif|html|jpeg|jpg|m4a|m4v|mov|mp3|mp4|mpeg|mpg|obj|otf|pdf|png|psd|svg|ttf|wav|webm|webp)$/i;
```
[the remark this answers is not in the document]
afoxman: Since this is for suppressing resolver errors in the log, I'd say the list is an approximation and doesn't need to align with Metro or Haul (which is where I got this). The extensions here are all legitimate asset file types which we wouldn't care about when investigating resolution failures. Even if the list is different from what Metro has.

packages/typescript-react-native-resolver/src/log.ts
```diff
@@ -0,0 +1,249 @@
...
+ * @param containingFile File from which the module was required/imported.
+ */
+export function logModuleBegin(
+  log: ResolverLog,
+  moduleName: string,
+  containingFile: string
+): void {
+  log.begin();
+  log.log(
+    "======== Resolving module '%s' from '%s' ========",
```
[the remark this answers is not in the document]
afoxman: I tried to make it look and similar to the messages in TypeScript's module resolver. Since this code replaces TypeScript's resolver, I lose their messaging and have to create my own. There are no low-level TS resolver functions I can pull in to reuse (and therefore get the logging that is built into them).

There is a shared "host" interface between my code and TypeScript, and it has a trace() function. I attached my logger to that function, and though I replaced the resolver, there are still other messages that TypeScript does emit through it. That's about as close as I can get to leveraging TS logging.

packages/typescript-react-native-resolver/src/module.ts
afoxman: `extension` !== `extensions`, so this wouldn't work. :)

packages/typescript-react-native-resolver/src/react-native-package-name.ts
afoxman: I had been doing that... but it is SLOOOOOOOOOW. Like really slow, even if I only call it once. It scans all the dependencies looking for RN CLI configs. I had to take it out because performance was horrid.

afoxman: I looked at what our internal repos were doing, and they have also hard-coded this way.

packages/typescript-react-native-resolver/src/resolve.ts
afoxman: I could add this if it was needed, but for the first cut, I'd like to stick with these fields since they are well-known and baked into TypeScript as well. I've been through the code and haven't seen "module" used anywhere in the resolver.

packages/typescript-react-native-resolver/src/resolve.ts
afoxman: Makes sense. Adding more fields would benefit from rolling up into a loop.

packages/typescript-react-native-resolver/src/resolve.ts
afoxman: It's not quite the same thing. `while (true)` will test for existence at the root, where your changes will skip the root.

packages/typescript-react-native-resolver/test/extension.test.ts
afoxman: Line 27 has a case for that

packages/typescript-react-native-resolver/src/resolve.ts
afoxman: Oh, yes. I meant to do that and remove the comment, but I've been working on this on and off for a few weeks, and I forgot.

afoxman: I removed two TODOs and filed #801.

packages/typescript-react-native-resolver/test/host.test.ts
afoxman: Yes, this is a great idea. Fixed.

change/@rnx-kit-typescript-react-native-resolver-23ea8f2c-96b0-4cac-bb3a-ef6e42d52814.json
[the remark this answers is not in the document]
tido64: I'm thinking more of what Beachball does with the version number of this package. In this case, it would bump it to 0.0.2. If that's what we want, then it's fine.

packages/typescript-react-native-resolver/src/extension.ts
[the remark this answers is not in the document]
tido64: Ah, I had missed that case. Then we should remove the length check:
```suggestion
  return p.endsWith(ext);
```

packages/typescript-react-native-resolver/src/extension.ts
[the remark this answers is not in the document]
tido64: You're right. I missed the `.d.ts` extension. I still think you can simplify `hasExtension()` though.

packages/typescript-react-native-resolver/src/log.ts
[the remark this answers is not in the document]
tido64: Would it make sense to invert the list instead? How many valid extensions are there?

packages/typescript-react-native-resolver/src/module.ts
tido64: 🥸

packages/typescript-react-native-resolver/src/resolve.ts
tido64: We've hit this issue internally just last week, actually. Apollo refactored a bunch of code and added the `module` field for ESM modules. Our bundles increased by 0.5 MB when we bumped to the latest version. I'm fine with leaving this as an initial implementation, but I think we should add support for more fields.

packages/typescript-react-native-resolver/src/log.ts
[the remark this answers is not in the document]
afoxman: The valid extensions are baked into the resolver (.ts, .tsx, .js, .jsx, .json). Anything else ends up down here. The approach is "I want to see errors, except for these specific types which I'm sure aren't interesting...". The list will grow as we bump into more non-interesting types, but I didn't want to ignore things by default -- i wanted it explicit.

packages/typescript-react-native-resolver/src/resolve.ts
afoxman: Tracked by #805.

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -0,0 +1,265 @@
...
+  const searchDir = path.dirname(containingFile);
+  if (isPackageModuleRef(moduleRef)) {
+    module = resolvePackageModule(context, moduleRef, searchDir, extensions);
+  } else if (isFileModuleRef(moduleRef)) {
+    module = resolveFileModule(context, moduleRef, searchDir, extensions);
+  }
+  if (module) {
+    module.isExternalLibraryImport = !!module.resolvedFileName.match(
+      /[/\\]node_modules[/\\]/
+    );
```
tido64, line 135: [Nit] There's a `test()` function on the RegExp object that we should use instead:
```suggestion
    module.isExternalLibraryImport =
      /[/\\]node_modules[/\\]/.test(module.resolvedFileName);
```

tido64 (approved):

PR #795, opened 2021-11-01
feat(eslint-plugin): implemented fixer for no-export-all

tido64:
### Description

Implemented a fixer for the `no-export-all` rule.

### Test plan

Tests were added.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-export-all.js
+++ b/packages/eslint-plugin/src/rules/no-export-all.js
@@ -3,7 +3,18 @@
 
 /**
  * @typedef {import("@typescript-eslint/types").TSESTree.Node} Node
+ * @typedef {import("eslint").Rule.RuleContext} ESLintRuleContext
  * @typedef {{ exports: string[], types: string[] }} NamedExports
+ *
+ * @typedef {{
+ *   id: ESLintRuleContext["id"];
+ *   options: ESLintRuleContext["options"];
+ *   settings: ESLintRuleContext["settings"];
+ *   parserPath: ESLintRuleContext["parserPath"];
+ *   parserOptions: ESLintRuleContext["parserOptions"];
+ *   parserServices: ESLintRuleContext["parserServices"];
+ *   filename: string;
+ * }} RuleContext
  */
 
 const fs = require("fs");
@@ -22,6 +33,8 @@ const DEFAULT_CONFIG = {
   eslintScopeManager: true,
 };
 
+const MAX_DEPTH = 5;
+
 /**
  * Returns whether there are any named exports.
  * @param {NamedExports?} namedExports
@@ -36,7 +49,7 @@ function isEmpty(namedExports) {
 
 /**
  * Creates and returns an ES tree traverser.
- * @returns {{ break: () => void; traverse: (node: Node, options: {}) => void}}
+ * @returns {{ traverse: (node: Node, options: {}) => void; }}
  */
 function makeTraverser() {
   const Traverser = require(path.join(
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin/src/rules/no-export-all.js
```diff
@@ -1,6 +1,125 @@
...
+const fs = require("fs");
+const path = require("path");
+
+/**
+ * Parses specified file and returns an AST.
+ * @param {import("eslint").Rule.RuleContext} context
+ * @param {string} moduleName
+ */
+function parse(context, moduleName) {
+  const { parse } = require(context.parserPath);
```
afoxman, line 17: Is there an alternative name you can choose to differentiate between the running function and the import? I know it works, but it's a little tricky to spot.

packages/eslint-plugin/src/rules/no-export-all.js
```diff
@@ -1,6 +1,125 @@
...
+ * @typedef {import("@typescript-eslint/types").TSESTree.Node} Node
+ */
+
+const fs = require("fs");
+const path = require("path");
+
+/**
+ * Parses specified file and returns an AST.
+ * @param {import("eslint").Rule.RuleContext} context
+ * @param {string} moduleName
```
afoxman, line 14: @returns?

afoxman (approved):
This seems like a good candidate for the public distro of eslint. What do you think?

tido64:
> This seems like a good candidate for the public distro of eslint. What do you think?

Yeah, I think that's what the TypeScript ESLint folks suggested to @dzearing as well. I think it makes sense, and I will try to upstream it, but I want to apply this in FURN first to weed out any issues.

dzearing:
nice work! I will give it a go later today and see it in action.

tido64:
> nice work! I will give it a go later today and see it in action.

I've added support for nested `export *`. Here's a sample diff from fluentui-react-native:

```diff
diff --git a/packages/theming/win32-theme/src/NativeModule/index.ts b/packages/theming/win32-theme/src/NativeModule/index.ts
index 6f8de92f..cf05c96a 100644
--- a/packages/theming/win32-theme/src/NativeModule/index.ts
+++ b/packages/theming/win32-theme/src/NativeModule/index.ts
@@ -1,3 +1,3 @@
-export * from './fallbackOfficeModule';
-export * from './getThemingModule';
-export * from './officeThemingModule';
+export { fallbackGetPalette, fallbackOfficeModule } from './fallbackOfficeModule';
+export { getThemingModule } from './getThemingModule';
+export type { CxxException, IEventEmitter, NativeColorNames, NativeColorRamps, OfficeThemingModule, PlatformDefaultsChangedArgs, PlatformDefaultsChangedCallback } from './officeThemingModule';
diff --git a/packages/theming/win32-theme/src/index.ts b/packages/theming/win32-theme/src/index.ts
index bde5e196..458283f4 100644
--- a/packages/theming/win32-theme/src/index.ts
+++ b/packages/theming/win32-theme/src/index.ts
@@ -1,6 +1,7 @@
-export * from './createOfficeTheme';
-export * from './createPartialOfficeTheme';
-export * from './NativeModule/index';
-export * from './paletteFromOfficeColors';
-export * from './createOfficeAliasTokens';
-export * from './createBrandedThemeWithAlias';
+export { createOfficeTheme } from './createOfficeTheme';
+export { createPartialOfficeTheme } from './createPartialOfficeTheme';
+export { fallbackGetPalette, fallbackOfficeModule, getThemingModule } from './NativeModule/index';
+export type { CxxException, IEventEmitter, NativeColorNames, NativeColorRamps, OfficeThemingModule, PlatformDefaultsChangedArgs, PlatformDefaultsChangedCallback } from './NativeModule/index';
+export { paletteFromOfficeColors } from './paletteFromOfficeColors';
+export { createOfficeAliasTokens } from './createOfficeAliasTokens';
+export { createBrandedThemeWithAlias, getCurrentBrandAliasTokens } from './createBrandedThemeWithAlias';
```

dzearing (approved):
awesome!

PR #800, opened 2021-11-02
chore: add a fake package to use as template for creating new packages

kelset:
### Description

As per title, I want to add a `template` package that we can use as a baseline of what every new package should be and have.

I've also added a dedicated new script `scripts/new-package.js` to generate new packages; it's fairly simple, but - fun fact! - I used GH Copilot to write it! (really, aside from a few tweaks and adding `fs-extra` for the easy folder copy, it's all autogenerated 🤯)

### Test plan

Simply run

```sh
yarn new-package newPackageName
```

To generate a sample project for you to use.


change/@rnx-kit-template-9d95aaf4-b824-4537-bfff-0c8047ee2d89.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
kelset, line 2: I guess we can agree that the first change entry for a new package should be of type `none`?

packages/template/README.md
```diff
@@ -0,0 +1,31 @@
...
+wants to start introducing something in that direction.
+
+## Usage
+
+Copy/paste this folder and change all the relevant files to match your new
+addition (so, change name of folder, in the package.json, etc).
+
+It's not ideal but at least it's a start.
+
+### What does every file do?
```
kelset, line 21: Is this section useful? I'd personally like it as a quick way for people to learn what they need to use

packages/template/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+- `src` folder - contains two files bla bla bla
+  - `index.ts` bla bla bla
+  - `types.ts` bla bla bla
+- `CHANGELOG.json` bla bla bla
+- `CHANGELOG.md` bla bla bla
+- `just.config.js` bla bla bla
```
kelset, line 28: ok so @tido64 can you tell me what I should write for the just file, I honestly still don't understand what's the deal with this file 😅

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "name": "@rnx-kit/template",
+  "version": "0.0.1",
+  "description": "Template folder to use as baseline for new packages in rnx-kit",
+  "homepage": "https://github.com/microsoft/rnx-kit/tree/main/packages/template#readme",
+  "license": "MIT",
+  "files": [
+    "src/*"
+  ],
+  "main": "src/index.ts",
+  "types": "src/types.ts",
```
kelset, line 11: do we even want the `main` and `types` entries? 🤔
or should be more of an opt in thing?

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "url": "https://github.com/microsoft/rnx-kit",
+    "directory": "packages/template"
+  },
+  "engines": {
+    "node": ">= 14"
+  },
+  "scripts": {
+    "build": "tsc",
+    "format": "prettier --write src/",
+    "lint": "eslint src/* --ext .js,.jsx,.ts,.tsx --config package.json --color"
```
kelset, line 23: any other default scripts I might be missing?

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "dependencies": {
+    "typescript": "^4.0.0",
+    "workspace-tools": "^0.16.2"
```
kelset, line 27: is `workspace-tools` actually needed?

packages/template/package.json
kelset: ex. do we need any extra ones for things like publishing new versions? My understanding is that that's handled automagically by the bots that run against main branch commits.

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
```
tido64, line 28: We should drop this section since we don't know what this package needs.
```suggestion
```

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "devDependencies": {
+    "@rnx-kit/eslint-config": "^0.0.1",
+    "eslint": "^8.0.0",
+    "prettier": "^2.0.0"
```
tido64, line 32: This should be the only package that we need here. It provides the build, format, lint, test commands.
```suggestion
    "rnx-kit-scripts": "*",
```

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "eslintConfig": {
+    "extends": "@rnx-kit/eslint-config"
```
tido64, line 35: Likewise, we should use the ESLint preset in `rnx-kit-scripts`. It inherits from `@rnx-kit/eslint-config`.
```suggestion
    "extends": "rnx-kit-scripts"
```

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "directory": "packages/template"
+  },
+  "engines": {
+    "node": ">= 14"
+  },
+  "scripts": {
+    "build": "tsc",
+    "format": "prettier --write src/",
+    "lint": "eslint src/* --ext .js,.jsx,.ts,.tsx --config package.json --color"
+  },
```
tido64, line 24:
```suggestion
  "scripts": {
    "build": "rnx-kit-scripts build",
    "format": "rnx-kit-scripts format",
    "lint": "rnx-kit-scripts lint",
    "test": "rnx-kit-scripts test",
  },
```

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "workspace-tools": "^0.16.2"
+  },
+  "devDependencies": {
+    "@rnx-kit/eslint-config": "^0.0.1",
+    "eslint": "^8.0.0",
+    "prettier": "^2.0.0"
+  },
+  "eslintConfig": {
+    "extends": "@rnx-kit/eslint-config"
+  }
```
tido64, line 36: We should add Jest preset as well:
```suggestion
  },
  "jest": {
    "preset": "rnx-kit-scripts"
  }
```

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@rnx-kit/template",
```
tido64, line 2: We don't want to publish the template package:
```suggestion
{
  "private": true,
  "name": "@rnx-kit/template",
```

packages/template/package.json
tido64: I think these should at the very least point to lib, e.g.: https://github.com/microsoft/rnx-kit/blob/e6b3f03cca6170caf92ed170ae8518e94fb9a617/packages/dep-check/package.json#L10-L11

packages/template/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "version": "0.0.1",
+  "description": "Template folder to use as baseline for new packages in rnx-kit",
+  "homepage": "https://github.com/microsoft/rnx-kit/tree/main/packages/template#readme",
+  "license": "MIT",
+  "files": [
+    "src/*"
```
tido64, line 8: Point to lib instead? We typically don't ship `.ts` files.
```suggestion
    "lib/*"
```

packages/template/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "rnx-kit-scripts/tsconfig-shared.json",
+  "compilerOptions": {
+    "noEmit": true
+  },
```
tido64, line 5: We typically ship the `.js` files:
```suggestion
```

packages/template/README.md
tido64: `just.config.js` sets up commands (or tasks as they are called) such as build, format, lint, and test. We use it to have a uniform way of building etc. across all packages.

packages/template/README.md
tido64: I think this may be useful for users of this template, but maybe we can hide them in a comment? I don't know how useful they are for consumers of packages that are initialized from this template.

packages/template/package.json
tido64: If you set this package to private, you can also remove the change file. Beachball ignores private packages.

packages/template/src/index.ts
```diff
@@ -0,0 +1,5 @@
+const fakeMethod = (message: string) => {
```
tido64, line 1: If we're adding placeholder `src/index.ts`, maybe we should have a placeholder `test/index.test.ts` as well?

packages/template/src/types.ts
```diff
@@ -0,0 +1 @@
+export declare type FakeMethod = (message: string) => void;
```
tido64, line 1:
```suggestion
export type FakeMethod = (message: string) => void;
```

tido64:
Do you also intend to write script that "generates" a package for you? Basically copies the template into place, and does search and replace?

packages/template/package.json
tido64: Sorry, I made a mistake. I was thinking of Jest preset. This should be reverted 😅

change/@rnx-kit-template-9d95aaf4-b824-4537-bfff-0c8047ee2d89.json
afoxman: I can't find docs for this field on npmjs.org. What are the possible values for this? I've seen "type": "module" for ES6 modules, but that's it.

afoxman:
This looks good to me. Thank you! I don't have any extra comments on top of what Tommy raised.

scripts/new-package.js
```diff
@@ -0,0 +1,52 @@
...
+  const fs = require("fs");
+  const path = require("path");
+  const templatePath = path.join(__dirname, "../packages/template");
+  const newProjectPath = path.join(
+    __dirname,
+    "../packages/" + cleanProjectName
+  );
+
+  // copy the template folder to the new project folder
+  fse.copySync(templatePath, newProjectPath, { dereference: true });
```
afoxman, line 26: Before copying, you should check if it exists already and error-out.

afoxman (commented):
Looking good. I think this is just about ready.

scripts/new-package.js
```diff
@@ -0,0 +1,52 @@
...
+  const packageJson = require(packageJsonPath);
+  packageJson.name = "@rnx-kit/" + cleanProjectName;
+  packageJson.description = "New package called " + cleanProjectName;
+  packageJson.homepage =
+    "https://github.com/microsoft/rnx-kit/tree/main/packages/" +
+    cleanProjectName +
+    "#readme";
+  packageJson.repository.directory = "packages/" + cleanProjectName;
+
+  fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2));
```
tido64, line 39: We should also append an EOL to avoid the "missing newline at end of file" Git warning.
```suggestion
  fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2) + "\n");
```

scripts/new-package.js
```diff
@@ -0,0 +1,52 @@
...
+if (process.argv.length === 3) {
+  var projectName = process.argv[2];
+
+  // do some quick sanitization
+  var cleanProjectName = projectName.replace(/[|&;$%@"<>()+,]/g, "");
+
+  // copy the package/template folder to a new folder with the name of the value
+  // and then change the package.json file to the new name
+  const fse = require("fs-extra");
+  const fs = require("fs");
```
tido64, line 17: Hmm… According to their README, we should be able to do:
```suggestion
  const fs = require("fs-extra");
```
Does this work?

kelset: That's what I did originally, but fs-extra's `writeFileSync` was erroring out 😅 so I decided to only rely on it for the `copySync` method.

scripts/new-package.js
kelset: Good shout - here's what will happen:

Screenshot 2021-11-11 at 10 59 44

tido64 (approved):
🚀

PR #806, opened 2021-11-03
New package: @rnx-kit/typescript-react-native-compiler

afoxman:
### Description

Add a new package `@rnx-kit/typescript-react-native-compiler` to provide a standalone replacement for `tsc` which has react-native module resolution built in. 

The new compiler name is `rn-tsc`. It accepts the same command-line arguments as `tsc`, and adds a few additional arguments which enable and control react-native module resolution.

This PR includes all of the compiler front-end code, a full suite of tests, and documentation.

### Test plan

All CI tests pass, including the tests in this PR.


packages/typescript-react-native-compiler/src/usage/colors.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  blueBackground(s: string): string;
+}
+
+export const enum UsageColorMode {
+  None,
+  Limited,
+  Normal,
+  Rich,
+}
```
tido64, line 15: Does this have to be a `const enum`? Can we use string literal union instead? In general, we tend to avoid enums because they generated more code than plain string literals.

packages/typescript-react-native-compiler/src/usage/colors.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export function getUsageColorMode(): UsageColorMode {
+  if (!process.stdout.isTTY || process.env["NO_COLOR"]) {
+    return UsageColorMode.None;
+  }
+
+  const isWindows =
+    !!process.env["OS"] &&
+    process.env["OS"].toLowerCase().indexOf("windows") !== -1;
```
tido64, line 24: Can we use [`os.platform()`](https://nodejs.org/api/os.html#osplatform) instead?

packages/typescript-react-native-compiler/src/usage/colors.ts
```diff
@@ -0,0 +1,86 @@
...
+    process.env["COLORTERM"] === "truecolor" ||
+    process.env["TERM"] === "xterm-256color"
+  ) {
+    return UsageColorMode.Rich;
+  }
+
+  return UsageColorMode.Normal;
+}
+
+export function createUsageColors(
```
tido64, line 42: Why can't we use [chalk](https://github.com/chalk/chalk) (or some other alternative)?

packages/typescript-react-native-compiler/src/usage/text.ts
```diff
@@ -0,0 +1,38 @@
+import os from "os";
+
+export function wrapAndIndent(
```
tido64, line 3: Why is this needed? Doesn't the terminal already wrap long lines?

packages/typescript-react-native-resolver/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "typescript": "^4.0.0"
+  },
+  "peerDependencies": {
+    "typescript": "^4.0.0"
+  },
+  "depcheck": {
+    "ignoreMatches": [
+      "@rnx-kit/jest-preset"
+    ]
+  },
```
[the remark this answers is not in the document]
afoxman: When I removed it, everything worked. :)

packages/typescript-react-native-compiler/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "typescript": "^4.0.0"
+  },
+  "peerDependencies": {
+    "typescript": "^4.0.0"
+  },
+  "depcheck": {
+    "ignoreMatches": [
+      "@rnx-kit/jest-preset"
+    ]
+  },
```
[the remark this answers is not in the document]
afoxman: Removed

packages/typescript-react-native-compiler/src/cli.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (cmdLine.ts.options.all) {
+    showAllHelp();
+    return;
+  }
+
+  reportUnsupportedTscOptions(cmdLine.ts.options, [
+    "generateCpuProfile",
+    "build",
+    "locale",
+    "init",
```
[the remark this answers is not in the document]
afoxman: Yes. These are command-line options that are part of `tsc` and I wanted a graceful failure message when they were used. They're baked into the `tsc` code.

packages/typescript-react-native-compiler/src/command-line/extract.ts
```diff
@@ -0,0 +1,24 @@
+export function extractParameterValue(
```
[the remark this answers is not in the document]
afoxman: This does arg-parsing and removal from the given arg list. See `splice` on line 11.

You're right, it doesn't handle user errors. Transitioning to minimist would help catch those, but the code to remove args would be complicated since --x=y is allowed, as well as --x y.

I'd rather add some rudimentary checking here to see if `value` is itself another arg.

packages/typescript-react-native-compiler/src/commands/version.ts
```diff
@@ -0,0 +1,13 @@
+import { findPackage, readPackage } from "@rnx-kit/tools-node";
+import { tsc } from "./tsc";
+
+export function showVersion(): void {
+  const pkgFile = findPackage(module.filename);
```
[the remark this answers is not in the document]
afoxman: I didn't use it because it throws, and this code will only run once per process so I didn't need the caching.

packages/typescript-react-native-compiler/src/program.ts
```diff
@@ -0,0 +1,119 @@
...
+    //  No react-native platform was specified. Use the standard TypeScript
+    //  resolver.
+    //
+    //  Add a trace message handler which writes to a file or to the console.
+    //  Trace messages are only used when the compiler option `traceResolution`
+    //  is enabled.
+    //
+    if (traceResolutionLog) {
+      compilerHost.trace = (message: string): void => {
+        fs.writeFileSync(traceResolutionLog, message + os.EOL, {
```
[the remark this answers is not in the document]
afoxman: I wanted to do what you're suggesting, but the node docs say you can't issue a 2nd async call until you know the 1st one is complete. In other words, max-depth = 1 and it's my problem to sort out how to enforce it. Would have been nice, though, if they allowed it and guaranteed write ordering.

packages/typescript-react-native-compiler/src/usage/colors.ts
afoxman: Oh wow, I looked at the transpiled code and it emitted a normal enum! It's not supposed to do that unless we have `preserveConstEnums` set to true in TS compiler options (which we don't). What am I missing here?

packages/typescript-react-native-compiler/src/usage/colors.ts
afoxman: I probably could, but this is the way TS decides which color values to use, and I wanted my help color logic to match theirs.

packages/typescript-react-native-compiler/src/usage/colors.ts
afoxman: I started with chalk, and tried to replicate the right escape codes. I tried a lot of things because I wanted chalk to work. But I couldn't make it do what TS does in its colorful help. So I just replicated their logic.

packages/typescript-react-native-compiler/src/usage/text.ts
afoxman: Yes, terminal wrapping happens. My routine wraps whole words, though. The terminal just continues long words from one line to the next.

packages/typescript-react-native-compiler/src/usage/colors.ts
afoxman: I changed to string literals, but I'd still like to know what is going on...

packages/typescript-react-native-compiler/src/usage/text.ts
tido64: What happens if users resize their terminal window?

tido64: Also, what happens with the output of copy/pasting? Will it also include the newlines we add?

afoxman: That's separate from this. Resizing a terminal window will wrap individual characters. Like Typescript help which has indenting and character-wrapping (not word-wrapping), I'm trying to make the usage look nice in the current window. Resizing will ruin all of it. That's true of any formatted usage text.

tido64: Nvm, I missed that this is for usage text only.

tido64 (approved):
🚀

packages/typescript-react-native-compiler/src/usage/text.ts
afoxman: I suppose it depends on the terminal and the app you are pasting into. My experience has generally been that copying/pasting from a terminal keeps newlines and spaces. I'm sure there are exceptions.

PR #809, opened 2021-11-04
fix(eslint-plugin): prefer `module` over `main` and add debug options

tido64:
### Description

Adds `module` to the list of main fields to consider, and options for setting max depth and enabling debug output.

### Test plan

CI should pass.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-export-all.js
+++ b/packages/eslint-plugin/src/rules/no-export-all.js
@@ -180,8 +180,7 @@ function extractExports(context, moduleId, depth) {
                 case "ClassDeclaration":
                 // fallthrough
                 case "FunctionDeclaration": {
-                  const id = node.declaration.id;
-                  const name = id && id.name;
+                  const name = node.declaration.id?.name;
                   if (name) {
                     result.exports.push(name);
                   }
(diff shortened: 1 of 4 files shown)


afoxman (approved):

packages/eslint-plugin/src/rules/no-export-all.js
```diff
@@ -140,16 +152,16 @@ function parse({ filename, parserPath, parserOptions }, moduleId) {
...
-function extractExports(context, moduleId, depth = 0) {
-  if (depth >= MAX_DEPTH || typeof moduleId !== "string") {
+function extractExports(context, moduleId, depth) {
+  if (depth === 0 || typeof moduleId !== "string") {
     return null;
   }
 
   const parseResult = parse(context, moduleId);
-  if (!parseResult?.ast) {
+  if (!parseResult || !parseResult.ast) {
```
dzearing, line 164:
```suggestion
  if (!parseResult?.ast) {
```

dzearing (approved):

dzearing:
I still am not sure this works with `@fluentui/react` in the `src/Utilities.ts` file. Try that one. I was seeing errors from the parser complaining about how `../utilities/lib/index.js` was not part of the project (it's not.)

PR #810, opened 2021-11-04
fix(eslint-plugin): no-export-all fails to parse files outside project

tido64:
### Description

Fix no-export-all failing to parse files that are outside the TypeScript project. Also adds an option for ignoring `export *` from internal files.

### Test plan

CI should pass. @dzearing should validate :)

### DIFF
--- a/packages/eslint-plugin/src/rules/no-export-all.js
+++ b/packages/eslint-plugin/src/rules/no-export-all.js
@@ -10,6 +10,7 @@
  *   id: ESLintRuleContext["id"];
  *   options: {
  *     debug: boolean;
+ *     expand: "all" | "external-only";
  *     maxDepth: number;
  *   };
  *   settings: ESLintRuleContext["settings"];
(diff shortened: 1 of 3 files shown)


dzearing:
testing...

dzearing:
Testing results after linking the branch to fluent ui:

1. Shows errors on export * still ✔️
2. Now the Utilities.ts file error which has `export * from '@fluentui/utilities'` can be auto-fixed ✔️
3. Auto fixing reveals the javascript export names. ✔️ 
4. The internal-package exports are being flagged as errors. Changing the option to `expand: 'external-only'` makes them go away, but still complains about externals. ✔️
5. When auto expanding, it's not exporting the types of the library because it's only expanding the javascript and ignoring the d.ts files. ❌ 

Overall a huge improvement, and #5 can be fixed in a separate PR. I'll sign off on this iteration. Thanks and nice work!

dzearing (approved):

packages/eslint-plugin/src/rules/no-export-all.js
```diff
@@ -48,6 +49,19 @@ function isEmpty(namedExports) {
...
 
+/**
+ * Returns whether a module likely belongs to a project.
+ * @param {string?} project
+ * @param {string} modulePath
+ */
+function isLikelyInProject(project, modulePath) {
+  return (
+    project &&
+    (modulePath.endsWith(".ts") || modulePath.endsWith(".tsx")) &&
```
afoxman, line 60: Do you need a .d.ts check here as well?

afoxman (approved):

packages/eslint-plugin/src/rules/no-export-all.js
tido64: We don't read .d.ts files yet, as David pointed out above. I will work on enabling it in a separate PR.

PR #812, opened 2021-11-05
fix(config): `projectRoot` default value overrides the value from Metro config

### LINKED ISSUE #631: Configuration from metro.config.js get overridden by CLI argument defaults
### What happened?

I was moving some files around in an empty project and started getting a haste map error around the lines of ‘MyApp/src could not be found’. At some point, I had added a MyApp/src dir and moved index.js and App.tax in there, then moved them back out and deleted the folder. Thinking it was some weird cache issue, I spent a couple of days trying every trick in the book to clear my npm, pnpm, watchman, metro, and cradle caches.

Eventually I started stepping through the metro server code, trying to find out what cache file I had missed. Eventually though, I found out that the overriding value came from rnx kit.

I had my setting defined as follows:
In package.json, I did NOT define a property for the rnx kit server and bundle project root.
In my tsconfig, I did not specify any base urls, root dirs, etc.
In my metro config, I had my project root defined as __dirname (./MyApp).
The command I ran was react-native rnx-start, with no arguments, in ./MyApp.

What I expected to happen was that the override order was CLI args > package json rnxkit config > metro config. Since I left the first two empty, I expected the third to take effect.

However, it seems that the default values for the cli are taking precedence, over defined values in metro config. Seems to be in the below file?

https://github.com/microsoft/rnx-kit/blob/main/packages/config/src/getServerConfig.ts

I guess if this is expected behavior, it would be nice if the docs explicitly mentioned that certain values on metro.config will essentially be ignored, while others will work. Maybe a CLI warning too. Or maybe specify that it’s a default? Perhaps this is something inherited from metro, but if they have a default value for it I couldn’t find it.

I would also like to note that the react native community template for typescript doesn’t come with a src folder.

https://github.com/react-native-community/react-native-template-typescript/tree/master/template

I guess it’s a bit of an edge case since most everyone will make a src folder, but it is pretty confusing if someone tries to set up all their config and get it running before adding anything to the app template.

### Affected Package

@rnx-kit/cli

### Version

Latest?

- [X] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
➜  pnpx react-native info
info Fetching system and libraries information...
System:
    OS: Windows 10 10.0.19043
    CPU: (12) x64 Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
    Memory: 13.00 GB / 31.95 GB
  Binaries:
    Node: 16.8.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.11 - C:\Program Files\nodejs\yarn.CMD
    npm: 7.21.0 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK:
      API Levels: 30
      Build Tools: 30.0.2, 31.0.0
      System Images: android-31 | Google APIs Intel x86 Atom_64    
      Android NDK: Not Found
    Windows SDK: Not Found
  IDEs:
    Android Studio: Not Found
    Visual Studio: 16.9.31205.134 (Visual Studio Professional 2019)
  Languages:
    Java: javac 15
  npmPackages:
    @react-native-community/cli: ^6.0.0-0 => 6.0.0
    react: 17.0.2 => 17.0.2
    react-native: ^0.65.0-0 => 0.65.1
    react-native-windows: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

### Steps to Reproduce

Oops, forgot to push to the repro repo. Will do it tomorrow morning.

### Code of Conduct

tido64:
### Description

We don't actually need a default value for `projectRoot`, and it currently overrides the value in `metro.config.js`.

Resolves #631.

### Test plan

1. Build test app's dependencies:
   ```
   yarn
   cd packages/test-app
   yarn build --dependencies
   ```
2. Make sure bundle works: `yarn bundle`
3. Make sure dev server works: `yarn start`

### DIFF
--- a/packages/cli/src/start.ts
+++ b/packages/cli/src/start.ts
@@ -104,16 +104,20 @@ export async function rnxStart(
   // load Metro configuration, applying overrides from the command line
   const metroConfig = await loadMetroConfig(cliConfig, {
     ...cliOptions,
-    projectRoot: path.resolve(serverConfig.projectRoot),
+    ...(serverConfig.projectRoot
+      ? { projectRoot: path.resolve(serverConfig.projectRoot) }
+      : undefined),
     reporter,
-    ...(serverConfig.sourceExts ? { sourceExts: serverConfig.sourceExts } : {}),
+    ...(serverConfig.sourceExts
+      ? { sourceExts: serverConfig.sourceExts }
+      : undefined),
     ...(serverConfig.assetPlugins
       ? {
           assetPlugins: serverConfig.assetPlugins.map((p) =>
             require.resolve(p)
           ),
         }
-      : {}),
+      : undefined),
   });
 
   // prepare for typescript validation, if requested
--- a/packages/config/src/bundleConfig.ts
+++ b/packages/config/src/bundleConfig.ts
@@ -8,17 +8,6 @@ import type { OutputOptions } from "metro";
  */
 export type BundlerRuntimeParameters = {
   /**
-   * Path to the root of your react-native experience project. The bundle server uses
-   * this root path to resolve all web requests. Either absolute, or relative to the
-   * package.
-   *
-   * Note that `projectRoot` should also contain your Babel config, otherwise
-   * Metro won't be able to find it. For details, see
-   * https://github.com/microsoft/rnx-kit/issues/706.
-   */
-  projectRoot?: string;
-
-  /**
    * Choose whether to detect cycles in the dependency graph. If true, then a default set
    * of options will be used. Otherwise the object allows for fine-grained control over
    * the detection process.
--- a/packages/config/src/bundleConfig.ts
+++ b/packages/config/src/bundleConfig.ts
@@ -8,6 +8,17 @@ import type { OutputOptions } from "metro";
  */
 export type BundlerRuntimeParameters = {
   /**
+   * Path to the root of your react-native experience project. The bundle server uses
+   * this root path to resolve all web requests. Either absolute, or relative to the
+   * package.
+   *
+   * Note that `projectRoot` should also contain your Babel config, otherwise
+   * Metro won't be able to find it. For details, see
+   * https://github.com/microsoft/rnx-kit/issues/706.
+   */
+  projectRoot?: string;
+
+  /**
    * Choose whether to detect cycles in the dependency graph. If true, then a default set
    * of options will be used. Otherwise the object allows for fine-grained control over
    * the detection process.
--- a/packages/config/src/getServerConfig.ts
+++ b/packages/config/src/getServerConfig.ts
@@ -14,7 +14,6 @@ export function getServerConfig(
   config: KitConfig
 ): ServerWithRequiredParameters {
   const defaultConfig: ServerWithRequiredParameters = {
-    projectRoot: process.cwd(),
     detectCyclicDependencies: true,
     detectDuplicateDependencies: true,
     typescriptValidation: true,
--- a/packages/config/src/getServerConfig.ts
+++ b/packages/config/src/getServerConfig.ts
@@ -14,7 +14,7 @@ export function getServerConfig(
   config: KitConfig
 ): ServerWithRequiredParameters {
   const defaultConfig: ServerWithRequiredParameters = {
-    projectRoot: "src",
+    projectRoot: process.cwd(),
     detectCyclicDependencies: true,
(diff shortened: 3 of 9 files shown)


packages/config/src/getServerConfig.ts
```diff
@@ -14,7 +14,7 @@ export function getServerConfig(
   config: KitConfig
 ): ServerWithRequiredParameters {
   const defaultConfig: ServerWithRequiredParameters = {
-    projectRoot: "src",
+    projectRoot: process.cwd(),
```
tido64, line 17: @afoxman: Should this value come from `metro.config.js` if it's present? Or is the intention that we don't consider it ever?

tido64: Had a closer look. I think this shouldn't be a required value at all. Let me know if you disagree.

kelset (approved):

packages/config/src/bundleConfig.ts
```diff
@@ -7,6 +7,17 @@ import type { OutputOptions } from "metro";
  * Parameters controlling bundler behavior at runtime.
  */
 export type BundlerRuntimeParameters = {
+  /**
```
afoxman, line 10: The react-native CLI only exposes `projectRoot` for the `start` command. I don't think it belongs here, otherwise it will be part of the `bundle` command as well.

packages/config/src/serverConfig.ts
```diff
@@ -1,17 +1,13 @@
...
-   * package.
-   *
-   * @default "src"
-   */
-  projectRoot: string;
-};
+export type ServerRequiredParameters = Omit<
+  BundlerRuntimeParameters,
+  "projectRoot"
+>;
```
afoxman, line 6: Reverting this, and making projectRoot optional should give you the right result. If neither kit config nor command line override sets projectRoot, then the Metro config value can come through.

afoxman, line 6: I think moving `projectRoot` down to `ServerParameters` and making it optional will give you right scoping for this property.

packages/cli/src/start.ts
```diff
@@ -104,16 +104,20 @@ export async function rnxStart(
   // load Metro configuration, applying overrides from the command line
   const metroConfig = await loadMetroConfig(cliConfig, {
     ...cliOptions,
-    projectRoot: path.resolve(serverConfig.projectRoot),
+    ...(serverConfig.projectRoot
+      ? { projectRoot: path.resolve(serverConfig.projectRoot) }
+      : undefined),
```
afoxman, line 109: Is there really a difference between ...({}) and ...(undefined)? I think they both yield an empty object {}.

afoxman (commented):
Letting Metro config's `projectRoot` come through is the right fix, but it should be limited to the `start` command.

packages/config/src/serverConfig.ts
tido64: Isn't `projectRoot` also used by the bundle command? I think it does, since you can set `projectRoot: "src"` and specify `--entry-file src/index.js` and Metro will tell you that it cannot find `src/src/index.js`

packages/cli/src/start.ts
tido64: In my mind, `{}` creates a new object, whereas `undefined` doesn't. But I have no basis for believing this and could be completely wrong. V8 could optimize this away.

packages/config/src/serverConfig.ts
tido64: Oh wait, I've just read your other comment now (I started from the bottom). Are you saying that these are only CLI args? If so, I will move it.

afoxman: Metro config has `projectRoot` regardless of the command being run. However, `react-native bundle ...` doesn't expose `projectRoot` as an option, so our bundling CLI and config doesn't either. `react-native start ...` does expose `projectRoot` which is why I had confined the CLI param and config to our "server" config/code. (To answer your question directly -- `projectRoot` is in both our config and in our CLI, but only for "server" mode)

packages/config/src/getServerConfig.ts
afoxman: This is going to need downstream changes for anyone using our CLI. I'm not sure if that list is empty, or not.

afoxman (approved):

PR #816, opened 2021-11-08
fix(eslint-plugin): prefer parsing `.d.ts` over `.js`

tido64:
### Description

Prefer parsing `.d.ts` over `.js` so we don't lose type information.

### Test plan

CI should pass. @dzearing to verify his use cases.

### DIFF
--- a/change/@rnx-kit-eslint-plugin-ae9acd7d-6ab1-46dc-984d-7d010616e4bc.json
+++ b/change/@rnx-kit-eslint-plugin-ae9acd7d-6ab1-46dc-984d-7d010616e4bc.json
@@ -3,5 +3,5 @@
   "comment": "Prefer parsing `.d.ts` over `.js` so we don't lose type information",
   "packageName": "@rnx-kit/eslint-plugin",
   "email": "4123478+tido64@users.noreply.github.com",
-  "dependentChangeType": "patch"
+  "dependentChangeType": "none"
 }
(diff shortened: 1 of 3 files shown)


dzearing (approved):
Tested, works great!

dzearing (changes requested):
Found an issue; see comment

dzearing:
I noticed that when I expand FocusZone.ts, I get this:

```ts
export { FocusZone, FocusZoneTabbableElements } from '@fluentui/react-focus';
export type { FocusZoneTabbableElements, IFocusZone, IFocusZoneProps } from '@fluentui/react-focus';
````

Which is weird:

image

It's exporting a `type` as a non-type?

**Comment by dzearing** [comment]
Other than this, this is great. I'll sign off, but if you can look at this issue before merging and resolve with a simple "prefer non-type export for redundant exports" check, that would help. Or, think about it, maybe you have a better approach.

**Review by dzearing — APPROVED** [review]

**Review by dzearing — APPROVED** [review]

**Comment by tido64** [comment]
> Other than this, this is great. I'll sign off, but if you can look at this issue before merging and resolve with a simple "prefer non-type export for redundant exports" check, that would help. Or, think about it, maybe you have a better approach.

Enums in general are really weird in JavaScript. I think we could've avoided the awkward double definition with string literal unions:

```typescript
export type FocusZoneTabbableElements = "none" | "all" | "inputOnly";
```

Anyway, I've added it as a test case.

PR #831, opened 2021-11-09
Transition to the new Metro TypeScript resolver

afoxman:
### Description

A recent PR introduced a new TypeScript resolver with react-native support -- @rnx-kit/typescript-react-native-resolver. Transition our CLI to use this, instead of the tailor-made resolver is currently uses. Also, remove the older typechecking plugin for Metro, now that there is a package and a CLI which can be used instead.

### Test plan

CI tests pass.


packages/cli/src/metro-config.ts
```diff
@@ -12,14 +12,16 @@ import {
...
+import { changeHostToUseReactNativeResolver } from "@rnx-kit/typescript-react-native-resolver";
 import {
   Project,
   readConfigFile,
   createDiagnosticWriter,
 } from "@rnx-kit/typescript-service";
 import type { DeltaResult, Graph } from "metro";
 import type { InputConfigT, SerializerConfigT } from "metro-config";
-import { MetroTypeScriptResolverHost } from "./metro-ts-resolver";
+import ts from "typescript";
```
tido64, line 23: Can we import the type only?
```suggestion
import type { LanguageServiceHost } from "typescript";
```

tido64 (approved):

PR #857, opened 2021-11-16
feat: prototype auth module for RNSDK

tido64:
### Description

Prototype RNX auth module implementation.

### Test plan

n/a


packages/react-native-test-app-msal/ios/RNXAuthMSALModule.m
```diff
@@ -0,0 +1,129 @@
...
+        dispatch_async(dispatch_get_main_queue(), ^{
+          [self acquireTokenWithScopes:scopes
+                     userPrincipalName:userPrincipalName
+                           accountType:accountType
+                       onTokenAcquired:onTokenAcquired];
+        });
+        return;
+    }
+
+    [TokenBroker.sharedBroker
```
alloy, line 43: @tido64 Where's TokenBroker defined, is it a MS lib? Because I'm really surprised by the lack of 3 char namespace prefix.

tido64: It's a Swift module, so it doesn't need the prefix (Apple has been discouraging it for some time now): https://github.com/microsoft/rnx-kit/blob/main/packages/react-native-test-app-msal/ios/TokenBroker.swift

packages/react-native-test-app-msal/android/build.gradle
```diff
@@ -145,6 +150,12 @@ android {
         preBuild.dependsOn(copyMsalConfig)
 
         main.res.srcDirs += generatedResDir
+
+        if (rnxAuth == null) {
+            main.java.srcDirs += "src/nornx/java"
```
afoxman, line 155: Now that rnxauth is available, should the nornx path go away?

packages/react-native-test-app-msal/ios/RNXAuthMSALModule.m
```diff
@@ -0,0 +1,129 @@
...
+                     onTokenAcquired(authResult, nil);
+                 }
+               }];
+}
+
+@end
+
+RNXAuthErrorType RNXAuthErrorTypeFromAuthErrorType(AuthErrorType type)
+{
+    switch (type) {
```
afoxman, line 69: Both Android and iOS seem to have this layer of mapping to RNX enums. Are they both platform-specific mappings and types that happen to have the name `AuthErrorType`, or is this something more generic that can be buried away inside of the interface definition? If everyone is speaking `AuthErrorType`, do you even need `RNXAuthErrorTypeXyz`?

packages/test-app/metro.config.js
```diff
@@ -1,13 +1,29 @@
...
+  // If USE_AUTH_MOCK=1, exclude the real module to enable the mock.
+  ...(process.env["USE_AUTH_MOCK"]
+    ? [/.*packages[/\\]react-native-auth.*/]
+    : []),
+]);
 
 module.exports = makeMetroConfig({
   resolver: {
+    extraNodeModules: {
+      "@rnx-kit/react-native-auth": require("path").join(
```
afoxman, line 17: Should this also be conditioned on `USE_AUTH_MOCK`?

packages/react-native-auth/android/src/main/java/com/microsoft/reactnativesdk/auth/ReactNativeAuthModule.kt
```diff
@@ -0,0 +1,91 @@
...
+    fun acquireToken(
+        scopes: ReadableArray,
+        userPrincipalName: String,
+        accountType: String,
+        promise: Promise
+    ) {
+        acquireToken(
+            scopes.toStringArray(),
+            userPrincipalName,
+            AccountType.from(accountType)
```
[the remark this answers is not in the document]
tido64: This is Kotlin's trailing closure syntax. If the last parameter is a function, these two are equivalent:

```kotlin
acquireToken(scopes, upn, accountType) { result, error ->
    ..
}
```

```kotlin
acquireToken(scopes, upn, accountType, { result, error ->
    ..
})
```

Ruby and Swift are two other languages I know of that also support this.

packages/react-native-auth/ios/RNXAuthModule.m
```diff
@@ -0,0 +1,81 @@
...
+                      resolver:(RCTPromiseResolveBlock)resolve
+                      rejecter:(RCTPromiseRejectBlock)reject)
+// clang-format on
+{
+    [self acquireTokenWithScopes:scopes
+               userPrincipalName:userPrincipalName
+                     accountType:RNXAccountTypeFromString(accountType)
+                 onTokenAcquired:^(RNXAuthResult *result, RNXAuthError *error) {
+                   if (result == nil) {
+                       reject(RNXStringFromAuthErrorType(error.type),
```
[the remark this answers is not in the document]
tido64: In Objective-C calling a method or accessing a property on a `nil` is well-defined. The former is a noop, while the latter returns a zero value. In this case, `error.type` returns `0`, which `RNXStringFromAuthErrorType()` will return `RNXAuthErrorTypeUnknown` for.

packages/react-native-test-app-msal/ios/RNXAuthMSALModule.m
tido64: I probably should've divided this PR in two (at least). `react-native-test-app-msal` is a module that provides auth capability to `react-native-test-app`. Its use case is for features that require auth, but need their own solution when not embedded directly in the hosting apps, e.g. your own repo vs. Outlook's.

What's new in this PR is that `react-native-test-app-msal` also implements the `RNXAuth` interface so that it can be accessed from the JS side. The relationship is essentially:

```mermaid
classDiagram
      RNXAuth <|-- ReactNativeTestAppMSAL
      ReactNativeTestAppMSAL <|-- MSAL
```

Where it is currently possible to access `react-native-test-app-msal` directly in native:

```mermaid
flowchart LR
      Native --> ReactNativeTestAppMSAL --> MSAL
```

And now in JS via the new `RNXAuth` module:

```mermaid
flowchart LR
      JS --> RNXAuth --> ReactNativeTestAppMSAL --> MSAL
```

`RNXAuth` should be agnostic to the underlying implementation. It just so happens that the names currently match with MSAL's.

packages/react-native-test-app-msal/android/build.gradle
tido64: We still need to keep this because it should be possible to consume `react-native-test-app-msal` without taking a dependency on `RNXAuth`. I go into more details below…

packages/react-native-auth/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  | "InteractionRequired"
+  | "NoResponse"
+  | "PreconditionViolated"
+  | "ServerDeclinedScopes"
+  | "ServerProtectionPoliciesRequired"
+  | "Timeout"
+  | "UserCanceled"
+  | "WorkplaceJoinRequired";
+
+export type AuthError = {
```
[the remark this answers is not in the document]
tido64: This type is used by the `userInfo` field of the error object. I've renamed it to make it clearer.

packages/react-native-auth/ios/RNXAccountType.m
```diff
@@ -0,0 +1,9 @@
+#import "RNXAccountType.h"
+
+RNXAccountType RNXAccountTypeFromString(NSString *type)
+{
+    if ([type isEqualToString:@"Organizational"]) {
+        return RNXAccountTypeOrganizational;
+    }
+    return RNXAccountTypeMicrosoftAccount;
```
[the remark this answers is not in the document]
tido64: Good point. Added the `invalid` type and an assertion. We can't use an exception here since they're not compatible with Swift.

packages/react-native-auth/android/src/main/java/com/microsoft/reactnativesdk/auth/ReactNativeAuthModule.kt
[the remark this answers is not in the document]
afoxman: Good to know.

packages/react-native-auth/README.md
```diff
@@ -6,10 +6,44 @@
...
+  scopes,
+  userPrincipalName,
+  "MicrosoftAccount"
+);
+```
+
+## Motivation
+
+Many features built at Microsoft require authentication. Most teams have their
+own solution for providing access tokens to their features during the
```
afoxman, line 38: maybe mention a few to balance out what MS does vs the rest of the world, like OAuth2 or WebAuthN.

packages/react-native-auth/android/src/main/java/com/microsoft/reactnativesdk/auth/AuthErrorType.kt
```diff
@@ -1,3 +1,7 @@
+// TODO: This file will be generated from the TypeScript type definitions. For
```
afoxman, line 1: If the code generator isn't ready yet, is there an issue tracking it?

packages/react-native-auth/src/index.ts
```diff
@@ -6,8 +6,15 @@ if (!NativeModules.RNXAuth) {
   );
 }
 
+/**
+ * Microsoft account type. Valid types are Microsoft accounts (or MSA), and
+ * organizational (M365).
```
afoxman, line 11: Are there any industry standard account types that aren't specific to Microsoft which could be added here? Without them, people may look at this package and decide not to use it because they don't see what they need.

packages/react-native-auth/src/index.ts
```diff
@@ -21,29 +28,55 @@ export type AuthErrorType =
...
+  message?: string;
+  userInfo: AuthErrorUserInfo;
+};
+
+/**
+ * The authentication error object contains a stack trace on Android.
+ * @property nativeStackAndroid Android stack trace
+ */
+export type AuthErrorAndroid = AuthErrorBase & {
   nativeStackAndroid?: string[];
```
afoxman, line 60: Do the stack trace props need to be different between android and ios? Seems like they should be the same since they'll probably be handled the same -- logged or put up in a redbox.

packages/react-native-auth/src/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Authentication error object thrown by {@link acquireToken}.
+ * @property code The type of error that occurred during authentication
+ * @property message The error message
+ * @property userInfo Error details provided by the underlying implementation
+ */
+export type AuthErrorBase = {
+  code: AuthErrorType;
+  message?: string;
```
afoxman, line 51: What is the difference between `message` and `userInfo.message`?

afoxman: Also, `code` and `userInfo.type`?

packages/react-native-test-app-msal/ios/RNXAuthMSALModule.m
afoxman: The abstraction makes more sense now. The only way for this logic duplication to go away is if both ios and android could share code (e.g. C++ instead of ObjC and Kotlin).

packages/react-native-auth/README.md
tido64: Changed the wording a bit to clarify why this native module is needed. Let me know if you have better suggestions.

packages/react-native-auth/android/src/main/java/com/microsoft/reactnativesdk/auth/AuthErrorType.kt
tido64: We will be using the TurboModules code generator whenever it is ready: https://github.com/microsoft/rnx-kit/issues/1106

packages/react-native-auth/src/index.ts
tido64: Since this is still a prototype, I think we want to start with what's most needed internally right now. I'll change the wording to not exclude any account types.

packages/react-native-auth/src/index.ts
tido64: The shape of `AuthErrorBase` and its platform specific variants are defined by React Native core. I couldn't find them in the types so I had to redefine them here. The stack trace is populated if you pass a native error object (`Exception` on Android and `NSError` on iOS/macOS).

packages/react-native-auth/src/index.ts
tido64: The shape of `AuthErrorBase` and its platform specific variants are defined by React Native core. I couldn't find them in the types so I had to redefine them here. `AuthErrorBase`'s fields are populated by passing a native error object (`Exception` on Android and `NSError` on iOS/macOS). Only `userInfo` can be specified by the module implementation. Both `type` and `message` may differ, e.g. the inner exception or a more user friendly error message.

tido64: Also note to self: We should probably upstream the Error type.

packages/react-native-auth/README.md
```diff
@@ -34,12 +34,17 @@ const result = await acquireToken(
 
 ## Motivation
 
-Many features built at Microsoft require authentication. Most teams have their
-own solution for providing access tokens to their features during the
-development loop, or they rely on their hosting app to provide such a solution.
-Solutions are often custom-made for their apps and cannot be shared with others
-without significant effort. They will also have to duplicate this effort when
-integrating into other apps.
+Many features built at Microsoft require authentication. The tricky thing about
```
afoxman, line 37: My hangup is that the starting-point for the docs is "at Microsoft, ...". Maybe remove "built at Microsoft" so devs reading this see it in terms of their apps which likely don't use any Microsoft tech.

packages/react-native-auth/src/index.ts
afoxman: Maybe just remove "Microsoft" from the start of the comment, and add "Other account types can be added as needed." at the end of the comment.

afoxman (approved):
Code looks good to me. Left a few comments about documentation.

packages/react-native-auth/README.md
tido64: This is a good point. I removed the "at Microsoft" part.

PR #860, opened 2021-11-17
chore: disable install scripts

tido64:
### Description

As a response to recent malware incidents (see https://github.blog/2021-11-15-githubs-commitment-to-npm-ecosystem-security/), we are disabling scripts as a local mitigation.

### Test plan

CI should pass.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+ignore-scripts true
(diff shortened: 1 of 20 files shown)


afoxman (approved):

afoxman:
I think this change breaks the Go toolchain, but I'm ok with that since we're not using it right now. Both the install and the build steps were running as part of build-tools and were guaranteed to be finished before any other packages were build/tested/etc.

tido64:
> I think this change breaks the Go toolchain, but I'm ok with that since we're not using it right now. Both the install and the build steps were running as part of build-tools and were guaranteed to be finished before any other packages were build/tested/etc.

I can take a look at this next.

PR #869, opened 2021-11-18
chore: replace just-scripts

tido64:
### Description

We are replacing just-scripts to:

1. Make it easier to add/modify commands as we work with our own tools
2. Remove the confusion from having two Just configs and not consuming the defined tasks within the scripts package
3. Reduce the overhead of managing transitive (dated) dependencies

### Test plan

CI should be green.


scripts/src/commands/bundle.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param {Record<string, unknown>=} args
+ */
+function bundle({ minify, platform } = { minify: false, platform: "node" }) {
+  const fs = require("fs");
+
+  const manifest = fs.readFileSync("package.json", { encoding: "utf-8" });
+  const { main, dependencies } = JSON.parse(manifest);
+
+  const targetPlatform = (() => {
```
afoxman, line 15: Why wrap this in a function? (There's nothing wrong with it, I just want to know if there's something subtle going on here).

afoxman (approved):

scripts/src/commands/bundle.js
tido64: I was just too lazy to extract and make a proper function of the thing. I avoid mutable variables, and this is a quick way to inline a function and call it immediately. I'll pull it out.

PR #872, opened 2021-11-19
Minor fixes to TypeScript utility packages

afoxman:
### Description

Make minor fixes to TypeScript utility packages.

`@rnx-kit/typescript-react-native-resolver`
When modifying a TypeScript host, only change trace() when logging is enabled. Also, when changing file/directory APIs, bind the original calls to the host so that the 'this' reference is set properly.

`@rnx-kit/typescript-service`
Don't format/print "suggestion" diagnostics. Doing so hits an internal debug assert in TypeScript.

### Test plan

CI tests pass. Manual validation in a private repo.

### DIFF
--- a/packages/typescript-react-native-resolver/src/log.ts
+++ b/packages/typescript-react-native-resolver/src/log.ts
@@ -89,11 +89,11 @@ export class ResolverLog {
 export function changeModuleResolutionHostToLogFileSystemReads(
   host: ModuleResolutionHostLike
 ): void {
-  const originalFileExists = host.fileExists;
-  const originalReadFile = host.readFile;
-  const originalDirectoryExists = host.directoryExists;
-  const originalRealpath = host.realpath;
-  const originalGetDirectories = host.getDirectories;
+  const originalFileExists = host.fileExists.bind(host);
+  const originalReadFile = host.readFile.bind(host);
+  const originalDirectoryExists = host.directoryExists.bind(host);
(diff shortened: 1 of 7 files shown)


packages/typescript-react-native-resolver/src/log.ts
```diff
@@ -89,11 +89,11 @@ export class ResolverLog {
...
-  const originalFileExists = host.fileExists;
-  const originalReadFile = host.readFile;
-  const originalDirectoryExists = host.directoryExists;
-  const originalRealpath = host.realpath;
-  const originalGetDirectories = host.getDirectories;
+  const originalFileExists = host.fileExists.bind(host);
+  const originalReadFile = host.readFile.bind(host);
+  const originalDirectoryExists = host.directoryExists.bind(host);
+  const originalRealpath = host.realpath ? host.realpath.bind(host) : undefined;
+  const originalGetDirectories = host.getDirectories.bind(host);
```
tido64, line 96: Nit: I find closures less confusing to read than `.bind()`:
```suggestion
  const originalFileExists = () => host.fileExists();
  const originalReadFile = () => host.readFile();
  const originalDirectoryExists = () => host.directoryExists();
  const originalRealpath = host.realpath ? () => host.realpath() : undefined;
  const originalGetDirectories = () => host.getDirectories();
```

tido64 (approved):

PR #880, opened 2021-11-19
chore: migrate from Beachball to Changesets

### LINKED ISSUE #168: Publish: Race condition is preventing packages from being published
image

When two PRs land close to each other, one of them will fail the publish step. This is because we currently run publish on every commit, removing the change files, and committing them back to `main`. The issue here is that the second commit still sees all of the change files that are now removed on latest `main`, tries to publish them, and fails.

### LINKED ISSUE #875: ci: Beachball bumps the whole repo despite change type `none`
### What happened?

Beachball bumped and published new packages even though all the changes were of type `none`: https://github.com/microsoft/rnx-kit/runs/4256768654?check_suite_focus=true

`@rnx-kit/console` hardly ever changes, and yet got bumped: https://github.com/microsoft/rnx-kit/commit/224bb2e7c19650439c14a1c0a448f87808794cdd#diff-2f99c5a440bee429e4b067bbc14d702c8d3cc76885d4ec01844dfbed9a292a3a

And because it got bumped, all of its dependents also get bumped.

### Affected Package

beachball

### Version

2.20.0

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
n/a
```

### Steps to Reproduce

n/a

### Code of Conduct

tido64:
### Description

- `change` remains mostly the same. The difference now is that we select the packages that should have the same change log entry. Where we used to specify `--type none`, we can do `yarn change --empty` instead.
- `publish:beachball` was split into `version:changesets` (for bumping version numbers) and `publish:changesets` (actually publishing).
- Packages are no longer published on push to `main`. Instead, a PR is created that collects all the changes. When we are ready to merge, it will publish all the packages and clean up the change files. For an example, see https://github.com/atlassian/changesets/pull/663.

Resolves #168
Resolves #875

### Test plan

n/a


tido64:
image

.changeset/thick-turkeys-hang.md
```diff
@@ -0,0 +1,2 @@
+---
```
afoxman, line 1: Is this file required?

afoxman (approved):

.changeset/thick-turkeys-hang.md
tido64: Yes, this is an empty change file to satisfy the every-commit-needs-a-changelog requirement. You can generate one by invoking `yarn change --empty`.

tido64: If we install their GitHub bot, we can drop the requirement. The bot will instead comment on the PR when it is missing. It also links you to a page where you can easily add one.

PR #874, opened 2021-11-19
fix(eslint/no-export-all): add support for namespaces

tido64:
### Description

Auto-fix namespaces.

Resolves #871.

### Test plan

See repro in #871. Tests were added.

```diff
--- src/Styling.prev.ts	2021-11-19 10:00:15.000000000 +0100
+++ src/Styling.ts	2021-11-19 10:04:53.000000000 +0100
@@ -31,6 +31,7 @@
   ScreenWidthMinXXXLarge,
   Stylesheet,
   ThemeSettingName,
+  ZIndexes,
   buildClassMap,
   concatStyleSets,
   concatStyleSetsWithProps,
```

PR #886, opened 2021-11-23
Updating minimum node requirement in eslint-plugin to 14.

dzearing:
Adding minimum node requirement (14) to `@rnx-kit/eslint-plugin` package.json definition (since optional chaining is not supported in 12.)

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -14,6 +14,9 @@
   "files": [
     "src/*"
   ],
+  "engines": {
+    "node": ">=14"
+  },
   "main": "src/index.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/package.json
```diff
@@ -14,6 +14,9 @@
   "files": [
     "src/*"
   ],
+  "engines": {
+    "node": ">=14"
+  },
```
tido64, line 19: I've added this field in #874. If you can approve it, we can close this PR.

dzearing: Dang you're on top of stuff 💯

dzearing: done and done

PR #897, opened 2021-11-25
fix(eslint/no-export-all): `const` enums should be exported as types

### LINKED ISSUE #887: no-export-all: const enums should be considered "types"
### What happened?

A library exported a const enum; the fixer of no-export-all exported it as a non-type, but this causes errors when being consumed because there is no named export in the produced javascript when compiling a const enum.

const enums are the devil. We shouldn't export them from a library. But we have a couple deprecated places which still export. I manually fixed, but I would propose the fixer exports them as types.

### Affected Package

@rnx-kit/eslint-plugin

### Version

0.2.4

- [ ] Android
- [ ] iOS
- [ ] macOS
- [X] Windows

### System Information

```true
n/a
```

### Steps to Reproduce

Fix Icons.ts in @fluentui/react package.

expected:
`IconNames` is exported as a type (const enum)

resulted:
`IconNames` is a named export but doesn't exist in `@fluentui/font-icons-mdl2`

### Code of Conduct

tido64:
### Description

[`const` enums](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) are a weird beast. The following code:

```typescript
export const enum Foo {
    Bar,
    Baz,
}
```

Gets transpiled to:

```ts
// .js
export {};

// .d.ts
export declare const enum Foo {
    Bar = 0,
    Baz = 1
}
```

Since no code is exported, the fixer has to export it as a type instead.

Resolves #887.

### Test plan

See repro in #887. The new test should also pass.

### DIFF
--- a/change/@rnx-kit-eslint-plugin-2c6ac533-ac8c-4d07-9f0c-5acd82e04525.json
+++ b/change/@rnx-kit-eslint-plugin-2c6ac533-ac8c-4d07-9f0c-5acd82e04525.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add proper support for `const` enums",
+  "packageName": "@rnx-kit/eslint-plugin",
+  "email": "4123478+tido64@users.noreply.github.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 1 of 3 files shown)


afoxman (approved):

dzearing (commented):
Awesome!

PR #905, opened 2021-11-30
fix(metro-resolver-symlinks): add lib -> src remapper util

tido64:
### Description

- Refactored resolvers to conform to a single interface
- Added a utility import path remapper
- Updated README.

### Test plan

CI should pass.

### DIFF
--- a/packages/metro-resolver-symlinks/README.md
+++ b/packages/metro-resolver-symlinks/README.md
@@ -45,22 +45,19 @@ module.exports = makeMetroConfig({
 ### `remapModule`
 
 `remapModule` allows additional remapping of modules. For instance, there is a
-`lib/utils/remapImportPath.js` that remaps requests of `lib/**/*.js` to
+`remapImportPath` utility that remaps requests of `lib/**/*.js` to
 `src/**/*.ts`. This is useful for packages that don't correctly export
 everything in their main JS file.
 
 ```diff
  const { makeMetroConfig } = require("@rnx-kit/metro-config");
  const MetroSymlinksResolver = require("@rnx-kit/metro-resolver-symlinks");
-+const {
-+  remapImportPath,
-+} = require("@rnx-kit/metro-resolver-symlinks/lib/utils/remapImportPath");
 
  module.exports = makeMetroConfig({
    projectRoot: __dirname,
    resolver: {
      resolveRequest: MetroSymlinksResolver({
-+      remapModule: remapImportPath({
++      remapModule: MetroSymlinksResolver.remapImportPath({
 +        test: (moduleId) => moduleId.startsWith("@rnx-kit/"),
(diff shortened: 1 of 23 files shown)


packages/metro-resolver-symlinks/src/utils/remapImportPath.ts
```diff
@@ -0,0 +1,156 @@
...
+    return undefined;
+  }
+
+  const fromDir = originModulePath
+    ? path.dirname(originModulePath)
+    : process.cwd();
+  const packageName = scope ? `${scope}/${name}` : name;
+  return resolver(
+    fromDir,
+    `${packageName}/${modulePath.replace(pattern, "src")}`
```
afoxman, line 66: It looks like the pattern regex could match on a lengthy path, replacing it with just "src" (the '*' matcher is what I'm looking at). Is that desired? I'd expect it to specifically look for ./lib or lib and replace it with ./src or src.

packages/metro-resolver-symlinks/README.md
```diff
@@ -23,3 +35,43 @@ module.exports = makeMetroConfig({
...
+     }),
+   },
+ });
+```
+
+> **Sidenote:** When Metro releases a version with the ability to set a
+> [custom resolver for Haste requests](https://github.com/facebook/metro/commit/96fb6e904e1660b37f4d1f5353ca1e5477c4afbf),
+> this way of remapping modules is preferable over
+> `@rnx-kit/babel-plugin-import-path-remapper`. The Babel plugin mutates the AST
+> and requires a second pass.
```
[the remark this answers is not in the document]
tido64: It will, but currently doesn't. The commit I referenced there has not been released yet. If you set `remapModule` today, it may only be called if you remap a file that does not live inside `node_modules` folder. Metro resolves using Haste first.

packages/tools-react-native/src/platform.ts
```diff
@@ -3,6 +3,51 @@
  */
 export type AllPlatforms = "ios" | "android" | "windows" | "win32" | "macos";
 
+// TODO: `react-native config` is too slow. Hard-coding this list until we can
+// figure out a better solution.
```
afoxman, line 7: I'm pretty sure this will show up in Intellisense as the doc comment for AVAILABLE_PLATFORMS. Can you replace it with documentation on what this is? If there's a TODO, maybe file an issue and reference it here.

afoxman (approved):

packages/tools-react-native/src/platform.ts
tido64: I think VS Code will only generate documentation if you use `/** */`:

image

I've created an issue here: https://github.com/microsoft/rnx-kit/issues/925

PR #912, opened 2021-11-30
Version Packages

afoxman:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases


afoxman:
@tido64 I'm ok with this running as me, but we should eventually move it over to rnsdkbot. Also, do you know why the CI pipeline isn't running? Is there something weird with the triggering system?

tido64:
> @tido64 I'm ok with this running as me, but we should eventually move it over to rnsdkbot. Also, do you know why the CI pipeline isn't running? Is there something weird with the triggering system?

I'm thinking of using the automatic token for this: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

But since a PR was already created, I wasn't sure whether it will break if I change it now.

The reason this isn't running is because the commit message contains `[skip ci]`. It only deletes change files and updates change logs, so I think this is fine. It just means we will have to use this checkbox whenever we want to merge:

image

tido64:
It looks like Changesets correctly detected that `@rnx-kit/metro-resolver-symlinks` is the only consumer of the new functions added in `@rnx-kit/tools-react-native` and bumped it. Kept all other packages on the same version.

tido64:
After a small hiccup, packages are now published: https://github.com/microsoft/rnx-kit/runs/4399652517?check_suite_focus=true

Currently looking into how to prevent this hiccup.

PR #914, opened 2021-12-01
ci: publish with the GitHub provided token

tido64:
### Description

Use the token provided by GitHub to create the release PR.

More details on the token here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

### Test plan

n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,10 +7,6 @@ jobs:
     name: "Build and publish"
     runs-on: ubuntu-latest
     steps:
-      - name: Set up Git
-        run: |
-          git config --global user.email "53619745+rnbot@users.noreply.github.com"
-          git config --global user.name "React Native Bot"
       - name: Set up Node 14
         uses: actions/setup-node@v2.5.0
         with:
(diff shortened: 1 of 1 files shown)


afoxman:
It's been awhile, but I think I tried using this in the past with beachball. I ran into problems when beachball tried to merge changes. The branch-protections in GitHub blocked this token from being able to merge.

It seems like that shouldn't be a problem now, since the merge will come from a 2nd PR rather than automation... but I wanted to bring it up and see if you thought of anything that may break.

afoxman:
If you go ahead with this, can you delete the GIT_TOKEN secret once the transition is complete?

afoxman (approved):

tido64:
> It's been awhile, but I think I tried using this in the past with beachball. I ran into problems when beachball tried to merge changes. The branch-protections in GitHub blocked this token from being able to merge.
> 
> It seems like that shouldn't be a problem now, since the merge will come from a 2nd PR rather than automation... but I wanted to bring it up and see if you thought of anything that may break.

Yeah, I've been thinking about what could break. The thing I've been most worried about are tags, but looking at Changesets repo's config, they're also using the automatic token and tags work fine. We know they are using the GitHub provided token because the name is reserved, so you can't use it for anything else.

> If you go ahead with this, can you delete the GIT_TOKEN secret once the transition is complete?

I'm going to merge this PR, then merge the release PR. If everything works fine, I will remove the secret.

PR #927, opened 2021-12-02
ci: workaround for release PR skipping CI

tido64:
### Description

This is a workaround for `changeset version` adding `[skip ci]` to the commit message of the release PR, making it too easy to merge and skip publishing on CI.

Changesets wants us to set `commit` to `false` if we don't want this behaviour, but that means that we would have to manually `git add` and `git commit` the change files that were generated.

The `commit` option used to be separate for `add` and `version`, but Changesets merged the two back in 2.0. An issue has been opened to get the separate options back: https://github.com/atlassian/changesets/issues/688

This workaround should go away once it has been resolved.

### Test plan

1. Generate an empty change file: `yarn change --empty`
2. Examine `git log` and `git status` to verify that step 1 created a commit
3. Run the bump script: `yarn version:changesets`
4. Examine `git log` and `git status` to verify that step 3 _did not_ create a commit


scripts/version.js
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
```
afoxman, line 1: Maybe name this script more explicitly to keep it differentiated from other scripts? Something like disableChangesetsCommit or whatever you come up with.

afoxman (approved):

PR #932, opened 2021-12-03
feat(dep-check): all deps should be checked under --vigilant

tido64:
### Description

By default, dep-check does not check dependencies that are not listed under `capabilities`. It assumes that the authors know what they are doing and want to manage them manually. The behaviour is the same even when `--vigilant` is specified. This leads to inconsistencies because unconfigured packages can be updated while configured ones are skipped.

### Test plan

Tests should pass.


kelset (approved):
Seen in action this morning 👍 Great improvement to ensure that all the dependencies (even the ones not declared in a capability section) are aligned when using vigilant.

Issue #375, opened 2021-07-08
Add tests for @rnx-kit/cli


afoxman:
### Description

We need a test suite to cover all of the CLI code.

### Your environment

- Affected package: @rnx-kit/cli
- Version: *
- Tested platforms:
  - [x] Android
  - [x] iOS
  - [x] macOS
  - [x] Windows
- System information:
  ```
  # Run `npx react-native info` and paste the output here
  ```

### Steps to reproduce

yarn test

PR #934, opened 2021-12-05
CLI updates

### LINKED ISSUE #521: CI: fail bundling when type-checking fails
### Description

In CI loops, fail bundling when type-checking fails. This would have caught a bug in the type-checking code which only happened on windows (which I don't run during development).

### Your environment

- Affected package:
- Version:
- Tested platforms:
  - [ ] Android
  - [ ] iOS
  - [ ] macOS
  - [ ] Windows
- System information:
  ```
  # Run `npx react-native info` and paste the output here
  ```

afoxman:
### Description

The CLI's TypeScript support is currently limited to validating source files in one package -- the app being bundled/served. If the app depends on other packages in the same monorepo, sources from those packages won't be type-checked.

This PR expands the reach of the CLI, allowing it to type-check sources from all packages in the monorepo. External packages (anything under node_modules) are ignored.

The CLI buckets source files by their parent package, and uses the package's tsconfig when checking those files.

This PR also adds type-checking options, which vary between bundling and serving. Bundling, for example, will throw when validation fails, causing the entire bundling run to fail. Serving, will only print type errors to the console, and continue without interruption.

Resolves #521.

### Test plan

Existing CI tests pass. There isn't much test coverage in this area of the CLI. That issue is tracked under #375.

### DIFF
--- a/packages/cli/src/start.ts
+++ b/packages/cli/src/start.ts
@@ -5,7 +5,6 @@ import {
   loadMetroConfig,
   startServer,
 } from "@rnx-kit/metro-service";
-import { findConfigFile, Service } from "@rnx-kit/typescript-service";
 import chalk from "chalk";
 import type { Reporter, ReportableEvent } from "metro";
 import type Server from "metro/src/Server";
@@ -14,7 +13,7 @@ import path from "path";
 import readline from "readline";
 import { getKitServerConfig } from "./serve/kit-config";
 import { customizeMetroConfig } from "./metro-config";
-import type { TSProjectInfo } from "./types";
+import type { TypeScriptValidationOptions } from "./types";
 
 export type CLIStartOptions = {
   host: string;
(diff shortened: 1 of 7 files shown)


packages/cli/src/typescript/project-cache.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+    return projects[root][platform]!;
+  }
+
+  function clearPlatform(platform: AllPlatforms): void {
+    Object.values(projects).forEach((projectsByPlatform) => {
+      if (projectsByPlatform[platform]) {
+        projectsByPlatform[platform]!.dispose();
+        delete projectsByPlatform[platform];
+      }
```
tido64, line 144: Nit: Can we store a temp variable to avoid the nonnull assertion?
```suggestion
      const project = projectsByPlatform[platform]
      if (project) {
        project.dispose();
        delete projectsByPlatform[platform];
      }
```

packages/cli/src/start.ts
```diff
@@ -120,35 +119,17 @@ export async function rnxStart(
...
+    print: (message: string): void => {
+      terminal.log(message);
+    },
+  };
   customizeMetroConfig(
     metroConfig,
     serverConfig.detectCyclicDependencies,
     serverConfig.detectDuplicateDependencies,
-    tsprojectInfo,
+    serverConfig.typescriptValidation ? typescriptValidationOptions : false,
```
tido64, line 132: Isn't it possible to pass options here? Won't this ignore them?

packages/cli/src/typescript/project-cache.ts
afoxman: Fixed

packages/cli/src/start.ts
afoxman: Defaults -- print messages to the console, don't throw on a type-validation error.

tido64: Aren't we getting `serverConfig` from the config (`package.json`)? If the user sets an object there, we would lose it here.

tido64: Never mind. It looks like we only allow boolean values for the `typescriptValidation` option: https://github.com/microsoft/rnx-kit/blob/12fa18216536fa4cbe81aed918326079b6969fe5/packages/config/src/bundleConfig.ts#L33-L38

tido64 (approved):

PR #942, opened 2021-12-07
docs(dep-check): add more words about meta capabilities

tido64:
### Description

Describe meta capabilities and how to use them.

### Test plan

n/a


packages/dep-check/README.md
```diff
@@ -296,6 +296,24 @@ of capabilities instead:
 }
 ```
 
+We call these **meta** capabilities. Note that the name needs to be `#meta` to
+differentiate them from "regular" capabilities. Otherwise, you'll use them as
+any other capabilities. For instance:
```
kelset, line 301:
```suggestion
We call these **meta** capabilities. To identify them, the field name needs to be `#meta` - also, for a meta capability there's no `version` field, and the `capabilities` array is expected to be an array of other capabilities. To use the meta capability in a rnx-kit configuration in a package.json, there's nothing specific to be done - for instance:
```

Small rewording - probably will fail the format test.

PR #947, opened 2021-12-07
Fix Metro type-checking

afoxman:
### Description

This PR limits the files which are type-checked during bundling/serving. When Metro reports a file add/update, we will only check that file if TypeScript considers it to be part of the "source code" file listing. This ignores transpiled files.

This addresses problems found while bumping the CLI forward in #857.

### Test plan

CI tests pass.

### DIFF
--- a/packages/cli/src/typescript/project-cache.ts
+++ b/packages/cli/src/typescript/project-cache.ts
@@ -12,6 +12,11 @@ import {
 import path from "path";
 import ts from "typescript";
 
+export type ProjectInfo = {
+  tsproject: Project;
+  tssourceFiles: Set<string>;
+};
+
 /**
  * Collection of TypeScript projects, separated by their target platform.
  *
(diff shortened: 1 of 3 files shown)


packages/cli/src/typescript/project-cache.ts
```diff
@@ -119,35 +130,49 @@ export function createProjectCache(
...
   function clearPlatform(platform: AllPlatforms): void {
     Object.values(projects).forEach((projectsByPlatform) => {
-      const project = projectsByPlatform[platform];
-      if (project) {
-        project.dispose();
+      const info = projectsByPlatform[platform];
+      if (info) {
+        if (info.tsproject) {
+          info.tsproject.dispose();
+        }
```
tido64, line 168: Nit:
```suggestion
        info.tsproject?.dispose();
```

tido64 (approved):

packages/cli/src/typescript/project-cache.ts
afoxman: We aren't using this operator anywhere else in the CLI except for writing third party notices, and I don't have an engines field in package.json, so I'm going to skip on this one for now.

tido64: This is TypeScript though. Wouldn't it be transpiled away?

afoxman: That's' true, but I'm not sure if ?. gets replaced or not. Just being cautious.

PR #955, opened 2021-12-11
Add `single` command to RN changelog generator script and refactor 

lunaleaps:
### What issue are you trying to solve?
Create a new command to format a single commit as well as remove the token dependency for getting commit data

### How does this change address the issue?
* Makes the default behavior as a command 
* Create a `single` command
* Extract out util functions

### Test plan

Compared the following command before and after these changes: 
```
# before
node lib/index.js --repo ~/react-native --changelog ~/react-native/CHANGELOG.md --base 0.67-stable --compare main
# after
node lib/index.js --repo ~/react-native --changelog ~/react-native/CHANGELOG.md --base 0.67-stable --compare main --token [mytoken]
```
and verified tests pass via `yarn test:local`


packages/rn-changelog-generator/README.md
```diff
@@ -61,22 +61,20 @@ git pull
 popd
 ```
 
-Then create a GitHub ‘personal access token’ that can be used by the script to
-fetch commit metadata from the GitHub API.
-
-1. Visit the [token settings page](https://github.com/settings/tokens).
-1. Generate a new token and **only** give it the `public_repo` scope.
-1. Store this token somewhere secure for future usage.
-
```
NickGerleman, line 70: GitHub may rate limit requests not authenticated with a token. I've noticed this become a problem when running on some public CI services.

Would it be possible to keep the ability to use a token, but to let it be optional?

packages/rn-changelog-generator/src/generator.ts
```diff
@@ -7,9 +7,13 @@ import fs from "fs";
...
 import deepmerge from "deepmerge";
-import https from "https";
 import child_process from "child_process";
-import { IncomingHttpHeaders } from "http";
+import {
+  fetchCommits,
+  Commit,
+  getChangeMessage,
+  formatCommitLink,
+} from "./utils";
```
NickGerleman, line 16: Nit: "Utils" files can become a bit of a dumping ground for unrelated functionality. Are there other factorings that could avoid this?

packages/rn-changelog-generator/src/generator.ts
```diff
@@ -741,80 +617,16 @@ export async function getAllChangelogDescriptions(
...
-        describe:
-          "Verbose listing, includes internal changes as well as public-facing changes",
-        demandOption: false,
-        default: false,
-      },
-    })
-    .version(false)
-    .help("help").argv;
-
+export default function (argv: any) {
```
NickGerleman, line 629: Really like this change 👍

packages/rn-changelog-generator/src/single.ts
```diff
@@ -0,0 +1,7 @@
+import { fetchCommit, getChangeMessage } from "./utils";
+
+export default function (argv: any) {
```
NickGerleman, line 3: Can we do something typesafe?

packages/rn-changelog-generator/src/single.ts
```diff
@@ -0,0 +1,7 @@
...
+  fetchCommit(argv.commit).then((commitData) => {
```
NickGerleman, line 4: Can we do async/await for new code, instead of explicit promise chaining?

packages/rn-changelog-generator/test/utils.test.ts
```diff
@@ -0,0 +1,106 @@
+import https from "https";
+import { EventEmitter } from "events";
+import fs from "fs";
+import path from "path";
+import util from "util";
+
+const readFile = util.promisify(fs.readFile);
```
NickGerleman, line 7: Node 10+ (older than we care about) added support for a native promise-based API, `fs.promises`.

A pattern is to do something like:
`import {promises as fs} from "fs";`

packages/rn-changelog-generator/test/utils.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    requestEmitter.emit("response", responseEmitter);
+    readFile(path.join(__dirname, "__fixtures__", fixture), "utf-8").then(
+      (data) => {
+        responseEmitter.emit("data", data);
+        responseEmitter.emit("end");
+      }
+    );
+  });
+  return requestEmitter;
+}
```
NickGerleman, line 29: 💯

NickGerleman (approved):

tido64:
Can you run `yarn change` at the workspace root and add a change log?
image

packages/rn-changelog-generator/README.md
lunaleaps: Sounds good, can do that

lunaleaps: Updated to make token optional

lunaleaps:
Have addressed all the above comments and is ready for another look! 

Have verified that running the following is the same as on main: 
```
node lib/index.js --repo ~/react-native --changelog ~/react-native/CHANGELOG.md --base v0.66.4 --compare v0.67.0-rc.6 --token TOKEN 
```

and running `single` command looks like: 
```
node lib/index.js single -c 3bd11599c9d6aca5fa29ad291e9ad8b4a353bef4
https://api.github.com/repos/facebook/react-native/commits/3bd11599c9d6aca5fa29ad291e9ad8b4a353bef4
- Update bump-oss-version.js to guide releaser through release actions ([3bd11599c9](https://github.com/facebook/react-native/commit/3bd11599c9d6aca5fa29ad291e9ad8b4a353bef4) by [@lunaleaps](https://github.com/lunaleaps))
{
  doesNotFollowTemplate: false,
  changeCategory: 'general',
  changeType: 'failed',
  fabric: false,
  internal: true,
  turboModules: false
}
```

tido64 (approved):
LGTM

@NickGerleman: Just give me a thumbs-up and I'll merge this.

tido64:
FYI: https://github.com/microsoft/rnx-kit/releases/tag/%40rnx-kit/rn-changelog-generator%400.2.0

PR #967, opened 2021-12-17
Update the typescript react-native resolver

afoxman:
### Description

Change the order in which file extensions are searched. First, look for TypeScript files everywhere, then, look for JavaScript files. And, if JSON modules are enabled, look for them last. This multi-pass approach matches the way TypeScript resolves modules.

One important part of this PR is that now the resolver will always match on JavaScript files. It used to only do this when `checkJs` was enabled. That's not how TypeScript's resolver does it, so I've taken this condition out to better align with expected behavior.

### Test plan

Existing tests updated. All CI tests pass.

### DIFF
--- a/packages/typescript-react-native-resolver/src/host.ts
+++ b/packages/typescript-react-native-resolver/src/host.ts
@@ -6,16 +6,21 @@ import {
 import path from "path";
 import ts from "typescript";
 
-import { ExtensionsTypeScript, hasExtension } from "./extension";
 import {
-  ResolverLog,
-  ResolverLogMode,
+  ExtensionsJavaScript,
+  ExtensionsJSON,
+  ExtensionsTypeScript,
+} from "./extension";
+import {
   changeModuleResolutionHostToLogFileSystemReads,
   logModuleBegin,
   logModuleEnd,
+  ResolverLog,
+  ResolverLogMode,
 } from "./log";
 import { createReactNativePackageNameReplacer } from "./react-native-package-name";
 import { resolveFileModule, resolvePackageModule } from "./resolve";
+
 import type { ResolverContext, ModuleResolutionHostLike } from "./types";
 
 /**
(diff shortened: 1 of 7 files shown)


packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -168,33 +163,51 @@ export function resolveModuleNames(
...
+      "Searching for module '%s' with target file type 'TypeScript'",
+      finalModuleName
+    );
+    let module = resolveModuleName(
       context,
-      replaceReactNativePackageName(moduleName),
+      finalModuleName,
       containingFile,
-      extensions
+      ExtensionsTypeScript
```
tido64, line 185: Why do we have to call `resolveModuleName` three times instead of concatenating the extensions in the right order and calling it once? For instance:
```suggestion
      [
        ...ExtensionsTypeScript,
        ...ExtensionsJavaScript,
        ...(options.resolveJsonModule ? ExtensionsJSON : []),
      ]
```

packages/typescript-react-native-resolver/test/extension.test.ts
```diff
@@ -1,3 +1,4 @@
+import "jest-extended";
```
tido64, line 1: Is this needed? It looks like you just added the import without actually using it.

packages/typescript-react-native-resolver/src/host.ts
afoxman: I explained in the PR description... making multiple passes matches what TypeScript does internally. I want to be as close to their resolution scheme as I can be, with the addition of RN resolution logic.

packages/typescript-react-native-resolver/test/extension.test.ts
afoxman: Yes, there were red squiggles in VS Code, and these types resolve them. This doesn't actually import or load the code, this is just type info. The jest-extended library is loaded for all tests in this package (and a few others) by way of `setupFilesAfterEnv` in package.json (https://github.com/microsoft/rnx-kit/blob/main/packages/typescript-react-native-resolver/package.json).

tido64 (approved):

PR #977, opened 2022-01-04
fix(dep-check): bump react-native-test-app to 0.10.2

tido64:
### Description

Bumps react-native-test-app to 0.10.2 to fix issue with parsing version strings containing pre-release identifiers.

### Test plan

n/a


sveinar (approved):
LGTM

PR #981, opened 2022-01-06
fix(metro-config): prefer `module` main field

tido64:
### Description

By default, we should prefer `module` field over the current default ones.

### Test plan

n/a

### DIFF
--- a/.changeset/fifty-clocks-work.md
+++ b/.changeset/fifty-clocks-work.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-config": patch
+---
+
+Prefer `module` main field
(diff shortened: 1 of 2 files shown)


afoxman:
Does Metro support ES modules? This seems strange in the context of react-native since it is CJS-based.

tido64:
> Does Metro support ES modules? This seems strange in the context of react-native since it is CJS-based.

Metro just passes JS code to Babel. I don't think it's something that Metro has to explicitly support. Babel should transpile ESM to CJS, so this shouldn't cause any problems. Preferring ESM over CJS will make tree-shaking easier for esbuild.

afoxman (approved):

PR #982, opened 2022-01-06
feat(cli): show QR code for use with react-native-test-app

tido64:
### Description

Added command to show QR code. This QR code can be scanned in React Native Test App to load the bundle from the dev server, eliminating the need to manually configure the bundler address.

| macOS (iTerm) | Windows (Command Prompt)  | Windows (Terminal) |
|:-:|:-:|:-:|
| image | image | image |

### Test plan

```
cd packages/test-app
yarn build --dependencies
yarn start
```

- Once the dev server starts, press `a` to show the QR code.
- If you have the test app installed on a device, open it. Shake the device to open the dev menu, select "Scan QR Code", and point it at the QR code. The test app should start loading the bundle from your local machine.

### DIFF
--- a/packages/cli/src/start.ts
+++ b/packages/cli/src/start.ts
@@ -92,7 +92,7 @@ export async function rnxStart(
         [
           ["r", "reload the app"],
           ["d", "open developer menu"],
-          ["a", "show QR code"],
+          ["a", "show bundler address QR code"],
           ["ctrl-c", "quit"],
         ].forEach(([key, description]) => {
           terminal.log(press + key + dim(` to ${description}.`));
@@ -204,30 +204,13 @@ export async function rnxStart(
         switch (name) {
           case "a": {
             const protocol = cliOptions.https ? "https" : "http";
+            const host = cliOptions.host || os.hostname();
             const port = metroConfig.server.port;
-            const hosts = cliOptions.host
-              ? [cliOptions.host]
-              : Object.entries(os.networkInterfaces()).reduce<string[]>(
-                  (hosts, [name, intf]) => {
-                    // Skip interfaces used for tunneling, e.g. VPNs
-                    if (intf && !name.startsWith("utun")) {
-                      intf.forEach(({ address, family, internal }) => {
-                        if (family === "IPv4" && !internal) {
-                          hosts.push(address);
-                        }
-                      });
-                    }
-                    return hosts;
-                  },
-                  []
-                );
-            hosts.forEach((address) => {
-              const url = `${protocol}://${address}:${port}/index.bundle`;
-              qrcode.toString(url, { type: "terminal" }, (_err, qr) => {
-                terminal.log("");
-                terminal.log(url + ":");
-                terminal.log(qr);
-              });
+            const url = `${protocol}://${host}:${port}/index.bundle`;
+            qrcode.toString(url, { type: "terminal" }, (_err, qr) => {
+              terminal.log("");
+              terminal.log(url + ":");
+              terminal.log(qr);
             });
             break;
           }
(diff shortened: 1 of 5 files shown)


packages/cli/src/start.ts
```diff
@@ -199,14 +200,39 @@ export async function rnxStart(
...
-        terminal.log(chalk.green("Reloading app..."));
-        messageSocket.broadcast("reload", undefined);
-      } else if (name === "d") {
-        terminal.log(chalk.green("Opening developer menu..."));
-        messageSocket.broadcast("devMenu", undefined);
       } else {
-        terminal.log(chalk.grey(_key));
+        switch (name) {
+          case "a":
+            Object.entries(os.networkInterfaces()).forEach(([name, intf]) => {
```
afoxman, line 206: Why not use the known host/port/protocol to build a URL?

`cliOptions.host`
`cliOptions.https`
`metroConfig.server.port`

packages/cli/src/start.ts
```diff
@@ -85,17 +87,16 @@ export async function rnxStart(
...
-            "-" +
-            chalk.cyanBright("c") +
-            "'"
-        );
+        const dim = chalk.dim;
+        const press = dim(" › Press ");
+        [
+          ["r", "reload the app"],
+          ["d", "open developer menu"],
+          ["a", "show QR code"],
```
afoxman, line 95: Is `a` the normal key for this? Seems like `c` for code or `q` for qr (but that is ambiguous with quit).

packages/cli/src/start.ts
tido64: I've tried using them, but the only populated value I get here is the port:
```
$ react-native rnx-start
host:
https: undefined
port: 8081
```

packages/cli/src/start.ts
tido64: Yeah, I was thinking `q` initially but I didn't want to confuse people with `quit`, so I picked `a` as in address. I have mixed feeling about `c` since it's so close to `ctrl-c`. If you have strong opinion about this, I can make it `c`.

packages/cli/src/start.ts
afoxman: In the absence of a host name, 'localhost' is always used by convention. And https === undefined means use http.

packages/cli/src/start.ts
afoxman: Then maybe make the text `show address QR code`? I wouldn't have guessed address from a.

packages/cli/src/start.ts
afoxman: How about using `os.hostname()`?

This seems like it's going to go sideways on dev machines with complex network configs. E.g. multiple NICs, virtual NICs, etc.

afoxman: It will work, but it will show multiple IP/QR pairs. Letting DNS resolve things means you only need to show one. Not sure if the devices can use DNS or not... but if not, you could query DNS yourself to get your own IP, then use that.

packages/cli/src/start.ts
tido64: This triggers a Local Network Access permission dialog on my phone, but I guess that's okay?

packages/cli/src/start.ts
tido64: This triggers a Local Network Access permission dialog on my phone, but I guess that's okay?

afoxman: You could query DNS to get your IP address then publish that to avoid the network permission prompt -- which seems totally broken (using an IP address to bypass a security feature).

tido64: I really like the simplicity of `os.hostname()` though, so this is fine.

PR #993, opened 2022-01-13
Introduce a generic "react native clean" script in the RNX-kit CLI

dennismunene:
### Description
Introduce a generic "react native clean" script in the RNX-kit CLI. 

This script is responsible for clearing all caches and similar that are involved in React Native. 

This script will: 
- clear metro cache
- clear yarn cache
- clear cocoapods caches 
- clean android cache 
- clean ios cache
- clean macos cache
- clear haste-map 
- clear npm cache

### Test plan

**Options** 

**--incude**  -    Comma-separated flag of caches to clear e.g npm,yarn . When not specified , only non-platform specific caches are cleared. Supported options [android,cocoapods,npm,metro,watchman,yarn]. 
**--project-root** -  specifies the root directory to your react native project. 

To run the script : 
e.g for ios 
**yarn react-native rnx-clean --include npm,android  --project-root "PATH_TO_YOUR_RN_PROJECT"**

### DIFF
--- a/packages/cli/react-native.config.js
+++ b/packages/cli/react-native.config.js
@@ -235,7 +235,7 @@ module.exports = {
       description: "Clears React Native project related caches",
       options: [
         {
-          name: "--include [android,ios,haste,npm,macos,metro,react,watchman,yarn]",
+          name: "--include [android,cocoapods,npm,metro,watchman,yarn]",
           description:
             "Comma-separated flag of caches to clear e.g npm,yarn . When not specified , only non-platform specific caches are cleared.",
         },
--- a/packages/cli/react-native.config.js
+++ b/packages/cli/react-native.config.js
@@ -235,28 +235,14 @@ module.exports = {
       description: "Clears React Native project related caches",
       options: [
         {
-          name: "--platform [ios|android|macos]",
+          name: "--include [android,ios,haste,npm,macos,metro,react,watchman,yarn]",
           description:
-            "Target Platform to clean. When not specified , only non-module specific caches are cleared.",
-          parse: parsePlatform,
+            "Comma-separated flag of caches to clear e.g npm,yarn . When not specified , only non-platform specific caches are cleared.",
         },
         {
           name: "--project-root <path>",
-          description: "Root path to your React Native project (*Required)",
+          description: "Root path to your React Native project (optional)",
           parse: (val) => path.resolve(val),
-        },
-        {
-          name: "--best-effort [boolean]",
-          description:
-            "Best Effort.If set to false cleaning will stop when there is an error.",
-          default: true,
-          parse: parseBoolean,
-        },
-        {
-          name: "--keep-node-modules [boolean]",
-          description: "Specifies wether or not to remove npm node_modules",
-          default: false,
-          parse: parseBoolean,
         },
       ],
     },
--- a/packages/cli/react-native.config.js
+++ b/packages/cli/react-native.config.js
@@ -7,6 +7,7 @@ const {
   rnxDepCheckCommand,
   rnxTestCommand,
   rnxWriteThirdPartyNotices,
+  rnxClean,
 } = require("./lib/index");
 
 module.exports = {
@@ -225,6 +226,22 @@ module.exports = {
           name: "--additional-text [path]",
           description:
             "A string to append at the end of the generated licence file.",
+        },
+      ],
+    },
+    {
+      name: "rnx-clean",
+      func: rnxClean,
+      description: "Clear React Native Caches",
+      options: [
+        {
+          name: "--platform [string]",
+          description: "Platform to clean android | ios ",
+        },
+        {
+          name: "--project-root [string]",
+          description: "Root path to your React Native project",
+          parse: (val) => path.resolve(val),
         },
       ],
     },
--- a/packages/cli/react-native.config.js
+++ b/packages/cli/react-native.config.js
@@ -232,16 +232,31 @@ module.exports = {
     {
       name: "rnx-clean",
       func: rnxClean,
-      description: "Clear React Native Caches",
+      description: "Clears React Native project related caches",
       options: [
         {
-          name: "--platform [string]",
-          description: "Platform to clean android | ios ",
+          name: "--platform [ios|android|macos]",
+          description:
+            "Target Platform to clean. When not specified , only non-module specific caches are cleared.",
+          parse: parsePlatform,
         },
         {
-          name: "--project-root [string]",
-          description: "Root path to your React Native project",
+          name: "--project-root <path>",
+          description: "Root path to your React Native project (*Required)",
           parse: (val) => path.resolve(val),
+        },
+        {
+          name: "--best-effort [boolean]",
+          description:
+            "Best Effort.If set to false cleaning will stop when there is an error.",
+          default: true,
+          parse: parseBoolean,
+        },
+        {
+          name: "--keep-node-modules [boolean]",
+          description: "Specifies wether or not to remove npm node_modules",
+          default: false,
+          parse: parseBoolean,
         },
       ],
     },
--- a/packages/cli/src/clean.ts
+++ b/packages/cli/src/clean.ts
@@ -1,64 +1,281 @@
 import type { Config as CLIConfig } from "@react-native-community/cli-types";
 import path from "path";
+import chalk from "chalk";
 
 type Args = {
-  platform: "android" | "ios";
+  platform: "android" | "ios" | "macos";
   projectRoot: string;
+  bestEffort: boolean;
+  keepNodeModules: boolean;
 };
 
-const COMMAND_NAME = "rnx-clean";
+type CLICommand = {
+  name: string;
+  command: string;
+  args: string[];
+};
+
+type CachePath = {
+  name: string;
+  path: string;
+};
+
+const fs = require("fs");
+const os = require("os");
+const npm = os.platform() === "win32" ? "npm.cmd" : "npm";
+const yarn = os.platform() === "win32" ? "yarn.cmd" : "yarn";
 
 export function rnxClean(
   _argv: string[],
   _config: CLIConfig,
-  { platform, projectRoot }: Args
+  cliOptions: Args
 ): void {
-  const commandIndex = process.argv.indexOf(COMMAND_NAME);
-  if (commandIndex < 0) {
-    throw new Error("Failed to parse command arguments");
+  //validate root path
+  if (!dirExists(cliOptions.projectRoot)) {
+    console.log(
+      chalk.red("error:"),
+      "Invalid path provided! " + cliOptions.projectRoot
+    );
+
+    return;
   }
 
-  // Copy and remove the first arguments from `node react-native rnx-clean ...`
-  const argv = process.argv.slice(commandIndex + 1);
+  const cachePaths: CachePath[] = [
+    {
+      name: "React native cache",
+      path: "$TMPDIR/react-*",
+    },
+    {
+      name: "Metro cache",
+      path: "$TMPDIR/metro-*",
+    },
+    {
+      name: "Haste cache",
+      path: "$TMPDIR/haste-map-*",
+    },
+    {
+      name: "node_modules",
+      path: `${cliOptions.projectRoot}/node_modules`,
+    },
+  ];
 
-  const platformIndex = argv.indexOf("--platform");
-  if (platformIndex < 0) {
-    throw new Error("A target platform must be specified");
-  }
+  cachePaths.forEach((item) => {
+    if (item.name !== "node_modules") cleanDir(item);
+    else if (!cliOptions.keepNodeModules) cleanDir(item);
+  });
 
-  const projectRootIndex = argv.indexOf("--project-root");
-  if (projectRootIndex < 0) {
-    throw new Error("A project root must be specified");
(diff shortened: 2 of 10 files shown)


packages/cli/src/clean.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  const projectRootIndex = argv.indexOf("--project-root");
+  if (projectRootIndex < 0) {
+    throw new Error("A project root must be specified");
+  }
+
+  //nuke metro && install dependecies
+  execute(
+    "killall watchman && watchman watch-del-all && rm -rf $TMPDIR/react-* && rm -rf $TMPDIR/metro-* && rm -rf node_modules && yarn cache clean && yarn install",
```
tido64, line 36: I don't think we can assume that all projects use Yarn. It's also really weird that it installs when the command is clearly named "clean".

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,64 @@
...
+    projectRoot
+  );
+
+  switch (platform) {
+    case "ios":
+      execute(
+        'cd ios; killall Xcode ; xcrun -k && xcodebuild -alltargets clean && rm -rf "$(getconf DARWIN_USER_CACHE_DIR)/org.llvm.clang/ModuleCache" && rm -rf "$(getconf DARWIN_USER_CACHE_DIR)/org.llvm.clang.$(whoami)/ModuleCache" && rm -rf ~/Library/Developer/Xcode/DerivedData/* && rm -rf ~/Library/Caches/com.apple.dt.Xcode/* && rm -rf Pods && pod cache clean --all && rm -rf ~/.cocoapods && pod install && yarn start --reset-cache',
```
tido64, line 43: - We cannot assume that there is an `ios` folder in the current working directory
- We shouldn't call `rm -rf` when we can use JS APIs to remove files
- We shouldn't install or start anything
- We shouldn't kill Xcode as there may be unsaved changes

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,64 @@
...
+        projectRoot
+      );
+      break;
+    case "android":
+      execute(
+        "cd android; ./gradlew clean && yarn watch --reset-cache ",
```
tido64, line 49: - We cannot assume that there is an `android` folder in the current working directory
- We shouldn't start anything

packages/cli/react-native.config.js
```diff
@@ -228,5 +229,21 @@ module.exports = {
...
       ],
     },
+    {
+      name: "rnx-clean",
+      func: rnxClean,
+      description: "Clear React Native Caches",
+      options: [
+        {
+          name: "--platform [string]",
+          description: "Platform to clean android | ios ",
```
tido64, line 239: We should add a `parse` function that validates the input.

tido64 (commented):
This looks like a good start. Thanks for doing this. I think we should try to detect the folder structure, or at least check that files are where we think they are, add flags for some of the stuff we're removing, and be a bit more informative about the stuff we're about to remove.

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const platformIndex = argv.indexOf("--platform");
+  if (platformIndex < 0) {
+    throw new Error("A target platform must be specified");
+  }
+
+  const projectRootIndex = argv.indexOf("--project-root");
+  if (projectRootIndex < 0) {
+    throw new Error("A project root must be specified");
+  }
```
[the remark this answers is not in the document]
dennismunene: Cool , will remove this

packages/cli/src/clean.ts
dennismunene: Will make changes to factor in projects using npm

afoxman: You'll need an equivalent set of commands that you run on Windows platforms, too. These will only work on MacOS and Linux.

Also, can you separate each command into its own execute() call, and tun them serially? That way, JS will know which part of the process failed, and will have more control over how to handle failures. For example, devs could pass an arg that tells you to continue on error, making the cleanup best-effort instead of pass/fail. Also, it lets you add command-level logging in verbose mode which is useful in CI loops, should this be used there.

afoxman (commented):
Thank you so much for your contribution! I think this is a really useful command. I've added a few comments requesting additions/changes, but I like the approach.

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function execute(command: string, rootPath: string) {
+  const { spawnSync } = require("child_process");
+
+  spawnSync(command, {
+    cwd: path.resolve(rootPath),
+    stdio: "inherit",
+    shell: true,
```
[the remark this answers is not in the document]
dennismunene: Used it to be able to chain the commands. No need for it since am changing the approach.

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,64 @@
+import type { Config as CLIConfig } from "@react-native-community/cli-types";
+import path from "path";
+
+type Args = {
+  platform: "android" | "ios";
```
kelset, line 5: Related to @afoxman's comment below, this should not only run on macos and linux AND windows, but I think it should also be able to handle cleaning up the caches for react-native-windows (for macos I think the overlap with ios is complete... or at least mostly the same)

packages/cli/src/clean.ts
dennismunene: made changes, please review

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,64 @@
...
+        projectRoot
+      );
+      break;
+  }
+}
+
+function execute(command: string, rootPath: string) {
+  const { spawnSync } = require("child_process");
+
+  spawnSync(command, {
```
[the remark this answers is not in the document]
dennismunene: made updates

packages/cli/src/clean.ts
dennismunene: @kelset made some changes , are there any react-native-windows specific caches I'm missing ?

kelset: @NickGerleman should know - I don't work with RNW much so I'm not sure 😅 worst case we can proceed without this and have it be handled in a separate PR

NickGerleman: react-native-windows, and it's community modules, have dependencies similar to other npm dependencies with a native build. I.e. there are native build outputs emitted to the package itself. For RNW specifically, these are output to "/build" for build artifacts like object files, and "/target" for completed builds targets.

I think RNW community modules may follow a slightly different scheme, that aligns to the default MSBuild directories for a vcxproj. Iirc those would be something like <package>/<platform>/<configuration>/<project>.

I think nuking the node_modules package directory, and restoring from yarn cache, or reinstalling the dependencies, would be a generalized solution where we have these sorts of native outputs.

NickGerleman: react-native-windows also adds ".folly" and ".fmt" folders to node_modules, but it really shouldn't, and I would consider that a bug to be fixed RNW side.

NickGerleman: There will be artifacts in a similar form (vcxproj defaults) in the users project. RNW prior to 0.68 will also restore NuGet packages to that folder structure, according to "packages.config" conventions

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,281 @@
...
+            },
+            iosPath
+          );
+        }
+      );
+
+      break;
+    }
+    case "macos": {
+      findPath(cliOptions.projectRoot, "-macOS", (path: string) => {
```
tido64, line 142: This is also a bad indicator. Should we combine with iOS above and look for Podfiles instead? There is no reliable way to distinguish between the two.

packages/cli/react-native.config.js
```diff
@@ -228,5 +229,36 @@ module.exports = {
...
+            "Target Platform to clean. When not specified , only non-module specific caches are cleared.",
+          parse: parsePlatform,
+        },
+        {
+          name: "--project-root <path>",
+          description: "Root path to your React Native project (*Required)",
+          parse: (val) => path.resolve(val),
+        },
+        {
+          name: "--best-effort [boolean]",
```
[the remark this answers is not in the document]
tido64: It also looks like we're not consistently applying this flag. Maybe we should remove it for now and come back to it later.

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,281 @@
...
+      reject();
+    }
+
+    console.log(`${item.name} completed successfully!`);
+    resolve(task.status);
+  });
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function QueryPromise(promise: any) {
```
tido64, line 245: Please don't use `any`. Can we use generics instead?
```suggestion
function QueryPromise<T>(promise: Promise<T>) {
```

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,281 @@
...
+  result.isRejected = function () {
+    return isRejected;
+  };
+
+  return result;
+}
+
+function dirExists(directoryName: string) {
+  return !fs.existsSync(directoryName) ? false : true;
+}
```
tido64, line 281: Is there any reason for why we can't use `fs.existsSync` directly?

packages/cli/react-native.config.js
```diff
@@ -228,5 +229,36 @@ module.exports = {
         },
       ],
     },
+    {
+      name: "rnx-clean",
+      func: rnxClean,
+      description: "Clears React Native project related caches",
+      options: [
+        {
+          name: "--platform [ios|android|macos]",
```
tido64, line 238: Looking through the code, I realized that this is probably not a good flag. It's practically really hard to reliably distinguish ios files from macos ones. What do you do when they are merged in the same project? Can we change this to a `--include` flag that takes a comma-separated flag of things to remove? E.g.
```
--include gradle,npm,xcode
```
This will be clearer in intent, and we no longer need `--keep-node-modules`.

tido64 (changes requested):

packages/cli/react-native.config.js
tido64: Further, we should also think about how we can easily add new commands in the future. Right now, it's a bit all over the place. My thinking is that we can have something like this:
```javascript
const COMMANDS = {
  "gradle": [...],
  "npm": [
    {
      label: "Removing node_modules",
      action: () => {
        // remove node_modules
      }
    },
  ],
  "xcode": [...],
};

const categories = args.include.split(",");
categories.forEach((category) => {
  const commands = COMMANDS[category];
  if (!commands) {
    warn("Unknown category:", category);
    return;
  }

  commands.forEach(({label, action}) => {
    info(label);
    action();
  })
});
```
It maps nicely to proposed the `--include` flag, and it's easier to maintain and extend.

packages/cli/react-native.config.js
```diff
@@ -228,5 +229,36 @@ module.exports = {
...
+          description:
+            "Target Platform to clean. When not specified , only non-module specific caches are cleared.",
+          parse: parsePlatform,
+        },
+        {
+          name: "--project-root <path>",
+          description: "Root path to your React Native project (*Required)",
```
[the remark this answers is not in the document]
dennismunene: Yes , its possible , will update this

packages/cli/src/clean.ts
dennismunene: Yes , i had trouble finding a good way to differentiate the two , will combine this into one for both macos & ios

packages/cli/react-native.config.js
dennismunene: Makes sense. This approach is much clearer . Will go ahead and adopt this

packages/cli/react-native.config.js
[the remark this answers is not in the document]
dennismunene: Will do away with this for now , we can handle it later on a separate PR

packages/cli/src/clean.ts
dennismunene: @NickGerleman  Thank you for sharing. I will leave this as is as I'm already nuking node_modules. 
Would it make sense to have an option for clearing the build artifacts ? if so this might be something we can consider working on in a separate PR

packages/cli/src/clean.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+    case "ios": {
+      findPath(
+        cliOptions.projectRoot,
+        "LaunchScreen.storyboard",
+        (path: string) => {
+          const iosPath = path.substring(0, path.lastIndexOf("/"));
+
+          cleanDir({ name: "Pods Cache", path: `${iosPath}/Pods` });
+          cleanDir({ name: "CocoaPods cache", path: "~/.cocoapods" });
```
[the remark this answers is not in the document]
dennismunene: Wanted to be able to remove the Pods directory altogether ,  not sure if we need to remove `~./cocoapods`

tido64 (approved):
LGTM 👏 

You should add a change log by running `yarn change`, and get an approval from either @afoxman or @kelset.

kelset (changes requested):
Looks ok, can you just add to the README.md a brief section about it (similar to what the other commands have?)

kelset (approved):

PR #995, opened 2022-01-13
fix(metro-serializer-esbuild): forward `sideEffects` property

tido64:
### Description

Forward `sideEffects` property to help esbuild tree-shake more unused code.

### Test plan

Tests should pass.

### DIFF
--- a/packages/metro-serializer-esbuild/metro.config.js
+++ b/packages/metro-serializer-esbuild/metro.config.js
@@ -1,0 +1,19 @@
+const { exclusionList, makeMetroConfig } = require("@rnx-kit/metro-config");
+const { MetroSerializer, esbuildTransformerConfig } = require(".");
+
+// Metro will pick up mocks if we don't exclude them
+const blockList = exclusionList([/.*__fixtures__.*package.json/]);
+
+module.exports = makeMetroConfig({
+  reporter: { update: () => undefined },
+  resetCache: true,
+  resolver: {
+    blacklistRE: blockList,
+    blockList,
+  },
+  serializer: {
+    customSerializer: MetroSerializer(),
+    getPolyfills: () => [],
+  },
+  transformer: esbuildTransformerConfig,
+});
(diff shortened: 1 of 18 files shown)


packages/metro-serializer-esbuild/metro.config.js
```diff
@@ -0,0 +1,19 @@
+const { exclusionList, makeMetroConfig } = require("@rnx-kit/metro-config");
+const { MetroSerializer, esbuildTransformerConfig } = require(".");
+
+// Metro will pick up mocks if we don't exclude them
+const blockList = exclusionList([/.*__fixtures__.*package.json/]);
+
+module.exports = makeMetroConfig({
+  reporter: { update: () => undefined },
```
kelset, line 8: interesting, are you doing this to stop it from like spamming messages in console?

packages/metro-serializer-esbuild/test/index.test.ts
```diff
@@ -0,0 +1,174 @@
+import buildBundle from "@react-native-community/cli-plugin-metro/build/commands/bundle/buildBundle";
+import * as path from "path";
+
+describe("metro-serializer-esbuild", () => {
+  jest.setTimeout(60000);
```
kelset, line 5: any specific reasons for this timeout to be this long?

kelset (approved):
overall - looks great! The fact that you made a test that manages to test this is so cool :)

packages/metro-serializer-esbuild/test/index.test.ts
tido64: Because we have to run full bundles, a test can take many seconds to finish. On my machine, most take under 10 seconds, but CI timed out even with 30 seconds.

packages/metro-serializer-esbuild/metro.config.js
tido64: Yeah, the giant React logo in particular 😄

PR #999, opened 2022-01-19
feat(dep-check): allow setting `customProfiles` only

### LINKED ISSUE #998: dep-check: simplify config when running in vigilant mode with multiple custom profiles
Today, when running dep-check in vigilant mode, we can additionally tell it to use a custom profile with `--custom-profiles`. This is fine for the scenarios where we only need one custom profiles file. If we need multiple files, we have to add a config to the package. A valid configuration requires at least two fields, `reactNativeVersion` and `customProfiles`:

```js
{
  "rnx-kit": {
    "reactNativeVersion": "0.66",
    "customProfiles": "./myCustomProfile.js"
  }
}
```

The proposal is to drop the requirement on the `reactNativeVersion` field. We already provide all the necessary information via the `--vigilant` flag. It should be possible to reuse that information if we don't need to make any additional overrides.

tido64:
### Description

dep-check should pick up `customProfiles` when running in `--vigilant` mode to allow individual packages to use different profiles without having to re-declare which React Native versions they support.

Resolves #998.

### Test plan

Tests were added.

### DIFF
--- a/packages/dep-check/src/check.ts
+++ b/packages/dep-check/src/check.ts
@@ -44,6 +44,8 @@ export function getCheckConfig(
     capabilities: targetCapabilities,
     customProfiles,
   } = getKitCapabilities({
+    // React Native versions declared in the package's config should always
+    // override the ones specified with the `--vigilant` flag.
     ...(versions ? { reactNativeVersion: versions } : undefined),
     ...kitConfig,
   });
--- a/packages/dep-check/src/check.ts
+++ b/packages/dep-check/src/check.ts
@@ -13,7 +13,7 @@ import type { CheckConfig, CheckOptions, Command } from "./types";
 
 export function getCheckConfig(
   manifestPath: string,
-  { loose, uncheckedReturnCode = 0 }: CheckOptions
+  { loose, uncheckedReturnCode = 0, versions }: CheckOptions
 ): number | CheckConfig {
   const manifest = readPackage(manifestPath);
   if (!isPackageManifest(manifest)) {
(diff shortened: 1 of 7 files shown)


packages/dep-check/src/check.ts
```diff
@@ -43,7 +43,10 @@ export function getCheckConfig(
     kitType,
     capabilities: targetCapabilities,
     customProfiles,
-  } = getKitCapabilities(kitConfig);
+  } = getKitCapabilities({
+    ...(versions ? { reactNativeVersion: versions } : undefined),
```
afoxman, line 47: Should this be after `...kitconfig` so that it overrides `reactNativeVersion`?

packages/dep-check/src/types.ts
```diff
@@ -29,6 +29,7 @@ export type CheckConfig = {
 export type CheckOptions = Options & {
   uncheckedReturnCode?: number;
   config?: number | CheckConfig;
+  versions?: string;
```
afoxman, line 32: Is this meant to hold one version or multiple versions?

packages/dep-check/src/check.ts
tido64: This is intentional. The package's configuration should always override the versions specified with the `--vigilant` flag.

packages/dep-check/src/types.ts
tido64: This should be the string passed with `--vigilant` flag, e.g. `--vigilant 0.62,0.63,0.64`. And it does need to be parsed. Thanks for pointing it out 😄

afoxman (approved):

PR #1009, opened 2022-01-24
fix(metro-serializer-esbuild): correctly handle `sideEffects` array

tido64:
### Description

We need to manually handle `sideEffects` set to an array of globs as esbuild only accepts boolean values.

### Test plan

A test was added.

### DIFF
--- a/packages/metro-serializer-esbuild/src/index.ts
+++ b/packages/metro-serializer-esbuild/src/index.ts
@@ -3,8 +3,10 @@ import type { MetroPlugin } from "@rnx-kit/metro-serializer";
 import { findPackage } from "@rnx-kit/tools-node";
 import type { BuildOptions, BuildResult, Plugin } from "esbuild";
 import * as esbuild from "esbuild";
+import * as fs from "fs";
 import type { Dependencies, Graph, Module, SerializerOptions } from "metro";
 import type { SerializerConfigT } from "metro-config";
+import * as path from "path";
 import * as semver from "semver";
 
 export { esbuildTransformerConfig } from "./esbuildTransformerConfig";
@@ -27,8 +29,6 @@ function escapePath(path: string): string {
 }
 
 function fixSourceMap(outputPath: string, text: string): string {
-  const path = require("path");
-
   /**
    * All paths in the source map are relative to the directory
    * containing the source map.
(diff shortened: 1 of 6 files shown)


packages/metro-serializer-esbuild/src/index.ts
```diff
@@ -44,15 +44,38 @@ function fixSourceMap(outputPath: string, text: string): string {
...
 }
 
-function getSideEffects(modulePath: string): boolean | undefined {
-  const pkgJson = findPackage(modulePath);
-  if (!pkgJson) {
-    return undefined;
-  }
+const getSideEffects = (() => {
+  const pkgCache: Record<string, boolean | string[]> = {};
+  const getSideEffects = (pkgJson: string) => {
```
afoxman, line 49:
```suggestion
  const getSideEffectsWorker = (pkgJson: string) => {
```
Differentiating these names makes it a little easier to mentally parse, and avoids potential scoping bugs.

packages/metro-serializer-esbuild/src/index.ts
```diff
@@ -44,15 +44,38 @@ function fixSourceMap(outputPath: string, text: string): string {
...
+    if (!(pkgJson in pkgCache)) {
+      const content = fs.readFileSync(pkgJson, { encoding: "utf-8" });
+      const { sideEffects } = JSON.parse(content);
```
afoxman, line 52:
```suggestion
      const content = readPackage(pkgJson);
      const { sideEffects } = content;
```
From `@rnx-kit/tools-node`

packages/metro-serializer-esbuild/src/index.ts
```diff
@@ -44,15 +44,38 @@ function fixSourceMap(outputPath: string, text: string): string {
...
+  return (modulePath: string): boolean | undefined => {
+    const pkgJson = findPackage(modulePath);
+    if (!pkgJson) {
+      return undefined;
+    }
 
-  const { sideEffects } = require(pkgJson);
-  return sideEffects;
-}
+    const sideEffects = getSideEffects(pkgJson);
```
afoxman, line 73:
```suggestion
    const sideEffects = getSideEffectsWorker(pkgJson);
```

packages/metro-serializer-esbuild/src/index.ts
```diff
@@ -44,15 +44,38 @@ function fixSourceMap(outputPath: string, text: string): string {
...
+            absolute: true,
+          })
+          .map((p: string) => p.replace(/[/\\]/g, path.sep));
+      } else {
+        pkgCache[pkgJson] = sideEffects;
+      }
+    }
+    return pkgCache[pkgJson];
+  };
+  return (modulePath: string): boolean | undefined => {
```
afoxman, line 67: This takes an argument, but it will always be undefined because of the `()` on line 78. Which one is the right behavior?

packages/metro-serializer-esbuild/src/index.ts
```diff
@@ -44,15 +44,38 @@ function fixSourceMap(outputPath: string, text: string): string {
...
+    const pkgJson = findPackage(modulePath);
+    if (!pkgJson) {
+      return undefined;
+    }
 
-  const { sideEffects } = require(pkgJson);
-  return sideEffects;
-}
+    const sideEffects = getSideEffects(pkgJson);
+    return Array.isArray(sideEffects)
```
afoxman, line 74: This return is hard to make sense of. It's a boolean if `sideEffects` is an array, and it reflects whether or not the module is in the array. Or, if it's not an array, it's whatever you found which could be truthy or falsey.

How will the caller know what the result means?

packages/metro-serializer-esbuild/src/index.ts
tido64: The `()` on line 78 returns a wrapper around the function on line 67 so the cache doesn't leak outside this scope.

packages/metro-serializer-esbuild/src/index.ts
tido64: https://webpack.js.org/guides/tree-shaking/ <-- This is the only place I could find that documents how `sideEffects` works. If you set it to a boolean, it will apply to all files in the package. The alternative is to list the files that do have side effects. The unlisted ones do not have side effects.

packages/metro-serializer-esbuild/src/index.ts
tido64: Yup, I've just added them:
https://github.com/microsoft/rnx-kit/blob/3656ef0b7f10cdb860f7221fbea4f9631c203d40/packages/metro-serializer-esbuild/src/index.ts#L46-L55

afoxman (approved):

PR #1017, opened 2022-01-28
feat(dep-check): allow declaring capabilities at root level

### LINKED ISSUE #1015: dep-check: Make it simpler to declare packages/versions across profile versions
To declare a dependency that does not change between profile versions today, one would have to add it to every version:

```js
const commonDeps = {
  "prettier": {
    "name": "prettier",
    "version": "^2.5.1"
  }
};

module.exports = {
  "0.62": {
    ...commonDeps,
    // 0.62 specific dependencies here
  },
  "0.63": {
    ...commonDeps,
    // 0.63 specific dependencies here
  },
  "0.64": {
    ...commonDeps,
    // 0.64 specific dependencies here
  }
};
```

It should be possible to declare the common dependency at the root level. Below should be equivalent to the above:

```js
module.exports = {
  "prettier": {
    "name": "prettier",
    "version": "^2.5.1"
  },
  "0.62": {
    // 0.62 specific dependencies here
  },
  "0.63": {
    // 0.63 specific dependencies here
  },
  "0.64": {
    // 0.64 specific dependencies here
  }
};
```

tido64:
### Description

Allow declaring common capabilities at the root level.

Resolves #1015.

### Test plan

Tests were added.

### DIFF
--- a/packages/dep-check/README.md
+++ b/packages/dep-check/README.md
@@ -248,6 +248,25 @@ module.exports = {
 };
 ```
 
+If you have capabilities that should be the same across all versions, you can
+declare them at the root level like below:
+
+```js
+module.exports = {
+  "my-capability": {
+    name: "my-module",
+    version: "1.0.0",
+  },
+  0.64: {
+    // This entry will override the common version
+    "my-capability": {
+      name: "my-module",
+      version: "1.1.0",
+    },
+  },
+};
+```
+
 For a more complete example, have a look at the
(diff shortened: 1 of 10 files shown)


packages/dep-check/README.md
```diff
@@ -248,6 +248,25 @@ module.exports = {
...
+declare them at the root level like below:
+
+```js
+module.exports = {
+  "my-capability": {
+    name: "my-module",
+    version: "1.0.0",
+  },
+  0.64: {
+    // This entry will override the common version
```
kelset, line 261: good thing that you are clarifying the overriding policy here 👍

kelset (approved):

PR #1044, opened 2022-02-02
fix(dep-check): use react-native-test-app 1.0 from react-native 0.66+

tido64:
### Description

Use react-native-test-app 1.0 from react-native 0.66 on.

### Test plan

1. Build dependencies:
   ```
   cd example/test-app
   yarn build --dependencies
   ```
2. Test Android: `yarn android`
3. Test iOS: `yarn ios`
4. Test macOS: `yarn macos`
5. Test Windows: `yarn windows`


packages/test-app/package.json
```diff
@@ -15,10 +15,10 @@
...
     "bundle:macos": "react-native rnx-bundle --platform macos",
     "bundle:windows": "react-native rnx-bundle --platform windows",
-    "android": "react-native run-android",
-    "ios": "react-native run-ios",
-    "macos": "react-native run-macos --scheme ReactTestApp",
-    "windows": "react-native run-windows --sln windows\\SampleCrossApp.sln",
+    "android": "react-native run-android --no-packager",
+    "ios": "react-native run-ios --no-packager",
+    "macos": "react-native run-macos --scheme ReactTestApp --no-packager",
+    "windows": "react-native run-windows --no-packager --sln windows\\SampleCrossApp.sln",
```
kelset, line 21: you are adding the `--no-packager` flag here so that we're "forced" to always use the yarn start below (that uses the `rnx-start`) below?

.github/workflows/pr.yml
```diff
@@ -94,6 +94,11 @@ jobs:
     name: "Build Android"
     runs-on: ubuntu-latest
     steps:
+      - name: Set up JDK
+        uses: actions/setup-java@v2.5.0
+        with:
+          distribution: temurin
+          java-version: 11
```
kelset, line 101: potentially something that we might want to do in another PR (and potentially in the RNTA repo), maybe we should consider adding a script such as the one Mike Hardy shows here https://github.com/reactwg/react-native-releases/discussions/13#discussioncomment-2076731 to our tools? To ensure that you are on the right version of JDK

kelset (approved):
left a couple comments but no blockers

.github/workflows/pr.yml
tido64: The bump to 1.0 requires JDK 11 though. It was one of the breaking changes that was introduced. RNTA has been using temurin for quite some time already: https://github.com/microsoft/react-native-test-app/commit/4407e8f63193efbba0f86ee4ea096c7eca29222f

Adding that script sounds like a good thing to do though.

packages/test-app/package.json
tido64: Yeah, I find the packager launch annoying. In most cases, I either already have one running or don't want to one.

.github/workflows/pr.yml
```diff
@@ -112,8 +112,10 @@ jobs:
         env:
           CI_SKIP_GO: 1
       - name: Build Android app
+        env:
+          JVM_ARGS: org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
         run: |
-          ./gradlew clean build
+          ./gradlew --stacktrace clean build
```
kelset, line 118: do you still need stacktrace here or was it just for debugging?

tido64: It's useful for debugging and doesn't affect the output unless there's an error. I added it initially for debugging the cause of lint task failure, but I think it's useful to keep it in.

PR #1098, opened 2022-02-14
fix(dep-check): bump packages for react-native 0.68

tido64:
### Description

- react-native-gestures 2.3.0 works with react-native 0.68
- react-native-reanimated 2.5.0 supports react-native 0.68
- react-native-screens 3.13.1 now supports Fabric

### Test plan

n/a


packages/dep-check/src/profiles/profile-0.68.ts
```diff
@@ -37,6 +37,10 @@ const profile: Profile = {
     version: "^0.68.0-0",
     capabilities: ["core"],
   },
+  animation: {
+    name: "react-native-reanimated",
+    version: "^2.4.1",
```
tido64, line 42: react-native 0.68 isn't out yet so we're going to wait a bit to see if a new official version gets released before then.

packages/dep-check/src/profiles/profile-0.68.ts
```diff
@@ -37,11 +37,19 @@ const profile: Profile = {
     version: "^0.68.0-0",
     capabilities: ["core"],
   },
+  animation: {
+    name: "react-native-reanimated",
+    version: "^2.5.0",
```
kelset, line 42: do you think it's fine to merge as is, or we should wait until 0.68.1 comes out with this fix? https://github.com/reactwg/react-native-releases/discussions/17#discussioncomment-2463502 apparently it's needed for reanimated+newarch

tido64: If it's only needed for new-arch, I think we should merge it now. It's an opt-in, experimental feature at this point.

kelset (approved):

PR #1105, opened 2022-02-16
feat(rn-changelog-generator): Prepend changelogs to the supplied file

gabrieldonadel:
### Description

This updates the `rn-changelog-generator` to prepend the newly generated changelog to the supplied --changelog file instead of relying on redirecting the terminal output to a file. It also updates the `rn-changelog-generator` readme to point to the new `@rnx-kit/rn-changelog-generator`  package

Resolves https://github.com/facebook/react-native/issues/33011  
 

### Test plan

Run `node lib/index.js --base v0.66.4 --compare v0.67.0-rc.0 --repo ../../../react-native/ --changelog ../../../react-native/CHANGELOG.md` and ensure the file provided is prepended

### DIFF
--- a/packages/rn-changelog-generator/src/generator.ts
+++ b/packages/rn-changelog-generator/src/generator.ts
@@ -514,8 +514,18 @@ function handler(argv: GenerateArgs) {
     .then(async () => {
       const existingChangelogData = fs.readFileSync(argv.changelog, "utf-8");
       const base = await getOffsetBaseCommit(gitDir, argv.base, argv.compare);
-      const data = await run({ ...argv, base, gitDir, existingChangelogData });
-      return console.log(data);
+      const newChangeLogData = await run({
+        ...argv,
+        base,
+        gitDir,
+        existingChangelogData,
+      });
+      const fd = fs.openSync(argv.changelog, "w+");
+      const insert = Buffer.from(newChangeLogData + existingChangelogData);
+      fs.writeSync(fd, insert, 0, insert.length, 0);
+      fs.close(fd, (err) => {
+        if (err) throw err;
+      });
     })
     .catch((e) => {
       console.error(e);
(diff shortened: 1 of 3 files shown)


gabrieldonadel:
@lunaleaps would love your input here as well, just to ensure this is the expected behaviour

packages/rn-changelog-generator/src/generator.ts
```diff
@@ -514,8 +514,18 @@ function handler(argv: GenerateArgs) {
...
-      const data = await run({ ...argv, base, gitDir, existingChangelogData });
-      return console.log(data);
+      const newChangeLogData = await run({
+        ...argv,
+        base,
+        gitDir,
+        existingChangelogData,
+      });
+      const fd = fs.openSync(argv.changelog, "w+");
+      const insert = Buffer.from(newChangeLogData + existingChangelogData);
```
afoxman, line 524: Instead of concatenating them, why not call writeSync twice?

tido64:
The "Review" check is failing because of a missing change log file. Can you run `yarn change` to add one? 🙏

packages/rn-changelog-generator/src/generator.ts
gabrieldonadel: Sure, I can update it to do so but wouldn't that require another system call?

afoxman: Yes, it will issue 2 system calls.

lunaleaps, line 524: Do we need any sort of newline or separator here?

lunaleaps:
In terms of tackling the issue, this looks good to me! This work will remove one manual step in using the changelog generator. Thank you @gabrieldonadel for working on it!

packages/rn-changelog-generator/src/generator.ts
gabrieldonadel: @afoxman I had to change the logic a bit to take into consideration the header of this file before prepending the changelog otherwise this would still require manual work to move the `# Changelog` to the top of the file. I've also updated it to use multiple write's calls instead of concatenating strings

gabrieldonadel: I don't think so because `newChangeLogData` already contains `\n` `s at the end of the string

PR #1126, opened 2022-02-22
feat(jest-preset): use `@swc/jest` if it's available

tido64:
### Description

Use `@swc/jest` if it's available. This PR is currently in draft so we can measure the difference in build times on CI. We need to find another way to enable it to avoid breaking existing code bases.

### Test plan

CI should pass.


tido64:
Moved the branch to my own fork. Will open a new PR if necessary.

PR #1135, opened 2022-02-23
feat(patcher-tool): adding new package, patcher-tool

kelset:
# WORK IS STILL IN PROGRESS

### Description

This PR wants to add a new rnx-kit package to the monorepo, called patcher-tool.

This tool is an updated & cleaned up version of https://github.com/mganandraj/office-android-patches

### Test plan

Testing locally against react-native-macos local clone.


packages/patcher-tool/src/diffRepos.ts
```diff
@@ -0,0 +1,191 @@
...
+  } else {
+    options.whitelistDirs.forEach((dir) => {
+      if (
+        options.blacklistDirs.includes(
+          dir.startsWith(".\\") ? dir.substr(2) : dir
+        )
+      ) {
+        log.info(
+          "diffRNFork",
+          `${dir} is present in both whitelist as well as blacklist. Ignoring it.`
```
rasaha91, line 176: we should probably update uses of `whitelist` and `blacklist` to be `inclusionList` and `exclusionList`

packages/patcher-tool/src/patch/apply.ts
```diff
@@ -118,7 +118,7 @@ function applyPatch(
 
   // modifying the file in place
   let fileContents = "";
-  let mode = 0;
+  let mode = 0o666; // For files not yet created, give them read/write permissions
```
tido64, line 121: Do we really need this? The mode should be inherited by the parent folder. If we write to existing files, the mode should be kept intact by the file system.

kelset: This is from @rasaha91's PR, so I'll let him speak about it.

packages/patcher-tool/test/index.test.ts
```diff
@@ -0,0 +1,7 @@
+// TODO: add actual tests!
+
+describe("HelloWorld", () => {
+  it("should return a string", () => {
+    expect("Hello World!").toBe("Hello World!");
+  });
+});
```
kelset, line 7: I'm just leaving this fake test here as a reminder of the fact that proper tests should be added

packages/dep-check/README.md
```diff
@@ -161,51 +161,51 @@ resolve to:
 <!-- The following table can be updated by running `yarn update-readme` -->
 <!-- @rnx-kit/dep-check/capabilities start -->
 
-| Capability                           | 0.68                                                              | 0.67                                                              | 0.66                                                              | 0.65                                                              | 0.64                                                              | 0.63                                                              | 0.62                                                              | 0.61                                                              |
```
kelset, line 164: @tido64 I'm not sure why but when running `yarn update-readme` these changes have also been created...? Should I revert this?

tido64: Weird… Can you try rebasing on top of latest `main`?

kelset: it's ok, the tool on CI fixed it 🤷‍♂️

afoxman:
I'll take a look today. Is this just for applying patches to out-of-tree RN platform impls? Just macos? If so, maybe change the name to reflect that -- 'patch-react-native-package' or 'patch-react-native-macos' (if it is limited to macos).

When I skimmed this before, it was a collection of .bat files. If that is still true, add a note in the readme that this only runs on Windows dev boxes.

kelset:
> I'll take a look today. Is this just for applying patches to out-of-tree RN platform impls? Just macos? If so, maybe change the name to reflect that -- 'patch-react-native-package' or 'patch-react-native-macos' (if it is limited to macos).

It's a `react-native-macos` specific tool atm.

> When I skimmed this before, it was a collection of .bat files. If that is still true, add a note in the readme that this only runs on Windows dev boxes.

No that's only the `local-scripts` folder, the rest can be run on all platforms, it's standard TS code

mganandraj:
> 
The code is not bound to react native. It can be used to diff between any repo and a dirty fork. It is an extension of the open source patch package. The objective was to diff between fb/RN and ms/RNM , and identify all changes, file additions and deletions, and traverse the diffs easily, and efficiently act on the forks. Essentially, the tool traverses the directory structure, and diffs each files and creates patch files keeping the same directory structure as the original repo.

packages/patcher-tool/src/patch/apply.ts
mganandraj: An alternative fix would have been to not send mode to fs.writeFileSync if the value is '0', so that the file will inherit parent's mode .. @rasaha91

packages/patcher-tool/local-scripts/CopyToRepo.bat
```diff
@@ -0,0 +1,27 @@
...
+
+REM Copy bundle
+robocopy %SOURCEDIR%\bundle %TARGETREPO%\android-patches\bundle /S
+
+REM Copy dist in case we need to debug from the repo.
+IF '%COPYDIST%'=='1' (robocopy %SOURCEDIR%\dist %TARGETREPO%\android-patches\dist  /S)
+else (rmdir /S /Q %TARGETREPO%\android-patches\dist)
+
+REM Copy grouped patches.
+robocopy %SOURCEDIR%\patches-droid-office-grouped %TARGETREPO%\android-patches\patches-droid-office-grouped /S
```
mganandraj, line 15: Many of these scripts are no more relevant.. but we could do the clean ups later ..

packages/patcher-tool/src/patch/apply.ts
rasaha91: Sure, inheriting from the parent seems fine to me. Might be cleaner to initialize mode to the parent's permissions and always pass in mode, rather than having multiple calls to `writeFileSync`.

mganandraj (approved):

afoxman:
> > 
> 
> The code is not bound to react native. It can be used to diff between any repo and a dirty fork. It is an extension of the open source patch package. The objective was to diff between fb/RN and ms/RNM , and identify all changes, file additions and deletions, and traverse the diffs easily, and efficiently act on the forks. Essentially, the tool traverses the directory structure, and diffs each files and creates patch files keeping the same directory structure as the original repo.

Makes sense. Maybe "patch-repo"?

packages/patcher-tool/src/filterAndGroupPatches.ts
```diff
@@ -0,0 +1,401 @@
...
+
+if (fse.existsSync(filteredGroupedPatchStore)) {
+  logger.error("Output directory exists !");
+  process.exit();
+}
+
+fse.ensureDirSync(filteredGroupedPatchStore);
+
+// Known groups
+const OfficeRNHostDir = "OfficeRNHost";
```
afoxman, line 38: A general patching tool won't have groups like this... it will discover them automatically (or not bother to group anything).

If this is too valuable to lose, maybe start with this package as a diffing/patching facility for react-native-macos, and then factor out the interesting pieces into a more generalized diffing/patching tool.

packages/patcher-tool/src/filterAndGroupPatches.ts
```diff
@@ -0,0 +1,401 @@
+import fse from "fs-extra";
```
afoxman, line 1: Is this file used anywhere? If not, I'd remove it. I started leaving comments in here about how it is RN-specific. Then I stopped and did a search and I can't find a reference to it.

packages/patcher-tool/src/fs_utils.ts
```diff
@@ -0,0 +1,244 @@
...
+    callbackFile(path, rootAbsPath);
+  } else if (isDirectory(path)) {
+    const children = fs.readdirSync(path);
+    children.forEach((childpath: string) => {
+      const absChildPath = fs_path.resolve(path, childpath);
+      const relChildPath = relPath + fs_path.sep + childpath;
+
+      // Ignore the '.\' prefix when doing black list comparison
+      if (
+        exclusionListDirs.includes(
```
afoxman, line 114: I've seen this pattern repeated a few times. Can the exclusion list be pre-processed to exclude ./ or .\ once, ahead of time?

packages/patcher-tool/src/fs_utils.ts
```diff
@@ -0,0 +1,244 @@
...
+) {
+  const absChildPath = fs_path.resolve(path, relativePath);
+  if (!fs.existsSync(absChildPath)) {
+    callbackOnMiss(absChildPath);
+  } else {
+    callbackOnHit(absChildPath);
+  }
+}
+
+export function initDirectory(path: string) {
```
afoxman, line 161: Maybe rename to something which reflects the destructive nature of this function, like eraseAndRecreateDirectory?

packages/patcher-tool/src/logger.ts
```diff
@@ -0,0 +1,113 @@
...
+    }),
+  ],
+});
+
+function setLogFolder(logFolder: string) {
+  // logFolder = fs_path.resolve(logFolder, `${Date.now()}`);
+
+  logger.add(
+    new winston.transports.File({
+      filename: "error.log",
```
afoxman, line 35: These file names should be mentioned in the README as part of the CLI's output.

packages/patcher-tool/src/patch/apply.ts
```diff
@@ -0,0 +1,295 @@
...
+        }
+        break;
+      case "rename":
+        if (dryRun) {
+          // TODO: see what patch files look like if moving to exising path
+          if (!fs.existsSync(eff.fromPath)) {
+            throw new Error(
+              "Trying to move file that doesn't exist: " + eff.fromPath
+            );
+          }
```
afoxman, line 34: What about a check to see if the target exists? If it does, that seems like a problem.... by default, `moveSync` fails if the target exists already.

packages/patcher-tool/src/patch/apply.ts
```diff
@@ -0,0 +1,295 @@
...
+import { log } from "../logger";
+
+export const executeEffects = (
+  effects: ParsedPatchFile,
+  targetFilePathOverride: string, // Override the target path in the patch file.
+  { dryRun }: { dryRun: boolean }
+) => {
+  effects.forEach((eff) => {
+    switch (eff.type) {
+      case "file deletion":
```
afoxman, line 15: The dryRun checks seem valuable in the non-dryrun case. Consider running them all the time, and then only doing the destructive ops when !dryRun.

packages/patcher-tool/src/patch/apply.ts
```diff
@@ -0,0 +1,295 @@
...
+        }
+        break;
+      case "patch":
+        log.info("patch\\apply", "Patches found.");
+        applyPatch(eff, targetFilePathOverride, { dryRun });
+        break;
+      case "mode change":
+        const currentMode = fs.statSync(eff.path).mode;
+        if (
+          ((isExecutable(eff.newMode) && isExecutable(currentMode)) ||
```
afoxman, line 63: What about other mode changes, like read and write?

packages/patcher-tool/src/patch/apply.ts
```diff
@@ -0,0 +1,295 @@
+import fs from "fs-extra";
```
afoxman, line 1: Can this file-level patching code be replaced with calls to git-apply?

packages/patcher-tool/src/patch/read.ts
```diff
@@ -0,0 +1,70 @@
...
+import { readFileSync } from "fs-extra";
+// import { relative, resolve } from "path";
+// import { normalize } from "path";
+// import {PackageDetails} from '../PackageDetails';
+import { parsePatchFile, PatchFilePart } from "./parse";
+import { log } from "../logger";
+
+export function readPatch({
+  patchFilePath,
+}: // packageDetails,
```
afoxman, line 11: There's commented-out code here. Does any of it need to be kept?

packages/patcher-tool/src/patch_utils.ts
```diff
@@ -0,0 +1,116 @@
...
+) {
+  const patchArgs = ["-i", patchPath, targetPath, "-s"];
+  if (reverse) {
+    patchArgs.push("-R");
+  }
+
+  const patch = spawn(patchExecutable, patchArgs);
+  log.info(
+    "Patch",
+    "Calling C:\\Program Files\\Git\\usr\\bin\\patch.exe " +
```
afoxman, line 59: This path shouldn't be hardcoded -- it should come from `patchExecutable`.

packages/patcher-tool/src/patch_utils.ts
```diff
@@ -0,0 +1,116 @@
...
+  targetFilePathOverride, // Override the target file path in the patch file.
+  reverse,
+}: {
+  patchFilePath: string;
+  targetFilePathOverride: string;
+  reverse: boolean;
+}): boolean {
+  const patch = readPatch({ patchFilePath });
+  try {
+    executeEffects(
```
afoxman, line 93: Git can generate and apply file-level patches. Why not use that, vs writing our own impl of parsing and applying patch files?

afoxman (commented):
Lots of code here. It looks like some of it can be removed from the package and left behind (.bat files, commented out sections, some RN-specific patch grouping code).

A big question I have is why implement file-level patching in JS, when you could call git-diff and git-apply?

packages/patcher-tool/src/patch/apply.ts
kelset: @tido46 @mganandraj @rasaha91 is this something that can be addressed separately or that we have to fix in this PR?

scripts/update-readme.js
```diff
@@ -18,8 +18,9 @@ const repoUrl = (() => {
...
 
 const repoRoot = getWorkspaceRoot();
-const packages = getAllPackageJsonFiles(repoRoot).reduce(
-  (packages, manifestPath) => {
+const packages = getAllPackageJsonFiles(repoRoot)
+  .filter((p) => !p.includes("packages/draft-"))
+  .reduce((packages, manifestPath) => {
     const content = readFile(manifestPath);
     const manifest = JSON.parse(content);
     const { private, name, description } = manifest;
```
tido64, line 26: I think it would be better if we had a `draft` property in `package.json` rather than relying on a naming convention.

packages/patcher-tool/src/patch/apply.ts
tido64: I think we should at the very least default to `0o644`.

kelset: ok, changed to `0o644`

scripts/update-readme.js
afoxman: I definitely want to see the draft- prefix in the name of the package to let devs know it is not safe, but I'm also ok with adding a new package.json prop to guide our automatic doc generation scripts. 

Please don't put it in rnx-kit. That is reserved for kit configuration. Maybe a top-level field named something like `excludeFromDocs` or something like unique like that.

packages/draft-patch-rnmacos/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@rnx-kit/patch-rnmacos",
```
afoxman, line 2: The package should be published as `draft-patch-rnmacos`. When it meets the quality bar, it can be re-released as a normal package with a new line of versioned releases.

packages/draft-patch-rnmacos/src/patch/read.test.ts.nope
```diff
@@ -0,0 +1,209 @@
+import { readPatch } from "./read"
```
afoxman, line 1: For these .nope files...

Do these run successfully in the other repo? 

If they don't work in the old repo, they're broken and don't have much value. I wouldn't check them in here at all.

If they do run, then they will probably work with a few tweaks here and there, and they are worth keeping. Can you put them into a "broken-tests" folder so it's clear that we have tests we can start with?

afoxman (commented):
This is looking good. Just a few comments. Also, please don't squash your commits. It makes it impossible to see "changes since last review". (GH will squash for you when your PR goes through to main).

scripts/update-readme.js
kelset: I'm not a big fan of custom top level properties 😓 can we revisit this conversation separately from this PR?

tido64: 1. Like Adam said, a top-level property next to `private` should be good.
2. I'm not going to block the PR because of this. It should be a quick fix, so I'll let you decide whether you want to do a follow-up.

afoxman (approved):

PR #1165, opened 2022-03-01
chore: migrate to npm

tido64:
### Description

Migrate the repo to npm 8.

The main motivation for this move is to get better cacheability (more stable) than Yarn Classic, which randomly gets corrupted and have to be force invalidated. There is also a desire to move back to npm at some point when [isolated mode](https://github.com/npm/statusboard/issues/414) gets implemented.

The alternative would be to [upgrade to Yarn 3](https://github.com/microsoft/rnx-kit/pull/1177), which is less intrusive, but comes with its own set of issues.

### Test plan

CI should pass.


packages/dep-check/package.json
```diff
@@ -19,7 +19,8 @@
     "directory": "packages/dep-check"
   },
   "scripts": {
-    "build": "rnx-kit-scripts build && rnx-kit-scripts bundle",
```
tido64, line 22: npm does _not_ like that we do `&&` here for some reason.

.gitignore
```diff
@@ -3,16 +3,18 @@
 *.log
 *.tgz
 *.xcworkspace/
+.DS_Store
 .gradle/
 .idea/
-Pods/
-coverage/
+.yarnrc
```
fernandopasik, line 9: In the spirit of ignoring all yarn files, perhaps would be idea to add the folder `.yarn/*` (come with yarn 2/3)

Meshl-555 (approved):

PR #1171, opened 2022-03-02
Do not remap externals in esbuild-plugin-import-path-remapper

steffenvv:
### Description

The onResolve hook in the esbuild plugin has the potential to override the resolution of externals. The esbuild-plugin-import-path-remapper plugin accidentally does this by potentially intercepting the resolution of @foo/bar and translating it into something like @foo/bar/src/index.ts. This should not be done if "@foo/bar" is in the list of externals in the build config.

### Test plan

Reproduced the issue and validated the fix in the unit tests.

### DIFF
--- a/.changeset/stale-beds-trade.md
+++ b/.changeset/stale-beds-trade.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/esbuild-plugin-import-path-remapper": minor
+---
+
+Do not remap externals.
--- a/packages/esbuild-plugin-import-path-remapper/src/index.ts
+++ b/packages/esbuild-plugin-import-path-remapper/src/index.ts
@@ -5,7 +5,6 @@ const ESBUILD_DEFAULT_EXTENSIONS = [
   ".ts",
   ".jsx",
   ".js",
-  ".mjs",
   ".css",
   ".json",
 ];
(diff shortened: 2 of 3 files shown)


tido64 (approved):
Thanks, @steffenvv! Can you run `yarn change` and add a change log entry?

steffenvv:
Thanks @tido64, I pushed the change file, going for a minor version bump. As a side note, `yarn change` does not actually work out of the box after you've done `yarn` in this repo, because @changesets/cli is a devDependency in the root package.json. If you make it a dependency instead, the binary is actually linked up so that `yarn change` works. Not sure if that should be considered a bug in yarn or what, but I don't think there would be any harm in changing those deps in the root package.json from devDependency to dependency.

tido64:
> Thanks @tido64, I pushed the change file, going for a minor version bump. As a side note, `yarn change` does not actually work out of the box after you've done `yarn` in this repo, because @changesets/cli is a devDependency in the root package.json. If you make it a dependency instead, the binary is actually linked up so that `yarn change` works. Not sure if that should be considered a bug in yarn or what, but I don't think there would be any harm in changing those deps in the root package.json from devDependency to dependency.

Thanks! Like you said, there's no harm so I will move it up.

PR #1189, opened 2022-03-04
chore(incubator): create a more structured approach for experimental packages

kelset:
### Description

When I introduced the rn-macos patcher (https://github.com/microsoft/rnx-kit/pull/1135), we had a quick iteration on the problem of "how can we signal that a package is not something that we want everyone to use". To allow the package to go in, we agreed for a quick&dirty solution of pre-pending the name with `draft-` to make it explicit to the users that it's not a piece of software that we recommend anyone to use... unless they know exactly what they are doing.

We've had a few more conversations since, and this PR wants to introduce a more structured approach.

The centre piece of this new approach is a dedicated folder, separate from `packages` in which we can store these experimental code. The name we decided on is `incubator`, since the overall goal is that at some point the packages in here would get good enough that should graduate into proper packages and moved to the main folder.

Along with this, I've iterated a bit more on the part of the problem of making sure that devs using it would have it "in their face". I wasn't particularly fond of `draft-` because it's a bit too narrow of a describer, and it'd also make the command longer when using said packages with `npx`.

So here's a variation on that, trying to keep the spirit of warning alive: we add to each package in the incubator the
```
    "prepare": "echo '⚠️ name-of-package is EXPERIMENTAL - USE WITH CAUTION  ⚠️'",
```
warning. As far as we know, this should make it appear every time you use the `npx` command.
On top of that, the description in the `package.json` has also been modified to start with a similar phrase (and the README too) and an extra field `experimental: true` has been added (convenient for scripts).

Given all the above, I went ahead and renamed `draft-patch-rnmacos` to `patcher-rnmacos` to remove the draft bit and to make it clearer that it's not a patch, but a patcher tool.

Also, I've moved the `rn-changelog-generator` in here too since we don't want folks to think that this tool is something that they can use for generating changelogs of their RN projects (the RNR episode make me consider that misunderstanding might happen)

#### Housekeeping

* I've modified the `update-readme` script to generate a dedicated table for experimental packages hosted in the incubator.
* I've cleaned up and strengthened the new package generator logic and added the option to pass the extra flag `--experimental` to generate the package in the incubator with all the extra tweaks necessary.

### Test plan

Root level commands locally work as usual (ex. yarn, yarn lint, yarn build, etc etc).
Tested locally the update-readme and new package (both scenarios).
CI should verify that we are ✅


README.md
```diff
@@ -113,37 +113,38 @@ individually, as features are added and fixes are made.
 <!-- The following table can be updated by running `yarn update-readme` -->
 <!-- @rnx-kit start -->
 
-| Name                                                                                                                                                    | Description                                                                                      |
```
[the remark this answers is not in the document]
kelset: oh yeah 🤦‍♂️ I didn't notice they went up at the top - let's have a separate table 👍

incubator/patcher-rnmacos/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# @rnx-kit/draft-patch-rnmacos
```
tido64, line 1: Will you also deprecate this package on npm?

incubator/patcher-rnmacos/package.json
```diff
@@ -14,17 +15,18 @@
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "bin": {
-    "draft-patch-rnmacos": "lib/index.js"
+    "patcher-rnmacos": "lib/index.js"
   },
```
tido64, line 19: Should this just be:
```json
  "bin": "lib/index.js",
```
?

tido64 (commented):
It would also be great if the package initializer adds the necessary fields if a package is experimental. But it doesn't have to be in this PR.

incubator/patcher-rnmacos/package.json
kelset: I think both alternative are viable - if you prefer I can change it 👍

incubator/patcher-rnmacos/CHANGELOG.md
kelset: good question - I think the answer is that I would like to do so, but I don't think I have the auths for it on npm

kelset:
re @tido64:
> It would also be great if the package initializer adds the necessary fields if a package is experimental. But it doesn't have to be in this PR.

it already does 👍 even the extra `experimental` field.

packages/template/README.md
```diff
@@ -5,9 +5,11 @@
 [![Build](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml/badge.svg)](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml)
 [![npm version](https://img.shields.io/npm/v/@rnx-kit/template)](https://www.npmjs.com/package/@rnx-kit/template)
 
+<!-- experimental-warning start -->
+<!-- experimental-warning end -->
```
tido64, line 9: Do these get removed if `experimental === false`?

scripts/new-package.js
```diff
@@ -1,65 +1,90 @@
...
-  const fse = require("fs-extra");
-  const fs = require("fs");
-  const path = require("path");
-  const templatePath = path.join(__dirname, "../packages/template");
-  const newProjectPath = path.join(
-    __dirname,
-    "../packages/" + cleanProjectName
-  );
+var argv = require("yargs/yargs")(process.argv.slice(2))
+  .usage("Usage: $0 --name [string] --experimental")
```
tido64, line 13: Any chance we can use positional argument for the name? It would be nicer with `new-package MyAwesomePackage`.

scripts/new-package.js
```diff
@@ -1,65 +1,90 @@
...
+
+// copy the package/template folder to a new folder with the name of the value
+// and then change the package.json file to the new name
+const fse = require("fs-extra");
+const fs = require("fs");
+const path = require("path");
+const templatePath = path.join(__dirname, "../packages/template");
+const newProjectPath = path.join(
+  __dirname,
+  `../${experimental ? "incubator" : "packages"}/` + cleanProjectName
```
tido64, line 32: This ternary is repeat several time below. Can we extract it to a variable?

tido64 (approved):

PR #1211, opened 2022-03-10
feat(babel-preset): allow enabling loose mode on `@babel/plugin-transform-classes`

tido64:
### Description

Allow enabling loose mode on `@babel/plugin-transform-classes`. Depending on how heavy classes are used, it can add up to a significant number of bytes saved. If you're using TypeScript, this should be safe to enable.

### Test plan

Tested internally.


kelset (approved):
👍

PR #1225, opened 2022-03-14
fix(react-native-test-app-msal): throw if `msal_config.json` is missing

tido64:
### Description

Throw if `msal_config.json` is missing, otherwise Android will throw a cryptic/generic exception that's hard to debug.

### Test plan

n/a


kelset (approved):

sveinar (approved):
lgtm!

PR #1248, opened 2022-03-18
feat(lazy-index): move `react-native-lazy-index` into its new home

tido64:
### Description

Moves [`react-native-lazy-index`](https://github.com/microsoft/react-native-lazy-index) into the repo.

Post-move tasks: #1249

### Test plan

CI should pass.


packages/react-native-lazy-index/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "react-native-lazy-index",
```
tido64, line 2: @afoxman, @JasonVMo, @kelset: Should this package be renamed or do we want to keep this name to minimise disruptions? It doesn't look like it gets lots of downloads, and we can probably change the few places that we integrate it internally.

PR #1374, opened 2022-04-07
fix(dep-check): dependencies should be sorted lexicographically

tido64:
### Description

Dependencies should be sorted lexicographically.

### Test plan

Tests were added. CI should pass.

### DIFF
--- a/packages/dep-check/src/helpers.ts
+++ b/packages/dep-check/src/helpers.ts
@@ -4,9 +4,9 @@ import detectIndent from "detect-indent";
 import fs from "fs";
 
 export function compare<T>(lhs: T, rhs: T): -1 | 0 | 1 {
-  if (lhs == rhs) {
+  if (lhs === rhs) {
     return 0;
-  } else if (lhs == null || lhs < rhs) {
+  } else if (lhs < rhs) {
     return -1;
   } else {
     return 1;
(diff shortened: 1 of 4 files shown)


packages/dep-check/src/helpers.ts
```diff
@@ -3,6 +3,16 @@ import { writePackage } from "@rnx-kit/tools-node/package";
 import detectIndent from "detect-indent";
 import fs from "fs";
 
+export function compare<T>(lhs: T, rhs: T): -1 | 0 | 1 {
+  if (lhs == rhs) {
```
afoxman, line 7: Should this function use === for a strict comparison without coercion?

tido64: This is for comparing `null` vs `undefined`:
```
> null === undefined
false
> null == undefined
true
```
But I think we can drop handling the null/undefined case since we will always be comparing strings.

packages/dep-check/src/manifest.ts
```diff
@@ -61,7 +62,7 @@ export function updateDependencies(
     }
     return result;
   }, Object.entries(dependencies));
-  return Object.fromEntries(entries.sort(([a], [b]) => a.localeCompare(b)));
```
afoxman, line 64: The docs for localeCompare seem to show it does lexicographic sorting. Was this change to include handling for non-string values like null and undefined, or does localeCompare not do lexicographic sorting in some cases?

afoxman (approved):
You've tested this, so I'm ok with the change, but I have some questions...

packages/dep-check/src/manifest.ts
tido64: It seems like `localeCompare` does not do lexicographic sorting. I hit an issue in an internal repo where Rush disagreed with what dep-check produces.

PR #1377, opened 2022-04-08
ci: fix unexpected symbol in `build.yml`

tido64:
### Description

Fixes `The workflow is not valid. .github/workflows/build.yml (Line: 8, Col: 9): Unexpected symbol: '"microsoft/rnx-kit"'. Located at position 22 within expression: github.repository == "microsoft/rnx-kit"`

See https://github.com/microsoft/rnx-kit/actions/runs/2114153842

### Test plan

I wish.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
 jobs:
   build_and_publish:
     name: "Build and publish"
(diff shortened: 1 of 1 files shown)


tido64:
Bypassing policies since this is a CI only change, and it unblocks publishing.

PR #1378, opened 2022-04-08
fix(eslint-plugin): retry with `.ts` when module id contains `.js`

tido64:
### Description

Due to how ESM works, you have to add the file extension to your imports. It looks like you have to add `.js` even though you use TypeScript. This makes the `no-export-all` rule trip.

### Test plan

Input:

```js
// index.ts
export * from "./types.js";


// types.ts
export type SomeType = true;
```

Before:

```
/~/src/index.ts
  5:1  error  Prefer explicit exports over `export *` to avoid name clashes, and improve tree-shakeability  @rnx-kit/no-export-all

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
```

After:

```
/~/src/index.ts
  5:1  error  Prefer explicit exports over `export *` to avoid name clashes, and improve tree-shakeability  @rnx-kit/no-export-all

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

error Command failed with exit code 1.
```

### DIFF
--- a/packages/eslint-plugin/src/rules/no-export-all.js
+++ b/packages/eslint-plugin/src/rules/no-export-all.js
@@ -102,21 +102,35 @@ const resolveFrom =
       });
 
       return (fromDir, moduleId) => {
-        const m = resolve(fromDir, moduleId);
-        if (!m) {
-          throw new Error(
-            `Module not found: ${moduleId} (start path: ${fromDir})`
-          );
-        }
-        if (m.endsWith(".js")) {
-          // `.js` files don't contain type information. If we find a `.d.ts`
-          // next to it, we should use that instead.
-          const typedef = m.replace(/\.js$/, ".d.ts");
-          if (fs.existsSync(typedef)) {
-            return typedef;
+        try {
+          const m = resolve(fromDir, moduleId);
+          if (!m) {
+            throw new Error(
+              `Module not found: ${moduleId} (start path: ${fromDir})`
+            );
           }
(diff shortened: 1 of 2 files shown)


afoxman (approved):

dzearing (approved):

packages/eslint-plugin/src/rules/no-export-all.js
```diff
@@ -102,21 +102,38 @@ const resolveFrom =
...
+          const m = resolve(fromDir, moduleId);
+          if (!m) {
+            throw new Error(
+              `Module not found: ${moduleId} (start path: ${fromDir})`
+            );
+          }
+          if (m.endsWith(".js")) {
+            // `.js` files don't contain type information. If we find a `.d.ts`
+            // next to it, we should use that instead.
+            const typedef = m.replace(/\.js$/, ".d.ts");
```
dzearing, line 115: The `d.ts` may not exist either; I'm not sure at this stage. Example:

* Add file1.ts with exported interface Foo
* Add file2.ts with `export * from './file1.js'`

Lint rule runs; will the `file1.d.ts` exist at this point?

tido64: This is for the case where you have `src/file.ts` be transpiled to `lib/file.js` + `lib/file.d.ts`. We want to pick up the `.d.ts` file if it exists.

PR #1432, opened 2022-04-27
feat(cli): add `rnx-copy-assets` command

tido64:
### Description

Adds a command to copy assets that are not referenced from JS. Usually, Metro copies imported assets for you, but sometimes you need additional files if they are only accessed from native modules.

The aim of this command is to generalize/standardize on how to get native assets into an app bundle. While this currently works for our internal app, it is incomplete. As such, I'm avoiding advertising it in the README for now.

### Test plan

Currently using this in an internal app.

### DIFF
--- a/packages/cli/src/copy-assets.ts
+++ b/packages/cli/src/copy-assets.ts
@@ -235,12 +235,12 @@ export async function copyProjectAssets(options: Options): Promise<void> {
       continue;
     }
 
-    info(`Copying assets for "${packageName}"`);
-
     const assets = await getAssets(context);
-    await copyAssets(context, packageName, assets);
     if (options.bundleAar) {
       assembleAarBundle(context, packageName, assets);
+    } else {
+      info(`Copying assets for "${packageName}"`);
+      await copyAssets(context, packageName, assets);
     }
   }
 }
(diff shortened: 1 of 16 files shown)


tido64:
cc @sweggersen @tom-un

packages/cli/src/copy-assets.ts
```diff
@@ -52,17 +69,135 @@ function keysOf(record: Record<string, unknown> | undefined): string[] {
   return record ? Object.keys(record) : [];
 }
 
+function versionOf(pkgName: string): string {
+  const { version } = require(`${pkgName}/package.json`);
```
afoxman, line 73: After running into ES module errors with require(package.json) in other areas, would it be safer to instead use file system operations? readPackage from tools-node is a good alternative.

packages/cli/src/copy-assets.ts
```diff
@@ -52,17 +69,135 @@ function keysOf(record: Record<string, unknown> | undefined): string[] {
...
+  if (dependencies) {
+    for (const [dependencyName, aar] of Object.entries(dependencies)) {
+      const { output, destination } = getAndroidPaths(
+        context,
+        dependencyName,
+        aar
+      );
+      if (output) {
+        if (!fs.existsSync(output)) {
+          targets.push(`:${aar?.targetName}:assembleRelease`);
```
afoxman, line 170: What happens when aar?.targetName yields null? Won't that run a non-sense gradle command?

tido64: That's… a stray `?`. I don't know how it got there 😆

afoxman (approved):

sweggersen (approved):

PR #1455, opened 2022-05-03
Website publishing, root README update

afoxman:
### Description

This PR updates the root README file, removing most content in favor of a website link.

It also adds PR and Build loop tasks for building/publishing the website.


.github/workflows/build.yml
```diff
@@ -30,6 +30,9 @@ jobs:
         run: yarn build:ci
       - name: Bundle packages
         run: yarn bundle:ci
+      - name: Build website
+        run: yarn build
```
JasonVMo, line 34: Question of whether this should be run in parallel or as part of this action? This isn't something to address now but we should look at CI length and parallelize where appropriate to speed things up.

JasonVMo (approved):

PR #1457, opened 2022-05-03
Fix twitch link, set up google analytics

afoxman:
### Description

Fix twitch link, set up google analytics


afoxman:
I'm pushing this through to get the website fixed. Will address any comments you have in a subsequent PR.

PR #1473, opened 2022-05-08
Transition tree shaking out of experimental mode

### LINKED ISSUE #1278: Transition esbuild tree-shaking out of experimental mode
Transition from an experimental package to a normal, trusted package. If needed, use a "beta" tag and indicate that results may vary. Current concerns are that bundles generated using internal bundle-splitting tools don't show expected shrinkage but other packages do.

afoxman:
### Description

Remove the experimental moniker from tree shaking, making it a fully supported feature.

The "experimental" config property and command-line parameter are still both supported, but both trigger a console warning message indicating that they are deprecated.

Resolves #1278 

### Test plan

Manually verified with test-app. Update automated tests to attempt using deprecated config and command-line "experimental" names. Update other tests to use new name.

### DIFF
--- a/packages/metro-serializer-esbuild/README.md
+++ b/packages/metro-serializer-esbuild/README.md
@@ -3,12 +3,9 @@
 # @rnx-kit/metro-serializer-esbuild
 
 [![Build](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml/badge.svg)](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml)
-![This plugin is highly experimental](https://img.shields.io/badge/state-experimental-critical)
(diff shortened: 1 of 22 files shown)


tido64 (commented):
Do we want to transition this to beta instead? Otherwise, I think we will have to questions like why this isn't enabled by default, etc. The rest of the changes look fine to me.

packages/metro-serializer-esbuild/README.md
```diff
@@ -101,6 +98,7 @@ react-native bundle --entry-file index.js --platform ios --dev false ...
   you can save the esbuild specific Metro config to a separate file and only
   specify it when needed, e.g.:
   ```sh
+  ???  FIX THIS IN THE CLI, AND CHANGE TO rnx-bundle BEFORE COMMITTING  ???
```
kelset, line 101: leftover comment?

afoxman: ooooooh thank you for catching it. removed.

tido64 (approved):

Issue #1487, opened 2022-05-10
Update Metro types in response to breaking changes in Metro's `resolver` configuration property


afoxman:
## Background

@react-native-community/cli 8 moved from Metro 0.67 to 0.70. This brought in a breaking change from Metro. They changed the `context` parameter that is passed to the optional custom module resolver in Metro config (`resolver` property).

The `context` param (first argument) had previously been a read/write object with a `resolveRequest` property set to whatever custom resolver was in place (the function pointed to by `resolver`). When Metro made the call out to the custom resolver, it wouldn't change `resolveRequest`. 

Our code has custom resolvers, and they sometimes need to call back into Metro to get the "default" resolution behavior. Metro needs the `context`, so to avoid infinite recursion, we would unhook `resolveRequest` from context and then put it back later.

With the breaking change from Metro, two things happened -- (i) `context` became a frozen (read-only) object, and (ii) the recursion detection was changed, effectively doing what we did by replacing `resolveRequest` before invoking any custom resolver.

Our code can no longer unhook and replace `resolveRequest`. JS will fail with a TypeError. To be compatible with old and new versions of Metro, we need to detect if `resolveRequest` is pointing at our resolver or Metro's resolver, and adjust as needed by making a _copy_ of context rather than modifying it.

## Timeline

Metro made this breaking change on 1/31/2022: https://github.com/facebook/metro/commit/d81d8877c05637eac2b4ea946a9fa1e8ae869b06. It shipped in version 0.68.

@react-native-community/cli updated from Metro 0.67 to 0.70 on 5/4/2022: https://github.com/react-native-community/cli/commit/65c5d02cd82704a0c0e78841d65f7da98db19342. This updated shipped in v8.0.0-alpha.5.

@react-native-community/cli-plugin-metro broke when alpha.5 was made, and was fixed in alpha.6 on 5/10/2022: https://github.com/react-native-community/cli/pull/1605.

RN 0.68.2, released yesterday, is still on Metro 0.67.x / CLI 7.x. RN 0.69-stable is on Metro 0.70.x and CLI 8.x, which means **@rnx-kit customers going to RN 0.69 will break**.

RN 0.69 is currently in the RC phase of being released as of 5/10/2022. It will be releasing soon (days, not weeks).

## Proposed Fix

We have 2 packages which are impacted by these changes: @rnx-kit/metro-resolver-symlinks and @rnx-kit/metro-service. 

### @rnx-kit/metro-resolver-symlinks

[...]

PR #1488, opened 2022-05-11
fix(metro-resolver-symlinks): support Metro 0.68+

tido64:
### Description

See #1487 for more details.

### Test plan

*We still don't have a good way to test with symlinks…

### DIFF
--- a/packages/metro-resolver-symlinks/src/index.ts
+++ b/packages/metro-resolver-symlinks/src/index.ts
@@ -1,4 +1,5 @@
 import { normalizePath } from "@rnx-kit/tools-node";
+import type { CustomResolver } from "metro-resolver";
 import {
   getMetroResolver,
   remapReactNativeModule,
@@ -10,33 +11,40 @@ import { remapImportPath } from "./utils/remapImportPath";
 function makeResolver({
   remapModule = (_, moduleName, __) => moduleName,
 }: Options = {}): MetroResolver {
-  const resolve = getMetroResolver();
+  const metroResolver = getMetroResolver();
   const resolvers = [remapModule, remapReactNativeModule, resolveModulePath];
 
-  return (context, moduleName, platform) => {
+  const resolver: MetroResolver = (context, moduleName, platform) => {
     if (!platform) {
       throw new Error("No platform was specified");
     }
 
-    const backupResolveRequest = context.resolveRequest;
-    delete context.resolveRequest;
+    let resolve: CustomResolver = metroResolver;
+    const resolveRequest = context.resolveRequest;
+    if (resolveRequest === resolver) {
+      delete context.resolveRequest;
+    } else if (resolveRequest) {
+      resolve = resolveRequest;
+    }
 
-    const modifiedModuleName = resolvers.reduce(
-      (modifiedName, remap) => remap(context, modifiedName, platform),
-      moduleName
-    );
+    try {
+      const modifiedModuleName = resolvers.reduce(
+        (modifiedName, remap) => remap(context, modifiedName, platform),
+        moduleName
+      );
 
-    const resolution = resolve(
-      context,
-      normalizePath(modifiedModuleName),
-      platform
-    );
-
-    // Restoring `resolveRequest` must happen last
-    context.resolveRequest = backupResolveRequest;
-
-    return resolution;
+      return resolve(
+        context,
+        normalizePath(modifiedModuleName),
+        platform,
+        null
+      );
+    } finally {
+      // Restoring `resolveRequest` must happen last
+      context.resolveRequest = resolveRequest;
+    }
   };
+  return resolver;
 }
 
 makeResolver.remapImportPath = remapImportPath;
--- a/packages/metro-resolver-symlinks/src/index.ts
+++ b/packages/metro-resolver-symlinks/src/index.ts
@@ -12,23 +12,23 @@ function makeResolver({
   remapModule = (_, moduleName, __) => moduleName,
 }: Options = {}): MetroResolver {
   const metroResolver = getMetroResolver();
-  const resolvers = [remapModule, remapReactNativeModule, resolveModulePath];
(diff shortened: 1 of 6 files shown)


kelset (approved):

packages/metro-resolver-symlinks/src/index.ts
```diff
@@ -10,32 +11,64 @@ import { remapImportPath } from "./utils/remapImportPath";
...
+      (context.resolveRequest as MetroResolver) || metroResolver
     );
+  };
 
-    // Restoring `resolveRequest` must happen last
-    context.resolveRequest = backupResolveRequest;
+  const withCompatibilityLayer =
+    metroResolverVersion < 6800
+      ? withContextManipulation
+      : withResolveRequestFirst;
```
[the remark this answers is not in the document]
tido64: I did it this way to move the check outside the loop, but it's probably premature. I'll change it.

packages/metro-resolver-symlinks/src/resolver.ts
```diff
@@ -19,12 +20,22 @@ function resolveFrom(moduleName: string, startDir: string): string {
...
-    return require(metroResolverPath).resolve;
+    const resolve = require(metroResolverPath).resolve;
+
+    const { version } = readPackage(metroResolverPath);
+    const [major, minor, patch] = version
+      .split("-")[0]
+      .split(".")
+      .map((value) => parseInt(value, 10));
+
+    return [resolve, major * 10000 + minor * 100 + patch];
```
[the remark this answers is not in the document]
tido64: Reverted.

afoxman (changes requested):
Testing: The condition that causes this is the fix Metro made in 0.68 -- Object.freeze() combined with the new code to change `resolveRequest`. Take a look at [resolve.js](https://github.com/facebook/metro/commit/d81d8877c05637eac2b4ea946a9fa1e8ae869b06#diff-b03f1b511a2be7abd755b9c2561e47f513f84931466f2cc20a17a4238d70f12b) from the Metro PR.

I think a good approach is creating your own context, simulating what Metro does, and feed it in. Simulate 0.67 and 0.68 behavior. You can use this against the original code to make sure you see a failure, then apply the changes and see success, and you'll know things are working as expected.

You'll need to mock getMetroResolver() to make it work.

packages/metro-resolver-symlinks/src/index.ts
```diff
@@ -10,33 +11,40 @@ import { remapImportPath } from "./utils/remapImportPath";
...
-
-    // Restoring `resolveRequest` must happen last
-    context.resolveRequest = backupResolveRequest;
+    let resolve: CustomResolver = metroResolver;
+    const resolveRequest = context.resolveRequest;
+    if (resolveRequest === resolver) {
+      delete context.resolveRequest;
+    } else if (resolveRequest) {
+      resolve = resolveRequest;
+    }
```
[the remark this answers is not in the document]
tido64: Tbh, I'm not a great fan of this. I means that we will always create a new `context` copy for every module that needs resolving. If we're using latest Metro, that means two copies per module. That's quite a lot garbage that gets generated. I would suggest keeping the logic as is until we can drop the manipulation entirely.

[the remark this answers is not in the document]
afoxman: I can't argue JS perf since it's a mystery to me. :) I like the simplicity of the other solution, but they both work, so I'm ok with with either one.

afoxman (approved):

afoxman:
Can you add a test to simulate Metro 0.67 and Metro 0.68 contexts?

packages/metro-resolver-symlinks/src/symlinkResolver.ts
```diff
@@ -0,0 +1,52 @@
...
+        (modifiedName, remap) => remap(context, modifiedName, platform),
+        moduleName
+      );
+
+      return resolve(
+        context,
+        normalizePath(modifiedModuleName),
+        platform,
+        null
+      );
```
afoxman, line 41: This is the "old" API signature, which seems like it should be compatible. The new one ignores the 4th param. I think the first three params are compatible...

Nothing to change, but something to be aware of if we start seeing weirdness/problems as we integrate with newer and older versions of Metro.

afoxman, line 41: This is the "old" API signature, which seems like it should be compatible. The new one ignores the 4th param. I think the first three params are compatible...

Nothing to change, but something to be aware of if we start seeing weirdness/problems as we integrate with newer and older versions of Metro.

packages/metro-resolver-symlinks/test/index.test.ts
```diff
@@ -1,120 +1,52 @@
...
-  beforeAll(() => {
-    process.chdir(useFixture("remap-platforms"));
-  });
+describe("makeResolver", () => {
+  const currentWorkingDirectory = process.cwd();
 
-  afterAll(() => {
-    process.chdir(currentDir);
+  afterEach(() => {
+    process.chdir(currentWorkingDirectory);
```
afoxman, line 26: Can you assert that this succeeded, so we get failures in the unlikely scenario that something goes wrong on disk?

afoxman, line 26: Can you assert that this succeeded, so we get failures in the unlikely scenario that something goes wrong on disk?

afoxman (approved):

packages/metro-resolver-symlinks/test/index.test.ts
tido64: `process.chdir` throws if it fails, which should fail the whole test run.

packages/metro-resolver-symlinks/src/symlinkResolver.ts
tido64: Yup, the reason I had to add a fourth now is because of the types. We need to go and fix them.

PR #1489, opened 2022-05-11
fix(metro-service): use `@react-native-community/cli-plugin-metro` if installed

tido64:
### Description

See #1487 for more details.

### Test plan

TODO

### DIFF
--- a/.changeset/selfish-bikes-reflect.md
+++ b/.changeset/selfish-bikes-reflect.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-service": patch
+---
+
+Use `@react-native-community/cli-plugin-metro` if installed
--- a/packages/metro-service/package.json
+++ b/packages/metro-service/package.json
@@ -30,12 +30,18 @@
     "make-dir": "^3.1.0"
   },
   "peerDependencies": {
+    "@react-native-community/cli-plugin-metro": "^6.1.0",
     "metro": ">=0.66.1",
     "metro-config": ">=0.66.1",
     "metro-core": ">=0.66.1",
     "metro-react-native-babel-transformer": ">=0.66.1",
     "metro-resolver": ">=0.66.1",
     "metro-runtime": ">=0.66.1"
+  },
+  "peerDependenciesMeta": {
+    "@react-native-community/cli-plugin-metro": {
+      "optional": true
+    }
   },
   "devDependencies": {
     "@react-native-community/cli-types": "^6.0.0",
--- a/packages/metro-service/src/config.ts
+++ b/packages/metro-service/src/config.ts
@@ -2,7 +2,12 @@ import path from "path";
 import type { Config as CLIConfig } from "@react-native-community/cli-types";
 import type { Reporter } from "metro";
 import { loadConfig, ConfigT, InputConfigT } from "metro-config";
-import { resolve, Resolution } from "metro-resolver";
+import type {
+  CustomResolver,
+  Resolution,
+  ResolutionContext,
+} from "metro-resolver";
+import { resolve as metroResolver } from "metro-resolver";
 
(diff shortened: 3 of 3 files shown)


kelset (approved):

afoxman:
I think there may be more to this than just default config, so don't merge just yet. I will take a closer look soon.

afoxman:
Changing to cli-plugin-metro is something we should do, but it is going to be more complicated than what you've got here. First, we want to target 8.0.0 official release (which isn't out yet) because it will have Andrew's fix. Second, we can only do this when the app is using Metro 0.70 or later because that is what 8.0.0 of the CLI will target (or maybe it was Metro 0.71, I don't remember).

And after wiring all that complexity into the codebase, you still need to patch the code I copied from the CLI the same way Andrew fixed it in the CLI repo: https://github.com/react-native-community/cli/commit/65c5d02cd82704a0c0e78841d65f7da98db19342

I'd honestly avoid the whole mess and just patch and keep using the copied CLI code we have in place today. I had hoped to remove the code soon, but it would mean we'd be moving the minimum Metro/CLI/RN versions we support forward by a lot and we couldn't be used in apps on older versions. I think we're going to need to keep a minbar of 0.64 for awhile longer.

tido64:
> I'd honestly avoid the whole mess and just patch and keep using the copied CLI code we have in place today. I had hoped to remove the code soon, but it would mean we'd be moving the minimum Metro/CLI/RN versions we support forward by a lot and we couldn't be used in apps on older versions. I think we're going to need to keep a minbar of 0.64 for awhile longer.

Not sure I follow. We have this piece of code because `cli-metro-plugin` wasn't a thing. The way I see it, if `cli-metro-plugin` is installed, we should start using it instead of our custom logic (I think they are mostly the same, as far as I can tell). When users upgrade to 8.x, there is nothing for us to do. We are already using `cli-metro-plugin` with all the latest fixes. Since Metro is pulled in by the CLI, there shouldn't be a case where people are using Metro 0.67 without also being on CLI 8. Am I missing something?

afoxman:
> Not sure I follow. We have this piece of code because `cli-metro-plugin` wasn't a thing. The way I see it, if `cli-metro-plugin` is installed, we should start using it instead of our custom logic (I think they are mostly the same, as far as I can tell). When users upgrade to 8.x, there is nothing for us to do. We are already using `cli-metro-plugin` with all the latest fixes. Since Metro is pulled in by the CLI, there shouldn't be a case where people are using Metro 0.70+ without also being on CLI 8+. Am I missing something?

You aren't following because I'm not making sense. Is there an I-overthought-this emoji? You're right.

But you still need to patch the fallback code the way Andrew did. Can you do that in this PR, too?

tido64:
> But you still need to patch the fallback code the way Andrew did. Can you do that in this PR, too?

Do we still? The breaking changes is introduced in `metro` 0.70. `cli-plugin-metro` is installed with CLI 6.x series. Meaning that from CLI 6.x on, we are no longer using our own logic.

| CLI version | Included Metro version | Use `cli-plugin-metro` |
| :---------: | :--------------------: | :--------------------: |
|     5.x     |          0.65          |           No           |
|     6.x     |          0.66          |          Yes           |
|     7.x     |          0.67          |          Yes           |
|     8.x     |          0.70          |          Yes           |

afoxman:
> Do we still? The breaking changes is introduced in `metro` 0.70. `cli-plugin-metro` is installed with CLI 6.x series. Meaning that from CLI 6.x on, we are no longer using our own logic.

That seems risky since control over when we fall back depends on the presence of cli-plugin-metro. What if they rename it? What if the detection goes wrong, like we introduce a bug down the road?

Having the fallback not work with newer versions of Metro makes me nervous.

afoxman (approved):

PR #1500, opened 2022-05-12
Fix TS errors/failures when bundling with type validation on non-TS packages

afoxman:
### Description

When doing typescript validation during bundling, ignore non-TS projects rather than failing.

### Test plan

Manual validation of both failure and success using internal test-app.


tido64 (approved):
Verified in an internal app. Thanks for fixing this so quickly!

PR #1513, opened 2022-05-16
chore(dep-check): add script for generating new profiles

tido64:
### Description

Super crude script for creating new profiles off react-native versions.

### Test plan

```
cd packages/dep-check
yarn build --dependencies
node scripts/update-profile.mjs 0.69

# Script should throw if we're missing a profile to base off
node scripts/update-profile.mjs 0.60
```


packages/dep-check/scripts/update-profile.mjs
```diff
@@ -0,0 +1,305 @@
...
+  metroVersion,
+}) {
+  const nextVersionCoerced = semverCoerce(targetVersion);
+  const currentVersion = `${nextVersionCoerced.major}.${
+    nextVersionCoerced.minor - 1
+  }`;
+  const currentVersionVarName = `${nextVersionCoerced.major}_${
+    nextVersionCoerced.minor - 1
+  }`;
+  return `import type { Profile, Package } from "../types";
```
afoxman, line 77: Maybe a quick check here to make sure ./profile-<currentversion>.[jt]s exists? Could help you catch a bug when generating if you skip a version or if RN skips a version for some reason.

tido64: Thanks for pointing that out. The script should now throw and exit:
```
% node scripts/update-profile.mjs 0.60
Error: Could not find 'src/profiles/profile-0.59.ts'
    at generateFromTemplate (file:///~/rnx-kit/packages/dep-check/scripts/update-profile.mjs:82:11)
    at makeProfile (file:///~/rnx-kit/packages/dep-check/scripts/update-profile.mjs:234:10)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async main (file:///~/rnx-kit/packages/dep-check/scripts/update-profile.mjs:257:26)
```

packages/dep-check/scripts/update-profile.mjs
```diff
@@ -0,0 +1,320 @@
...
+ * @param {string=} targetVersion
+ */
+async function main(targetVersion = "") {
+  const allVersions = /** @type {import("../src/types").ProfileVersion[]} */ (
+    Object.keys(defaultProfiles).reverse()
+  );
+
+  const latestProfile = defaultProfiles[allVersions[0]];
+
+  if (targetVersion) {
```
afoxman, line 255: Maybe add a sanity check here, something like this (maybe even more relaxed so you only check major/minor):

```
const SemVer = require('semver/classes/semver');
const semverEq = require('semver/functions/eq');

const semverTarget = new SemVer(targetVersion);
if (allVersions.find(semverEq)) {
  throw Error("profile already exists");
}
```

packages/dep-check/scripts/update-profile.mjs
```diff
@@ -0,0 +1,320 @@
...
+    return undefined;
+  }
+
+  const getPackageVersion = (packageName, dependencies) =>
+    dependencies[packageName]
+      ? semverCoerce(dependencies[packageName]).version
+      : null;
+  const { dependencies, peerDependencies } = reactNativeInfo;
+
+  if (peerDependencies) {
```
afoxman, line 199: Is there a case where you'd want to proceed without getting the `react` version out of `peerDependencies` or somewhere else? It seems like you'd want confirmation that it was read from the incoming target RN package metadata.

packages/dep-check/scripts/update-profile.mjs
```diff
@@ -0,0 +1,320 @@
...
+    newProfile.reactVersion = getPackageVersion("react", peerDependencies);
+  }
+
+  if (dependencies) {
```
afoxman, line 203: Similar question here. It seems strange to allow the script to proceed without explicitly reading `@react-native-community/cli` version. Same with the Metro version (via `cli-plugin-metro`).

packages/dep-check/scripts/update-profile.mjs
```diff
@@ -0,0 +1,353 @@
...
+      defaultTag: getPackageVersion(packageName, await dependencies),
+      fullMetadata: true,
+    });
+    return packageInfo.dependencies;
+  }, Promise.resolve(dependencies));
+
+  return generateFromTemplate({
+    targetVersion,
+    reactVersion: getPackageVersion("react", peerDependencies),
+    hermesVersion: getPackageVersion("hermes-engine", dependencies),
```
kelset, line 243: I don't think this part is needed 'cause of Hermes not being published on npm anymore.

packages/dep-check/scripts/update-profile.mjs
```diff
@@ -0,0 +1,353 @@
...
+    metroVersion: getPackageVersion("metro", cliMetroPluginDependencies),
```
kelset, line 244: I like the idea of picking up the version of metro from the CLI instead that from npm directly 👍

kelset (approved):
I tested it locally, overall I like how it works 👍 (aside the hermes thing)

we should probably have a bit of documentation at least in the README of dep-check, that way we won't forget how it works.

packages/dep-check/scripts/update-profile.mjs
```diff
@@ -187,49 +201,38 @@ async function makeProfile(targetVersion, latestProfile) {
...
-      }
-    }
-  }
+  // Fetch `metro` version from `@react-native-community/cli-plugin-metro` > `@react-native-community/cli`
+  const cliMetroPluginDependencies = await [
+    "@react-native-community/cli",
+    "@react-native-community/cli-plugin-metro",
+  ].reduce(async (dependencies, packageName) => {
+    const packageInfo = await pacote.manifest(packageName, {
+      defaultTag: getPackageVersion(packageName, await dependencies),
```
afoxman, line 234: Why the await on dependencies? It looks like coming in, dependencies will be set to a value. And inside the loop, it it set to packageInfo.dependencies.

afoxman (approved):

packages/dep-check/scripts/update-profile.mjs
tido64: It's less confusing if you consider the type definition of `reduce()`, which in this case, would be something like: `Array<T>.reduce<R>(reducer: (result: R, next: T) => R, initialValue: R)`, where `T = string, R = Promise<Record<string, string>>`. We have to `await dependencies` because its type is `Promise<Record<string, string>>`.

afoxman: Bummer. I'm starting to see that once you go async/await, everything is effected (it has to be). 👍

PR #1543, opened 2022-05-23
chore: migrate to Nx

tido64:
### Description

[Nx](https://nx.dev/) has made improvements since I last checked. A separate `workspace.json` is no longer required, and the new [dynamic output](https://blog.nrwl.io/new-terminal-output-performance-improvements-in-nx-13-5-c407bb1c963a) is really nice. On CI, the output is clean, allowing GitHub Actions to highlight test results:

| Lage | Nx |
|:-:|:-:|
| image | image |

### Test plan

CI should pass. Any `yarn *` command that we use should work as before.


nx.json
```diff
@@ -0,0 +1,50 @@
...
+        "target": "prepare",
+        "projects": "dependencies"
+      }
+    ],
+    "package": [
+      {
+        "target": "package",
+        "projects": "dependencies"
+      }
+    ],
```
afoxman, line 38: These are new verbs for the repo. Do you use them? Should I be using them? Or are they part of npm publishing, and they happen via changesets?

package.json
```diff
@@ -12,24 +12,24 @@
...
   "scripts": {
-    "build": "lage build --grouped",
-    "build:ci": "lage build:ci --verbose --grouped",
-    "build-scope": "lage build --grouped --log-level warn --no-deps --scope",
-    "bundle": "lage bundle --grouped",
-    "bundle:ci": "lage bundle --verbose --grouped",
-    "bundle+esbuild": "lage bundle+esbuild --verbose --grouped",
+    "build": "nx run-many --target build --all",
+    "build:ci": "nx affected --target test --parallel 2",
+    "build:ci:all": "nx run-many --target test --parallel 2 --all",
```
afoxman, line 17: Why the parallel limiters here? (The affected command is going to be nice for PRs :)

afoxman (approved):
Let's get this shipped.

This seems to suggest we can use `affected` more in our PRs for linting and testing: https://nx.dev/ci/monorepo-ci-github-actions. Maybe that would me a more complex config to tell it how to detect changes...

Do you know if we can leverage their distributed cache? e.g. CI build runs and exports artifacts to <somwhere>. Devs sync and build, and Nx checks <somewhere> for cache results, downloading them instead of running a build.

package.json
tido64: Tests time out if we don't set a limit. Nx doesn't detect number of cores, it just uses whatever is set in `nx.json`. It's currently set to 8, but my machine can do 16. The VMs that run our jobs only have 2 cores. That's why they are limited to 2 here.

nx.json
tido64: These probably shouldn't be here. They were generated and I didn't know what to do with them. I'll remove them since we use Changesets for publishing.

tido64:
> Do you know if we can leverage their distributed cache? e.g. CI build runs and exports artifacts to . Devs sync and build, and Nx checks for cache results, downloading them instead of running a build.

That's one of the things I'd like to learn more about. It seems that to enable _their_ distributed cache solution, you would change the runner to `@nrwl/nx-cloud`: https://github.com/microsoft/rnx-kit/blob/bf2ce1afed035a0426bec517686d8ec2ed32d5d6/nx.json#L5

My hunch is that we can replace this with something that points to e.g., Azure Blob storage instead. On the surface, it looks trivial, but I haven't done any digging.

package.json
tido64: I implemented a simple runner that detects cores.

tido64:
> My hunch is that we can replace this with something that points to e.g., Azure Blob storage instead. On the surface, it looks trivial, but I haven't done any digging.

While looking into how to detect number of cores, I found this setting: https://github.com/nrwl/nx/blob/6793e14d4ecf1c6af0a468eb1a312358fa847820/packages/nx/src/tasks-runner/default-tasks-runner.ts#L23

It seems to suggest that we only need to implement the interface.

package.json
afoxman: awesome!

afoxman (approved):

PR #1592, opened 2022-05-31
chore: address Dependabot alerts

tido64:
### Description

- https://github.com/advisories/GHSA-93q8-gq69-wqmw
- https://github.com/advisories/GHSA-gff7-g5r8-mg8m

### Test plan

n/a

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -25,7 +25,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .yarn-offline-mirror
-          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-6
+          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-7
       - name: Install package dependencies
         run: yarn ci
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .yarn-offline-mirror
-          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-6
```
afoxman, line 28: Do these caches communicate across agents, or do they just happen to have the same name everywhere? Asked another way, does run 1 leave data behind in this dir, and then run 2 picks it up?  (like, the agent isn't clean/wiped between runs)

I'm wondering why we don't just use a random id for each run... And if the answer is because our agents carry state from run to run, I'm going to be sad.

afoxman (approved):

.github/workflows/pr.yml
afoxman: Looking at the YAML file, it seems like we only have one instance of yarn offline mirror + yarn ci per agent: `review`, `build`, `build-android`, `build-ios`, and `build-website`. Each of these should be running on a brand new, clean agent machine. 

But if we're having cache problems, maybe the agent machines aren't clean? Or I have the wrong picture in my head about how the YAML is interpreted and used.

tido64: Yes, that's how I interpret this as well. And I do believe that's how this works in general. One potential scenario I can think of is if a job is uploading its cache gets interrupted by another job just finishing and starting to upload its cache to the same key, that they somehow clash and corrupt it. In that case, keying on both OS and Node version should fix that. But I think I've seen scenarios where the cache was working fine, but then you queue a new build and suddenly it's corrupted. It's kinda hard to debug when you don't have direct access. I just know that I've no such incident since moving to Yarn 3 in react-native-test-app.

PR #1657, opened 2022-06-19
Add expo test app

afoxman:
### Description

This PR adds an Expo test app. This app was created using expo-cli, and has been integrated (partially) into the monorepo.

Why add an Expo app? The bare workflow is an interesting alternative to ejecting. It lets you keep current more easily with Expo updates, while still building and controlling the source code for iOS and Android.

Further, the Expo tools are very nice, and if we plan to interop with them, we need a test app to ensure we aren't breaking anything.

I have plans to turn this into a more interesting app in a future PR, but I wanted to commit the code-generated baseline app first with the absolute minimum set of modifications needed for working in our monorepo.


afoxman:
Why does this PR look familiar? I re-created it from the one you already reviewed... It turns out that I fouled up my fork pretty badly, and had to hard-reset main and origin/main to match upstream/main. So I no longer trusted git to do the right thing re: merging. To be safe, after the hard reset, I re-created this PR.

tido64 (approved):

PR #1658, opened 2022-06-19
Integrate expo app2

afoxman:


.github/actions/yarn/action.yml
```diff
@@ -0,0 +1,20 @@
...
+  steps:
+    - name: Generate cache key
+      id: cache-key-generator
+      run: echo "::set-output name=cache-key::$(node scripts/shasum.mjs yarn.lock)"
+      shell: bash
+    - name: Cache /.yarn-offline-mirror
+      uses: actions/cache@v3
+      with:
+        path: .yarn-offline-mirror
+        key: yarn-${{ runner.os }}-${{ steps.cache-key-generator.outputs.cache-key }}
```
tido64, line 15: I think we should be using the `cache` property of `actions/setup-node` instead: https://github.com/actions/setup-node#caching-global-packages-data

This caches the downloaded cache instead of `node_modules`, which may be more stable.

.github/workflows/pr.yml
```diff
@@ -80,89 +66,42 @@ jobs:
           yarn bundle+esbuild
         shell: bash
         working-directory: packages/test-app
-  build-android:
+  build-android-expo-app:
     name: "Build Android"
```
tido64, line 70: We should use a different name so we can distinguish them in the sidebar.
```suggestion
  build-android-expo-app:
    name: "Build Expo Android"
```

.github/workflows/pr.yml
```diff
@@ -80,89 +66,42 @@ jobs:
...
+    uses: ./.github/workflows/build-android.yml
+    with:
+      package: expo-app
+  build-android-test-app:
+    name: "Build Android"
+    uses: ./.github/workflows/build-android.yml
+    with:
+      package: test-app
+  build-ios-expo-app:
     name: "Build iOS"
```
tido64, line 80: We should use a different name so we can distinguish them in the sidebar.
```suggestion
    name: "Build Expo iOS"
```

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,132 @@
...
+}
+
+xcpretty=$(which xcpretty)
+if [[ -z "${xcpretty}" ]]; then
+  echo ""
+  echo "ERROR: xcpretty is required to run this script. Install it using 'gem install xcpretty'."
+  echo "       You may need to use 'sudo' to install it."
+  echo ""
+  exit 1
+fi
```
tido64, line 31: We should drop `xcpretty`. It hasn't been updated for some time and is known to swallow errors.

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,132 @@
...
+  usage
+fi
+
+package=$(basename $(pwd))
+
+echo "Building iOS project for package '${package}'"
+echo "  Build Actions .. ${build_actions[@]}"
+echo "  Device ......... ${device}"
+echo "  Workspace ...... ${workspace}"
+echo "  Scheme ......... ${scheme}"
```
tido64, line 90: `xcodebuild` should already output this information (if you drop `xcpretty`).

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,132 @@
...
+if [[ $device_info =~ $re ]]; then
+  device_id=${BASH_REMATCH[1]}
+else
+  echo ""
+  echo "ERROR: failed to extract ID for device '${device}' from '${device_info}'"
+  echo ""
+  exit 1
+fi
+
+echo "  Device ID ...... ${device_id}"
```
tido64, line 111: `xcodebuild` should already output this information (if you drop `xcpretty`).

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,132 @@
...
+
+cd ios
+export RCT_NO_LAUNCH_PACKAGER=1
+
+command=(xcodebuild -workspace "${workspace}.xcworkspace" -scheme "${scheme}" -destination "platform=iOS Simulator,id=${device_id}" CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO)
```
tido64, line 116: The `platform=iOS Simulator` part is no longer necessary:
```suggestion
command=(xcodebuild -workspace "${workspace}.xcworkspace" -scheme "${scheme}" -destination "id=${device_id}" CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO)
```

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,132 @@
...
+command+=(${build_actions[@]})
+echo ""
+echo "Running: ${command[@]} | xcpretty"
+echo ""
```
tido64, line 120: As mentioned above, we should just invoke the command. `xcodebuild` will output all the parameters if you drop `xcpretty`:

Screenshot 2022-06-21 at 11 35 06

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,132 @@
...
+result=$?
+
+if [[ $result -ne 0 ]]; then
+  echo ""
+  echo "ERROR: build failed with error code: $result"
+  echo ""
+  exit $result
+fi
+
+exit 0
```
tido64, line 132: We should drop these as well. You just need `set -eo pipefail` set somewhere at the top. For reference, this is RNTA's build script: https://github.com/microsoft/react-native-test-app/blob/trunk/scripts/xcodebuild.sh

scripts/shasum.mjs
```diff
@@ -0,0 +1,18 @@
...
+ * @param {string=} algorithm
+ * @returns {string}
+ */
+function hashFile(filename, algorithm = "sha256") {
+  const data = readFileSync(filename, { encoding: "utf-8" }).replace(/\r/g, "");
+  return createHash(algorithm).update(data).digest("hex");
+}
+
+const { [2]: filename } = process.argv;
+console.log(hashFile(filename));
```
kelset, line 18: do we need this console.log to stick around? or just a leftover?

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env bash
+
+device_default="iPhone 12"
```
kelset, line 3: headsup, in @tido64 I think he was using a iPhone 13

kelset: sidenote: personally I think we should go for whatever is the lowest of the iPhone 1* group - it doesn't really affect what we do since it's not visual regression testing, and the higher means the more resources on CI needed (I think... since it's a macos machine with simulator)

scripts/shasum.mjs
tido64: We output the SHA so that GH can capture the output and pass it as a var to the next job step.

.github/actions/yarn/action.yml
afoxman: cool!

.github/workflows/pr.yml
afoxman: The sidebar gets differentiated using the package name combined with this name:
Screen Shot 2022-06-21 at 8 38 09 PM

scripts/src/commands/build-ios.sh
afoxman: will do. you might want to update RNTA, too, since it uses xcpretty in its CI loops.

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,26 @@
...
+  workflow_call:
+    inputs:
+      package:
+        required: true
+        type: string
+
+jobs:
+  build-android:
+    name: ${{ inputs.package }}
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
afoxman: take a crack at this. ADO is much better at templates and file inclusion, but I think I can make it work with environment variables.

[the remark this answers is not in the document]
afoxman: turns out the only vars you can control are env vars, and those are filled in at runtime, after the GH action workflow is read/parsed. So you can't control the matrix using variables directly in the yml files. Though you can use a json file and parse it and that all sounds like a hot mess.

scripts/src/commands/build-ios.sh
afoxman: so i'll leave this for now?

.github/workflows/pr.yml
tido64: Neat! I didn't notice this before.

scripts/src/commands/build-ios.sh
tido64: Where did you see it? I'm pretty sure I removed it all:
```
~rnta % git grep xcpretty
~rnta %
```

scripts/src/commands/build-ios.sh
tido64: Yes, we can leave it for now. Though bumping it to 13 _may_ win us a year or two.

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,109 @@
...
+xcodebuild -workspace "${workspace}.xcworkspace" -scheme "${scheme}" -destination "id=${device_id}" CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO "${build_actions[@]}"
+result=$?
+if [[ $result -ne 0 ]]; then
+  echo ""
+  echo "ERROR: build failed with error code: $result"
+  echo ""
+  exit $result
+fi
+
+exit 0
```
tido64, line 109: I don't think we'll reach this point. `set -eo pipefail` means this script should fail if `xcodebuild` fails.

.github/workflows/build-ios.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: macos-11
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup the toolchain
+        uses: ./.github/actions/setup-toolchain
+        with:
+          node-version: 14
```
tido64, line 26: Should we use the default Node version?
```suggestion
```

.github/workflows/build-android.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup the toolchain
+        uses: ./.github/actions/setup-toolchain
+        with:
+          node-version: 14
```
tido64, line 20: Should we use default Node version?
```suggestion
```

.github/workflows/build.yml
```diff
@@ -6,26 +6,19 @@ jobs:
...
         with:
           fetch-depth: 0
-      - name: Cache /.yarn-offline-mirror
-        uses: actions/cache@v3
+      - name: Setup the toolchain
+        uses: ./.github/actions/setup-toolchain
         with:
-          path: .yarn-offline-mirror
-          key: ${{ hashFiles('yarn.lock') }}
+          node-version: 16
```
tido64, line 18: Same here. We should use default Node version.
```suggestion
```

.github/workflows/pr.yml
```diff
@@ -3,16 +3,16 @@ on: pull_request
...
-        with:
-          node-version: 14
       - name: Checkout
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Setup the toolchain
+        uses: ./.github/actions/setup-toolchain
+        with:
+          node-version: 14
```
tido64, line 15:
```suggestion
```

scripts/src/commands/build-ios.sh
afoxman: good point

.github/workflows/build-ios.yml
afoxman: I want to be explicit, for the same reasons using ubuntu-latest is not as safe as using ubuntu-20.02. Defaults change. But I'm open to moving forward to 16 or even 18 (which I don't think weve really tested). I like the idea of using the latest node for most tasks and using ALL supported versions in PR/Build matrix runs of JS builds.

afoxman: What do you think?

afoxman: oh wait.. you mean let the script we are calling (which we own) decide. I will fix this.

scripts/src/commands/build-ios.sh
afoxman: bumped!

scripts/src/commands/build-ios.sh
```diff
@@ -0,0 +1,101 @@
...
+  echo "ERROR: failed to extract ID for device '${device}' from '${device_info}'"
+  echo ""
+  exit 1
+fi
+
+cd ios
+export RCT_NO_LAUNCH_PACKAGER=1
+
+xcodebuild -workspace "${workspace}.xcworkspace" -scheme "${scheme}" -destination "id=${device_id}" CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO "${build_actions[@]}"
+exit 0
```
tido64, line 101: We can remove this line as well.
```suggestion
```

.gitignore
```diff
@@ -3,6 +3,7 @@
 *.log
 *.tgz
 *.xcworkspace/
+.ccache/
```
tido64, line 6: You should copy over the conf file from RNTA as well.

afoxman: got it - thanks!

afoxman:
I think this is ready to go now. I've verified that the caches are all unique and are working. We have pod-caching and compiler-caching and now yarn-caching through node-setup. Plus expo-app is now using some of the rnx-kit scripts and commands.

afoxman:
The changes on this PR won't show as fully passed because some of the names changed. The new-name checks are passing. Once approved, I'll submit this PR and update branch protections to use the new names.

tido64 (approved):

PR #1659, opened 2022-06-20
feat(build): introduce experimental rnx-build command

tido64:
### Description

Introduces a new command for building your app in the cloud.

### Test plan

The workflow file needs to be checked in before the command starts working. I recommend testing this from my fork:

```sh
git clone https://github.com/tido64/rnx-kit.git rnx-kit-tido64
cd rnx-kit-tido64
git checkout tido/experiments/rnx-build
yarn
cd incubator/build
yarn build
yarn rnx-build
```

### DIFF
--- a/.github/workflows/rnx-build.yml
+++ b/.github/workflows/rnx-build.yml
@@ -1,0 +1,79 @@
+name: rnx-build
+on:
+  workflow_dispatch:
+    inputs:
+      packageManager:
+        description: "Supported package managers are `npm`, `yarn`, `pnpm` (v6.10+)"
+        required: true
+        default: "yarn"
+      platform:
+        description: "Supported platforms are `android`, `ios`, `macos`, `windows`"
+        required: true
+      projectRoot:
+        description: "Root of the project"
+        required: true
+jobs:
+  build-android:
+    name: "Build Android"
+    if: ${{ github.event.inputs.platform == 'android' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up JDK
+        uses: actions/setup-java@v3.3.0
+        with:
+          distribution: temurin
+          java-version: 11
+      - name: Set up Node 16
+        uses: actions/setup-node@v3.3.0
+        with:
+          node-version: 16
+          cache: ${{ github.event.inputs.packageManager }}
+      - name: Install npm dependencies
+        run: ${{ github.event.inputs.packageManager }} install
+      - name: Build Android app
+        uses: gradle/gradle-build-action@v2.2.0
+        with:
+          gradle-version: wrapper
+          arguments: --no-daemon clean assembleDebug
+          build-root-directory: ${{ github.event.inputs.projectRoot }}/android
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: android-artifact
+          path: ${{ github.event.inputs.projectRoot }}/android/app/build/outputs/apk/debug/app-debug.apk
+  build-ios:
+    name: "Build iOS"
+    if: ${{ github.event.inputs.platform == 'ios' }}
+    runs-on: macos-11
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Node 16
+        uses: actions/setup-node@v3.3.0
+        with:
+          node-version: 16
+          cache: ${{ github.event.inputs.packageManager }}
+      - name: Install npm dependencies
+        run: ${{ github.event.inputs.packageManager }} install
+      - name: Install Pods
+        run: pod install --project-directory=ios
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Build iOS app
+        run: |
+          device_name='iPhone 13'
+          device=$(xcrun simctl list devices "${device_name}" available | grep "${device_name} (")
+          re='\(([-0-9A-Fa-f]+)\)'
+          [[ $device =~ $re ]] || exit 1
+          xcodebuild -workspace SampleCrossApp.xcworkspace -scheme ReactTestApp -destination "id=${BASH_REMATCH[1]}" -configuration Debug -derivedDataPath DerivedData CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO build
+        working-directory: ${{ github.event.inputs.projectRoot }}/ios
+      - name: Prepare build artifact
+        run: |
+          tar -cvf ios-artifact.tar -C DerivedData/Build/Products/Debug-iphonesimulator ReactTestApp.app
+        working-directory: ${{ github.event.inputs.projectRoot }}/ios
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ios-artifact
+          path: ${{ github.event.inputs.projectRoot }}/ios/ios-artifact.tar
--- a/.github/workflows/rnx-build.yml
+++ b/.github/workflows/rnx-build.yml
@@ -66,11 +66,13 @@ jobs:
           device=$(xcrun simctl list devices "${device_name}" available | grep "${device_name} (")
           re='\(([-0-9A-Fa-f]+)\)'
           [[ $device =~ $re ]] || exit 1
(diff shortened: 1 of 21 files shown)


.github/workflows/rnx-build.yml
```diff
@@ -0,0 +1,81 @@
...
+        required: true
+        default: "yarn"
+      platform:
+        description: "Supported platforms are `android`, `ios`, `macos`, `windows`"
+        required: true
+      projectRoot:
+        description: "Root of the project"
+        required: true
+jobs:
+  build-android:
```
afoxman, line 16: I, too, just submitted a PR which fools with workflow files in/around ios and android builds (mine was refactoring and moving commands into "scripts" which CI then calls). We should rationalize them and meet in the middle.

incubator/build/README.md
```diff
@@ -0,0 +1,66 @@
...
+[![Build](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml/badge.svg)](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml)
+[![npm version](https://img.shields.io/npm/v/@rnx-kit/build)](https://www.npmjs.com/package/@rnx-kit/build)
+
+🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧
+
+### This tool is EXPERIMENTAL - USE WITH CAUTION
+
+🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧
+
+An experimental tool for building your native apps in the cloud.
```
afoxman, line 12: This is really cool! Free, too!

incubator/build/src/archive.ts
```diff
@@ -0,0 +1,28 @@
+import * as fs from "node:fs";
+import * as path from "node:path";
+import yauzl from "yauzl";
```
afoxman, line 3: JS library for compression and archiving is going to be slow. Have you looked into natively available apps on the GH runners? I'm pretty sure tar is available on windows and mac and linux. I've used it in other repos to unpack archives.

incubator/build/src/async.ts
```diff
@@ -0,0 +1,20 @@
+export function idle(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+export async function withRetries<R>(
```
afoxman, line 5: Is this for Windows? :(

incubator/build/src/constants.ts
```diff
@@ -0,0 +1,9 @@
+import * as os from "node:os";
+import * as path from "node:path";
+
+export const MAX_ATTEMPTS = 8;
+export const MAX_DOWNLOAD_ATTEMPTS = 3;
+
+export const BUILD_ID = "rnx-build";
```
afoxman, line 7: Probably going to need a number here, or we'll be have to fail if they try this concurrently.

incubator/build/src/remotes/github.ts
```diff
@@ -0,0 +1,230 @@
...
+      runId
+    );
+    if (data.total_count === 0) {
+      throw new Error("No artifacts were uploaded");
+    }
+    return data.artifacts;
+  }, MAX_DOWNLOAD_ATTEMPTS);
+
+  const data = await withRetries(async () => {
+    const { data } = await octokit().rest.actions.downloadArtifact({
```
afoxman, line 59: Is there a variant which takes an output stream as input, and flushes directly to it?

incubator/build/src/remotes/github.ts
```diff
@@ -0,0 +1,230 @@
...
+  );
+  const filename = path.join(buildDir, artifacts[0].name + ".zip");
+
+  await fs.mkdir(buildDir, { recursive: true, mode: 0o755 });
+  await fs.writeFile(filename, Buffer.from(data));
+  return filename;
+}
+
+function getPersonalAccessToken(): string | undefined {
+  if (process.env.GITHUB_TOKEN) {
```
afoxman, line 82: Will this be set on a dev machine? Expo has a login and logout command. Is that something you plan on adding (persisted auth tokens)? Security is a concern there.

incubator/build/src/remotes/github.ts
```diff
@@ -0,0 +1,230 @@
...
+    run_id,
+  };
+}
+
+async function watchWorkflowRun(
+  runId: WorkflowRunId,
+  spinner: ora.Ora
+): Promise<void> {
+  spinner.start("Starting build");
+  while (runId) {
```
afoxman, line 115: TODO: work in a responsive interrupt to abort and cancel the build. And probably to also run in the background (if you don't already have this), and poll / download via the command-line. Builds could take a long time (> 5 mins).

afoxman (approved):
Most of my comments are about what could be -- this is an awesome and huge set of commands and scripts!

Is all of the GH code in the workflow file, and the "build" package runs on the user's machine? That would make the most sense, but not sure if you intended for some of "build" to run server side.

.github/workflows/rnx-build.yml
tido64: I want this to be standalone so it can be "installed" with this tool. I'd like for it to be a single file to drop in your repo.

incubator/build/src/archive.ts
tido64: Downloaded build artifacts are always compressed. This is for extracting and installing them on the local machine.

incubator/build/src/remotes/github.ts
tido64: This is already running in the background. We even have CTRL+C support. The only thing we need to add is sending cancelation to GH. I'll add that to the TODO list in the README.

incubator/build/src/remotes/github.ts
tido64: Yes, this is something we should add.

incubator/build/src/remotes/github.ts
tido64: I don't think so. We have the option to pass the buffer directly to `yauzl` for decompression, but I thought it would be more debuggable if we saved the file to disk. Should we skip this step?

tido64:
> Is all of the GH code in the workflow file, and the "build" package runs on the user's machine? That would make the most sense, but not sure if you intended for some of "build" to run server side.

I think the ideal case would be that you `npm add` this package, run a `rnx-build init` or similar that installs the correct workflow file in your repo (`.github/workflows/rnx-build.yml` for GH), then `rnx-build` on subsequent runs.

tido64:
I'm merging this now. I'll address comments in separate PRs.

incubator/build/src/remotes/github.ts
tido64: https://github.com/microsoft/rnx-kit/pull/1668

.github/workflows/rnx-build.yml
```diff
@@ -0,0 +1,81 @@
...
+    name: "Build Android"
+    if: ${{ github.event.inputs.platform == 'android' }}
+    runs-on: ubuntu-latest
```
kelset, line 19: I know I'm late on this (sorry 😓) - quick nit here (that can be dealt with as a follow up). I think that we want to avoid using `latest` CI images in favour of fixed versions. Sometimes they bump them and things might break... better to just be specific (kind of like we do for macos VMs)

tido64: Addressed in #1671

incubator/build/README.md
tido64: 1. Sounds good. We'll probably add support for Azure DevOps as well, and at the very least provide a template for others to contribute more services.
2. #1671 should address this

incubator/build/src/archive.ts
afoxman: Right. Extraction logic is math-heavy and hasn't performed well when run as JS code (I've tried in other projects). I think if you run the same commands natively you'll find it is MUCH faster. Something to try...

afoxman: I suspect "tar" comes with all modern versions of Windows, like robocopy does. Here's a link to a project where I'm using it from an ADO agent. GH Windows agents should have it too. Of course linux/mac agents will have it (and probably unzip).

https://github.com/microsoft/continuity/blob/8f65d165444b7f4cf3b37f2f239938b1d1094294/CMake/boost.cmake#L86

PR #1671, opened 2022-06-21
fix(build): prompt to add workflow file

tido64:
### Description

Prompt the user to add the workflow file.

### Test plan

```
yarn
rm .github/workflows/rnx-build.yml
cd incubator/build
yarn build
yarn rnx-build
```

Example output:
```
% yarn rnx-build
yarn run v1.22.19
$ node lib/index.js
A workflow file needs to be checked in first before you can start using rnx-build.

rnx-build will now copy the file to your repository. Please check it into your main branch.

Press any key to continue
```


incubator/build/src/remotes/github.ts
```diff
@@ -164,16 +165,38 @@ export function getRepositoryInfo(
...
 
-export function isSetUp(spinner: ora.Ora): boolean {
+export async function install(): Promise<number> {
   const workflowFile = path.join(
     getRepositoryRoot(),
     ".github",
     "workflows",
     WORKFLOW_ID
   );
   if (!fileExists(workflowFile)) {
```
tido64, line 175: We should probably also make sure that they don't deviate? Or is that okay? Do we want users to be able to customize this file?

tido64: We probably do want it to be customizable, especially if people use Dependabot or Renovate to bump dependencies.

afoxman (approved):
My approach to these incubator reviews is to skim slowly, rather than fully understanding the implications of every change, since things are in the very early stages. Basically, I'm trying to learn a bit, but will only point out things that look suspect or areas where you may want to consider additional things.

I figure we'll do a more thorough review when we move out of incubation.

PR #1676, opened 2022-06-21
fix(build): add macOS and Windows support

tido64:
### Description

Adds macOS and Windows support.

### Test plan

1. Build
   ```
   cd incubator/build
   yarn build
   ```
2. Build for macOS: `yarn rnx-build --platform macos --project-root ../../packages/test-app`
3. Build for Windows: `yarn rnx-build --platform wWindows --project-root ../../packages/test-app`

### DIFF
--- a/.github/workflows/rnx-build.yml
+++ b/.github/workflows/rnx-build.yml
@@ -2,6 +2,10 @@ name: rnx-build
 on:
   workflow_dispatch:
     inputs:
+      architecture:
+        description: "Supported architectures are `arm64`, `x64`"
+        required: true
+        default: "x64"
       packageManager:
         description: "Supported package managers are `npm`, `yarn`, `pnpm` (v6.10+)"
         required: true
@@ -79,3 +83,78 @@ jobs:
         with:
           name: ios-artifact
           path: ${{ github.event.inputs.projectRoot }}/ios/ios-artifact.tar
+  build-macos:
+    name: "Build macOS"
+    if: ${{ github.event.inputs.platform == 'macos' }}
+    runs-on: macos-11
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Node 16
+        uses: actions/setup-node@v3.3.0
+        with:
+          node-version: 16
+          cache: ${{ github.event.inputs.packageManager }}
+      - name: Install npm dependencies
+        run: ${{ github.event.inputs.packageManager }} install
+      - name: Install Pods
+        run: pod install --project-directory=macos
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Build macOS app
+        run: |
+          xcworkspace=$(find . -maxdepth 1 -name '*.xcworkspace' -type d | head -1)
+          xcodebuild -workspace ${xcworkspace} -scheme ReactTestApp -configuration Debug -derivedDataPath DerivedData CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO build
+        working-directory: ${{ github.event.inputs.projectRoot }}/macos
+      - name: Prepare build artifact
+        run: |
+          tar -cvf macos-artifact.tar -C DerivedData/Build/Products/Debug ReactTestApp.app
+          shasum --algorithm 512 macos-artifact.tar
+        working-directory: ${{ github.event.inputs.projectRoot }}/macos
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: macos-artifact
+          path: ${{ github.event.inputs.projectRoot }}/macos/macos-artifact.tar
+  build-windows:
+    name: "Build Windows"
+    if: ${{ github.event.inputs.platform == 'windows' }}
+    runs-on: windows-2022
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up MSBuild
+        uses: microsoft/setup-msbuild@v1.1
+      - name: Set up Node 16
+        uses: actions/setup-node@v3.3.0
+        with:
+          node-version: 16
+          cache: ${{ github.event.inputs.packageManager }}
+      - name: Install npm dependencies
+        run: ${{ github.event.inputs.packageManager }} install
+      - name: Install Windows test app
+        run: |
+          npx --package react-native-test-app -- install-windows-test-app --use-nuget
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Set OID for unsigned packages
+        run: |
+          # https://docs.microsoft.com/en-us/windows/msix/package/unsigned-package
+          (Get-Content node_modules\.generated\windows\ReactTestApp\Package.appxmanifest) -replace '"CN=ReactTestApp"', '"CN=ReactTestApp, OID.2.25.311729368913984317654407730594956997722=1"' | Out-File -Encoding utf8 node_modules\.generated\windows\ReactTestApp\Package.appxmanifest
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Install NuGet packages
+        run: |
+          nuget restore
+        working-directory: ${{ github.event.inputs.projectRoot }}/windows
+      - name: Build Windows app
+        run: |
+          MSBuild -maxCpuCount -property:Configuration=Debug -property:Platform=${{ github.event.inputs.architecture }} -property:AppxBundlePlatforms=${{ github.event.inputs.architecture }} -property:AppxBundle=Always -property:UapAppxPackageBuildMode=SideloadOnly -property:UseBundle=false -target:Build
+        working-directory: ${{ github.event.inputs.projectRoot }}/windows
+      - name: Prepare build artifact
+        run: |
+          cp ${{ github.event.inputs.architecture }}/Debug/ReactTestApp/AppxManifest.xml AppPackages/ReactTestApp/*
+        shell: bash
+        working-directory: ${{ github.event.inputs.projectRoot }}/windows
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: windows-artifact
+          path: ${{ github.event.inputs.projectRoot }}/windows/AppPackages/ReactTestApp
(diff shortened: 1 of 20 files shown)


.github/workflows/rnx-build.yml
```diff
@@ -79,3 +83,78 @@ jobs:
...
+          node-version: 16
+          cache: ${{ github.event.inputs.packageManager }}
+      - name: Install npm dependencies
+        run: ${{ github.event.inputs.packageManager }} install
+      - name: Install Pods
+        run: pod install --project-directory=macos
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Build macOS app
+        run: |
+          xcworkspace=$(find . -maxdepth 1 -name '*.xcworkspace' -type d | head -1)
```
afoxman, line 105: Should xcworkspace file and scheme be CI parameters? The local client can find/select them and pass the right combo in.

.github/workflows/rnx-build.yml
```diff
@@ -79,3 +83,78 @@ jobs:
...
+          xcodebuild -workspace ${xcworkspace} -scheme ReactTestApp -configuration Debug -derivedDataPath DerivedData CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO build
+        working-directory: ${{ github.event.inputs.projectRoot }}/macos
+      - name: Prepare build artifact
+        run: |
+          tar -cvf macos-artifact.tar -C DerivedData/Build/Products/Debug ReactTestApp.app
+          shasum --algorithm 512 macos-artifact.tar
```
afoxman, line 111: Did you intend for the output to go somewhere? Did you just want it in the logs?

.github/workflows/rnx-build.yml
```diff
@@ -79,3 +83,78 @@ jobs:
...
+        run: pod install --project-directory=macos
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Build macOS app
+        run: |
+          xcworkspace=$(find . -maxdepth 1 -name '*.xcworkspace' -type d | head -1)
+          xcodebuild -workspace ${xcworkspace} -scheme ReactTestApp -configuration Debug -derivedDataPath DerivedData CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO build
+        working-directory: ${{ github.event.inputs.projectRoot }}/macos
+      - name: Prepare build artifact
+        run: |
+          tar -cvf macos-artifact.tar -C DerivedData/Build/Products/Debug ReactTestApp.app
```
afoxman, line 110: ReactTestApp.app should probably be a variable too, so this can be expanded beyond RNTA or in case RNTA changes.

.github/workflows/rnx-build.yml
```diff
@@ -79,3 +83,78 @@ jobs:
...
+          # https://docs.microsoft.com/en-us/windows/msix/package/unsigned-package
+          (Get-Content node_modules\.generated\windows\ReactTestApp\Package.appxmanifest) -replace '"CN=ReactTestApp"', '"CN=ReactTestApp, OID.2.25.311729368913984317654407730594956997722=1"' | Out-File -Encoding utf8 node_modules\.generated\windows\ReactTestApp\Package.appxmanifest
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Install NuGet packages
+        run: |
+          nuget restore
+        working-directory: ${{ github.event.inputs.projectRoot }}/windows
+      - name: Build Windows app
+        run: |
+          MSBuild -maxCpuCount -property:Configuration=Debug -property:Platform=${{ github.event.inputs.architecture }} -property:AppxBundlePlatforms=${{ github.event.inputs.architecture }} -property:AppxBundle=Always -property:UapAppxPackageBuildMode=SideloadOnly -property:UseBundle=false -target:Build
```
afoxman, line 149: PowerShell, nuget, msbuild... this YML job is so foreign/odd. This has to be an area of change for RNW down the road.

incubator/build/src/archive.ts
```diff
@@ -1,28 +1,60 @@
...
-            .pipe(fs.createWriteStream(entryFilename))
-            .once("finish", () => resolve(entryFilename));
+          const absolutePath = path.join(destination, entry.fileName);
+
+          // Our workflows will always produce build artifacts that are either
+          // an APK (Android), an archive (iOS/macOS), or a folder (Windows).
+          output ||= absolutePath;
+
+          if (absolutePath.endsWith("/") || absolutePath.endsWith("\\")) {
+            fs.mkdir(absolutePath, mkdirOptions, readNextEntry);
```
afoxman, line 32: ensureDir instead?

incubator/build/src/platforms/ios.ts
```diff
@@ -19,19 +21,19 @@ type Devices = {
   devices: Record<string, SimDevice[]>;
 };
 
-const open = makeCommand("open");
 const xcrun = makeCommand("xcrun");
 
-async function bootSimulator({ udid }: SimDevice): Promise<void> {
-  ensure(await xcrun("simctl", "boot", udid));
+async function bootSimulator({ udid }: SimDevice): Promise<Error | null> {
+  const { stderr, status } = await xcrun("simctl", "boot", udid);
```
afoxman, line 27: Does `simctl` have a blocking command with a timeout you can use?

incubator/build/src/platforms/windows.ts
```diff
@@ -0,0 +1,127 @@
...
+    const startArgs = prelude.join(" ");
+    return (...args: string[]) =>
+      start(
+        "-FilePath",
+        "powershell",
+        "-ArgumentList",
+        `'${startArgs} "${args.join(" ")}"'`,
+        "-PassThru",
+        "-Verb",
+        "RunAs"
```
afoxman, line 37: The number of arguments to run a command 😳

incubator/build/src/platforms/windows.ts
```diff
@@ -0,0 +1,127 @@
...
+  } else {
+    const ps = makeCommand("powershell", withPowerShell);
+    return (...args: string[]) => ps(...prelude, `"${args.join(" ")}"`);
+  }
+}
+
+async function getPackageInfo(app: string): Promise<PackageInfo | null> {
+  const filename = path.join(app, "AppxManifest.xml");
+  const manifest = await fs.readFile(filename, { encoding: "utf-8" });
+  const m = manifest.match(/<Identity[^>]*Name="(.*?)"/s);
```
afoxman, line 48:
```suggestion
  const m = manifest.match(/<Identity[^>]+Name="(.*?)"/s);
```
At least one char between Identity and Name? Also, maybe an XML parser instead? I know... but its safer and less error prone.

afoxman: And there are other instances below where it would be handy. readAppxManifest -> json object (from xml reader)

incubator/build/src/platforms/windows.ts
```diff
@@ -0,0 +1,127 @@
...
+
+function makePowerShell(
+  cwd: string,
+  elevated = false
+): ReturnType<typeof makeCommand> {
+  const withPowerShell = { cwd, shell: "powershell" };
+  const prelude = [
+    "-NoLogo",
+    "-NoProfile",
+    "-NonInteractive",
```
afoxman, line 21: How does noninterative play with elevation prompts, or are they unrelated?

incubator/build/src/platforms/windows.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  const files = await fs.readdir(app);
+  const msixbundle = files.find((file) => file.endsWith(".msixbundle"));
+  if (!msixbundle) {
+    return new Error("Missing MSIX package from build artifact");
+  }
+
+  const pwsh = makePowerShell(app);
+  const pwshElevated = makePowerShell(app, true);
```
afoxman, line 78: Does this prompt the user? I'd expect it would...

incubator/build/workflows/github.yml
```diff
@@ -2,6 +2,10 @@ name: rnx-build
 on:
```
afoxman, line 2: Is this the same as .github/workflows/rnx-build.yml?

afoxman (approved):

.github/workflows/rnx-build.yml
tido64: No, they should be pulled from `react-native config`. Right now, these are equivalent to what the CLI does.

.github/workflows/rnx-build.yml
tido64: I want them in the logs. I use them to make sure we downloaded them correctly.

incubator/build/src/platforms/ios.ts
tido64: It kinda does block, but it doesn't block until a simulator is up and running.

incubator/build/src/platforms/windows.ts
tido64: Yeah, I don't like this at all. There has to be a better solution. To install the unsigned app bundle, you have to be elevated. My guess is with a signed bundle, you don't have to. Something to look into later.

incubator/build/workflows/github.yml
tido64: Yes, it is. It gets copied into the repo if it's missing.

incubator/build/src/archive.ts
tido64: `openReadStream` doesn't take an async callback. Not sure it will behave correctly if I change it. I'll test later.

.github/workflows/rnx-build.yml
tido64: Yeah, it's not pretty. `run-windows` does abstract most of it away, but I needed extra props for MSBuild to get the `.msixbundle`.

incubator/build/src/platforms/windows.ts
tido64: Yeah… I guess we'll have to do this. I'll fix it in a separate PR today.

.github/workflows/rnx-build.yml
tido64: Agreed. I'll add it to the list of TODOs.

incubator/build/workflows/github.yml
tido64: #1689

PR #1696, opened 2022-06-26
fix(build): better handling of iOS device selection

tido64:
### Description

Improved handling of iOS device selection. Builds for physical devices differ from simulator ones, meaning we should pick the right device when deploying the app. Note that this does not quite add the capability to deploy to physical devices just yet. We still need to figure out how to deal with developer certificates.

This change also fixes an issue with corrupted artifacts.

### Test plan

1. Build: `cd incubator/build && yarn build`
2. Build and deploy to simulator (even if you have a device plugged in): `yarn rnx-build --platform ios --project-root ../../packages/test-app`
3. Build and deploy to device (which will fail because of certificates): `yarn rnx-build --platform ios --device-type device --project-root ../../packages/test-app`

### DIFF
--- a/.github/workflows/rnx-build.yml
+++ b/.github/workflows/rnx-build.yml
@@ -6,6 +6,10 @@ on:
         description: "Supported architectures are `arm64`, `x64`"
         required: true
         default: "x64"
+      deviceType:
+        description: "Supported device types are `device`, `emulator`, `simulator`"
+        required: true
+        default: "simulator"
       packageManager:
         description: "Supported package managers are `npm`, `yarn`, `pnpm` (v6.10+)"
         required: true
@@ -50,7 +54,7 @@ jobs:
   build-ios:
     name: "Build iOS"
     if: ${{ github.event.inputs.platform == 'ios' }}
-    runs-on: macos-11
+    runs-on: macos-12
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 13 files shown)


.github/workflows/rnx-build.yml
```diff
@@ -62,21 +66,39 @@ jobs:
...
+        working-directory: ${{ github.event.inputs.projectRoot }}
+      - name: Disable Clang sanitizers
+        run: |
+          # We need to disable Clang sanitizers otherwise the app will crash on
+          # startup trying to load Clang sanitizer libraries that would only
+          # exist if Xcode was attached.
+          xcconfig=node_modules/.generated/ios/ReactTestApp/ReactTestApp.debug.xcconfig
+          sed -i '' 's/CLANG_ADDRESS_SANITIZER = YES/CLANG_ADDRESS_SANITIZER = NO/g' ${xcconfig}
+          sed -i '' 's/CLANG_UNDEFINED_BEHAVIOR_SANITIZER = YES/CLANG_UNDEFINED_BEHAVIOR_SANITIZER = NO/g' ${xcconfig}
+          sed -i '' 's/-fsanitize=bounds//g' ${xcconfig}
```
afoxman, line 80: Is there a way to expose these more formally from RNTA, even down the road as a longer-term TODO, so you can control them from the command line? Kind of like CODE_SIGNING_ALLOWED=NO? Or maybe through an environment variable you set during the CI step for `xcodebuild`?

tido64: Filed https://github.com/microsoft/react-native-test-app/issues/974

PR #1700, opened 2022-06-28
fix(build/ios): add ability to launch on device

tido64:
### Description

Added ability to launch on device.

### Test plan

1. Build for Simulator: `yarn rnx-build --platform ios --project-root ../../packages/test-app`
    - Example build: https://github.com/tido64/rnx-kit/runs/7094524653?check_suite_focus=true
2. Build for device: `yarn rnx-build --platform ios --device-type device --project-root ../../packages/test-app`
    - Example build: https://github.com/tido64/rnx-kit/runs/7094128933?check_suite_focus=true

For the device build, you might have to change the bundle identifier and development team to match your provisioning profile:

```diff
diff --git a/packages/test-app/app.json b/packages/test-app/app.json
index 2ca6bad8..7da48bdc 100644
--- a/packages/test-app/app.json
+++ b/packages/test-app/app.json
@@ -18,6 +18,10 @@
   "android": {
     "package": "com.msft.identity.client.sample.local"
   },
+  "ios": {
+    "bundleIdentifier": "<my unique app bundle id>",
+    "developmentTeam": "<my development team id>"
+  },
   "react-native-test-app-msal": {
     "clientId": "4b0db8c2-9f26-4417-8bde-3f0e3656f8e0",
     "msaScopes": ["user.read"],
```

### DIFF
--- a/.github/workflows/rnx-build.yml
+++ b/.github/workflows/rnx-build.yml
@@ -55,6 +55,10 @@ jobs:
     name: "Build iOS"
     if: ${{ github.event.inputs.platform == 'ios' }}
     runs-on: macos-12
+    env:
+      CERTIFICATE_FILE: build-certificate.p12
+      KEYCHAIN_FILE: app-signing.keychain-db
+      PROVISION_PATH: 'Library/MobileDevice/Provisioning Profiles/Provisioning_Profile.mobileprovision'
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -66,8 +70,7 @@ jobs:
       - name: Install npm dependencies
         run: ${{ github.event.inputs.packageManager }} install
       - name: Install Pods
-        run: |
-          pod install --project-directory=ios
+        run: pod install --project-directory=ios
         working-directory: ${{ github.event.inputs.projectRoot }}
       - name: Disable Clang sanitizers
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/rnx-build.yml
```diff
@@ -79,16 +82,44 @@ jobs:
...
+          KEYCHAIN_PASSWORD: ${{ secrets.KEYCHAIN_PASSWORD }}
+          P12_PASSWORD: ${{ secrets.P12_PASSWORD }}
+        run: |
+          # https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development
+          mkdir -p "$(dirname "${HOME}/${PROVISION_PATH}")"
+
+          echo -n "${BUILD_CERTIFICATE_BASE64}" | base64 --decode --output "${RUNNER_TEMP}/${CERTIFICATE_FILE}"
+          echo -n "${BUILD_PROVISION_PROFILE_BASE64}" | base64 --decode --output "${HOME}/${PROVISION_PATH}"
+
+          security create-keychain -p "${KEYCHAIN_PASSWORD}" "${RUNNER_TEMP}/${KEYCHAIN_FILE}"
```
afoxman, line 99: Paranoia - can you check the CI logs to make sure this is properly redacted?

.github/workflows/rnx-build.yml
```diff
@@ -79,16 +82,44 @@ jobs:
...
           else
             destination='generic/platform=iOS Simulator'
+            code_signing='CODE_SIGNING_ALLOWED=NO'
           fi
           xcworkspace=$(find . -maxdepth 1 -name '*.xcworkspace' -type d | head -1)
-          xcodebuild -workspace ${xcworkspace} -scheme ReactTestApp -destination "${destination}" -configuration Debug -derivedDataPath DerivedData CODE_SIGNING_ALLOWED=NO COMPILER_INDEX_STORE_ENABLE=NO build
+          xcodebuild -workspace ${xcworkspace} -scheme ReactTestApp -destination "${destination}" -configuration Debug -derivedDataPath DerivedData ${code_signing} COMPILER_INDEX_STORE_ENABLE=NO build
         working-directory: ${{ github.event.inputs.projectRoot }}/ios
+      - name: Remove Apple signing certificate and provisioning profile
+        if: ${{ always() && github.event.inputs.deviceType == 'device' }}
```
afoxman, line 119: After some reading, I now understand what this does. Can you add a comment explaining that the use of always() ensures it runs even if previous steps fail?

afoxman (approved):

.github/workflows/rnx-build.yml
tido64: Yup, I've checked the output here: https://github.com/tido64/rnx-kit/runs/7094128933?check_suite_focus=true

The steps here are documented in: https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development#add-a-step-to-your-workflow

.github/workflows/rnx-build.yml
tido64: This step is documented here: https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development#required-clean-up-on-self-hosted-runners

I'll add a comment as well.

afoxman: Sorry - should have been more specific. I did read the GH docs. I meant the use of "always()" and how it reads like "if true && something else" (which doesn't make sense) but is really "always do this when deviceType == 'device' even if a previous task failed".

PR #1701, opened 2022-06-28
fix(build/windows): replace W11 only solution with code signing

tido64:
### Description

Replace Windows 11 only solution with proper code signing

### Test plan

```
cd incubator/build
yarn build
yarn rnx-build --platform windows --project-root ../../packages/test-app
```


incubator/build/src/platforms/windows.ts
```diff
@@ -64,6 +64,16 @@ async function getPackageInfo(app: string): Promise<PackageInfo | null> {
   return { packageName, appId };
 }
 
+function normalizePath(p: string): string {
```
afoxman, line 67: Can you add a comment for this? It's hard to understand why it is doing what it does...

afoxman (approved):

PR #1705, opened 2022-06-30
ci(ios): simplify destination parameter

tido64:
### Description

I've just learned that you can build for a generic iOS simulator.

### Test plan

CI should pass.


tido64:
> ok, let's go for this route. Is there any docs that explain what "generic/platform=iOS Simulator" corresponds to? Maybe in the Apple/xcode docs?

Yes, if you do `man xcodebuild` in the terminal, and skip to **Destinations**, it will tell you that "generic" can be used to "build against a platform generically instead of a specific device".

PR #1706, opened 2022-06-30
fix(metro-serializer-esbuild): default to the new `hermes` target

### LINKED ISSUE #1670: metro-serializer-esbuild: target ES6 with individual features disabled
As of esbuild [v0.14.46](https://github.com/evanw/esbuild/releases/tag/v0.14.46), we can target ES6 and disable individual JS features that Hermes do not support.

tido64:
### Description

Default to the new `hermes` target introduced in esbuild [0.14.48](https://github.com/evanw/esbuild/releases/tag/v0.14.48).

Resolves #1670.

### Test plan

Existing tests should pass.

### DIFF
--- a/.changeset/cool-seahorses-allow.md
+++ b/.changeset/cool-seahorses-allow.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-serializer-esbuild": patch
+---
+
+Default to the new `hermes` target introduced in esbuild [0.14.48](https://github.com/evanw/esbuild/releases/tag/v0.14.48).
--- a/packages/esbuild-plugin-import-path-remapper/package.json
+++ b/packages/esbuild-plugin-import-path-remapper/package.json
@@ -22,7 +22,7 @@
   },
   "devDependencies": {
     "@rnx-kit/scripts": "*",
-    "esbuild": "^0.14.10",
+    "esbuild": "^0.14.48",
     "prettier": "^2.0.0"
   },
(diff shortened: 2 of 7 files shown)


kelset (approved):

ecreeth:
@tido64 I just found that in the esbuild compact table of features, the [ConstAndLet](https://github.com/evanw/esbuild/blob/master/internal/compat/js_table.go#L395) one is added in the `hermes` target. I don't know how they handle that, but i think is wrong.

tido64:
> @tido64 I just found that in the esbuild compact table of features, the [ConstAndLet](https://github.com/evanw/esbuild/blob/master/internal/compat/js_table.go#L395) one is added in the `hermes` target. I don't know how they handle that, but i think is wrong.

@ecreeth: No, I think this is "correct". Hermes does support `const`, kind of. It supports the keyword, but none of its features 😛  Not sure if it affects how esbuild transpiles stuff. Do you know? Did you file an issue?

ecreeth:
> @ecreeth: No, I think this is "correct". Hermes does support `const`, kind of. It supports the keyword, but none of its features 😛 Not sure if it affects how esbuild transpiles stuff. Do you know? Did you file an issue?

🤷‍♂️ I don't know how they do it, nor did I fill out a issue.

tido64:
Filed https://github.com/evanw/esbuild/issues/2365. Let's see what Evan says.

PR #1707, opened 2022-07-01
fix(rnx-build): ensure clean-up only happens once

tido64:
### Description

Ensure `cleanUp()` is only called once, and some types fixes.

### Test plan

n/a

### DIFF
--- a/incubator/build/src/build.ts
+++ b/incubator/build/src/build.ts
@@ -1,5 +1,6 @@
 import ora from "ora";
 import { extract } from "./archive";
+import { once } from "./async";
 import { deleteBranch, pushCurrentChanges } from "./git";
 import * as platforms from "./platforms";
 import type { BuildParams, Remote, RepositoryInfo } from "./types";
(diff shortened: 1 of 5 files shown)


kelset (approved):

incubator/build/src/build.ts
```diff
@@ -22,14 +23,14 @@ export async function startBuild(
   spinner.succeed(`Created build branch ${buildBranch}`);
 
   const context = { ...repoInfo, ref: buildBranch };
-  const cleanUp = async () => {
+  const cleanUp = once(async () => {
```
afoxman, line 26: The result of this function will be undefined, right? Does that mean once() will run more than once in this case?

Also, what happens when cleanup fails? Does this throw and once() won't capture anything, letting cleanup happen again? And is that ok, or will a partial cleanup leave you in a bad state?

tido64: The return of this function should be `Promise<void>` (since it's `async`). It should only run once. I've checked it under various circumstances.

Clean-up does three things: 1) it deletes the build branch locally, 2) it deletes the build branch at remote, and 3) it cancels the build if it isn't already completed. The local operation should not fail. The other two could fail, but would only leave a dangling branch and/or a running workflow. None of these should leave your checkout in a bad state.

PR #1717, opened 2022-07-04
feat(tools-workspaces): slimmer workspaces package

tido64:
### Description

| @rnx-kit/dep-check w/    | workspace-tools | @rnx-kit/tools-workspaces |
| ------------------------ | --------------: | ------------------------: |
| Bundle size (unminified) |       1 531 380 |                 1 063 232 |
| Find ~1000 packages      |         ~367 ms |                   ~120 ms |
| Package footprint        |          368 KB |                    100 KB |

### Test plan

CI should pass.


packages/tools-workspaces/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "dependencies": {
+    "fast-glob": "^3.2.7",
+    "find-up": "^5.0.0",
+    "read-yaml-file": "^2.1.0"
+  },
+  "devDependencies": {
+    "@rnx-kit/scripts": "*"
+  },
+  "engines": {
+    "node": ">= 14"
```
kelset, line 37: this is a sidenote more than anything - should we bump the monorepo to require 16? It's been LTS for a while now

kelset (approved):
left a few comments, overall LGTM

packages/tools-workspaces/package.json
tido64: We should already be running 14+16 for builds and tests, and 16 for everything else.

afoxman:
If size is a concern, npm says fast-glob is 88k. Wondering if you really need it for that one line of things to ignore, or if there is another way to get the same result using a regex (or a few regexes)?

afoxman:
Is this a subset / fork of workspace-tools, or is it all new? If it's from workspace-tools, it's had a lot of bake time. Can probably go with a closer-to-1.0 version.

Also, can you add an entry in the docsite for this? You need a stub readme file and a sidebar entry. Look at the other packages for reference.

tido64:
> If size is a concern, npm says fast-glob is 88k. Wondering if you really need it for that one line of things to ignore, or if there is another way to get the same result using a regex (or a few regexes)?

All the strings under `workspaces.packages` are glob patterns. I'd rather not reimplement it. [glob](https://bundlephobia.com/package/glob@8.0.3) is about 58k smaller and could be a good alternative if it's not slower. But overall, I'm pretty happy with shaving off ~0.5 MB.

> Is this a subset / fork of workspace-tools, or is it all new? If it's from workspace-tools, it's had a lot of bake time. Can probably go with a closer-to-1.0 version.

This is mostly new code/copy-pasta. I started out using the official libs, i.e. [@npmcli/map-workspaces](https://www.npmjs.com/package/@npmcli/map-workspaces) and [@pnpm/find-workspace-packages](https://www.npmjs.com/package/@pnpm/find-workspace-packages), but they do a lot more than they advertise so I ended up implementing it myself.

> Also, can you add an entry in the docsite for this? You need a stub readme file and a sidebar entry. Look at the other packages for reference.

Will do.

PR #1742, opened 2022-07-14
fix(build): implement plugin system for app distribution

tido64:
### Description

Implemented plugin system for app distribution.

### Test plan

The test plan is similar to #1739. I've updated the `README`, but let me know if something is unclear or missing.

```sh
yarn
cd incubator/build
yarn build --dependencies

# Local deployment should still work
yarn rnx-build --platform android ../../packages/test-app

# Firebase config should be picked up if targeting device
yarn rnx-build --platform android --device-type device ../../packages/test-app

# Local deployment to devices should still work
yarn rnx-build --platform android --deploy local-only --device-type device ../../packages/test-app
```


incubator/build-plugin-firebase/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "url": "https://github.com/microsoft/rnx-kit",
+    "directory": "incubator/build-plugin-firebase"
+  },
+  "scripts": {
+    "build": "rnx-kit-scripts build",
+    "format": "rnx-kit-scripts format",
+    "lint": "rnx-kit-scripts lint"
+  },
+  "peerDependencies": {
+    "@rnx-kit/build": "*"
```
afoxman, line 28: This should point to a specific version range of the package.

incubator/build/src/distribution.ts
```diff
@@ -1,37 +1,33 @@
...
-  if (deployment === "autodetect") {
-    switch (service) {
-      case "firebase": {
-        return import("./distribution/firebase");
-      }
-    }
+  [plugin, pluginConfig]: Partial<Plugin>,
+  projectRoot: string
+): Promise<DistributionPlugin> {
+  if (deployment === "auto" && typeof plugin === "string") {
```
afoxman, line 17: "auto" doesn't seem to be tied to any decision making logic. When the second plugin comes along, how will auto pick between them?

Maybe it's more of a choice between local and device?

I may be missing something obvious.

afoxman (approved):

incubator/build/src/distribution.ts
tido64: The "full" list of values would probably be `auto`, `local-only`, `remote-only`. `auto` here basically means try remote first, then fallback to local if no plugin is specified. You can only specify one distribution plugin.

tido64: I renamed the values. Hopefully this is less confusing:

- `auto` -> `remote-first`
- `local` -> `local-only`

Issue #1743, opened 2022-07-14
metro-serializer-esbuild: "Transforming generator functions to the configured target environment ("es5") is not supported yet"


jamonholmgren:
### What happened?

I'm currently trying out `@rnx-kit/metro-serializer-esbuild` on [Ignite](https://github.com/infinitered/ignite) and running into the following error:

```
Transforming generator functions to the configured target environment ("es5") is not supported yet
```

Current diff: https://github.com/infinitered/ignite/compare/rnx-kit-esbuild

I can't figure out where the target is configured for esbuild so I can change it to `esnext` or similar. Ideas?

### Affected Package

@rnx-kit/metro-serializer-esbuild

### Version

0.1.9

### Which platforms are you seeing this issue on?

- [X] Android
- [X] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
System:
    OS: macOS 12.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 17.55 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.19.1 - /opt/homebrew/opt/node@14/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 6.14.16 - /opt/homebrew/opt/node@14/bin/npm
    Watchman: 2022.03.21.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.4, iOS 15.5, macOS 12.3, tvOS 15.4, watchOS 8.5
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.2 AI-212.5712.43.2112.8512546
    Xcode: 13.4.1/13F100 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.14 - /Users/jh/.sdkman/candidates/java/current/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2
    react-native: 0.68.2 => 0.68.2
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```
```


### Steps to Reproduce

Clone down the repo:

git clone https://github.com/infinitered/ignite.git

Switch to the branch

git checkout rnx-kit-esbuild

cd into the boilerplate folder
yarn
yarn build-ios
[...]

PR #1745, opened 2022-07-15
docs(metro-serializer-esbuild): document options

tido64:
### Description

Documents currently supported options. See #1743 for context.

### Test plan

n/a


afoxman (approved):
How do you feel about adding a note near line 45 about Hermes and how _it works best_ using "es5"? And that esbuild doesn't have full support for transpiling to es5.

That's the crux of the confusion for users -- knowing that they are constrained to using es5 or that they have to take extra steps to get there. The new options you documented help with this, but babel may also be needed. Maybe list the generator function transform as an example.

tido64:
> How do you feel about adding a note near line 45 about Hermes and how _it works best_ using "es5"? And that esbuild doesn't have full support for transpiling to es5.

This is a good callout. I've added a note about Hermes and a ref to the issue so people can read the whole thread. Let me know if I'm missing anything.

Issue #1638, opened 2022-06-14
RFC: Proposal for new configuration scheme for dep-check


tido64:
## Goals

- To make it easier to relate to profiles, especially for web only consumers
- To make it easier to provide custom presets (collection of [profiles](https://microsoft.github.io/rnx-kit/docs/architecture/dependency-management#profiles))
  - This used to be hidden from the public, but will be exposed in this proposal. One example of a preset is the [built-in one](https://github.com/microsoft/rnx-kit/blob/main/packages/dep-check/src/presets/microsoft/index.ts).
- The old configuration scheme will still be supported and take precedence

## Motivation

While `dep-check` is a tech agnostic solution for managing dependencies, its configuration schema is currently very coupled with React Native versions. This makes the configuration look out of place when used in web only packages. Further, users have expressed a desire to centralize their configuration around arbitrary dependencies (e.g. `react` or `@fluentui/react` instead of `react-native`). They also want the ability to bring their own presets.

## Proposal

Taking inspiration from [browserslist](https://github.com/browserslist/browserslist), we want to introduce queries as a way to pick profiles.

For example, to ensure that all packages are aligned on React 18.x:

```json
{
  "dep-check": {
    "requirements": [
      "react@^18.0"
    ]
  }
}
```

To ensure all packages declare compatibility with React >=16, and align on 18.x for development:

```json
{
  "dep-check": {
    "requirements": {
      "development": [
        "react@^18.0"
      ],
      "production": [
        "react@>=16.13"
      ]
    }
  }
}
```

You can bring your own presets to enrich the built-in one:

```json
{
  "dep-check": {
    "presets": [
      "microsoft",
      "/path/to/my-collection-of-multiple-profiles"
    ],
    "requirements": [
      "react@^18.0"
    ]
  }
}
```

Or replace the built-in preset entirely:

```json
{
  "dep-check": {
    "presets": [
      "/path/to/my-collection-of-multiple-profiles"
    ],
    "requirements": [
      "react@^18.0"
    ]
  }
}
```

You can also add single profiles:

```json
{
  "dep-check": {
    "profiles": [
      "/path/to/additional-profile"
    ],
    "requirements": [
      "react@^18.0"
    ]
  }
}
```

Note that this config is just a different "frontend" to the tool. Everything else still works the same. For instance, the following configurations are equivalent:

```json
{
  "rnx-kit": {
    "reactNativeVersion": "^0.66 || ^0.67 || ^0.68",
[...]

Issue #484, opened 2021-08-13
RFC: rename `dep-check` to something else


kelset:
I'm opening this issue because I want to start a conversation on changing the name of one of the tools we provide, currently named `dep-check`.

I've been thinking about this for a while because of the name collision with a tool in the same space called [`depcheck`](https://github.com/depcheck/depcheck)... which kind of works in the same problem space too, so there could be situation in which both tools might be used in the same codebase.

This topic has also been brought up by @afoxman [here](https://github.com/microsoft/rnx-kit/pull/455#discussion_r683055206):
> The name dep-check is getting stretched awfully thin at this point... maybe a more general name since it can do so much to handle dependencies? Or several commands?

And I agree - `dep-check` is now a defacto incredibly powerful tool that:
* keeps dependencies versions aligned across packages and repos, based on curated profiles
* fixes package.json configurations (puts deps, devDeps, peerDeps correctly)
* addresses missing dependencies
* can do these things even for non RN packages
* (and I'm probably forgetting something)

So I would propose that we should come up with a better name for it.

Here are some quick ideas to use as baseline for coming up with the new name, and attached is the npmjs search for that term (to validate over time that the name wasn't taken/stolen since the moment I'm writing this).
* `tailor` - [npmjs search](https://www.npmjs.com/search?q=tailor)
* `deps-tailor` - [npmjs search](https://www.npmjs.com/search?q=deps-tailor)
* `align-deps` - [npmjs search](https://www.npmjs.com/search?q=align-deps)
* `deps-fixer` - [npmjs search](https://www.npmjs.com/search?q=deps-fixer)
* `shape-deps` - [npmjs search](https://www.npmjs.com/search?q=shape-deps)
* `deps-shaper` - [npmjs search](https://www.npmjs.com/search?q=deps-shaper)

The only minor point about this, is that I would wait to make this happen to >= mid September, since on Sept 1/2 there's RNEU where we announce and talk about it 😅

kelset:
Quick follow up, the name `dep-check` is also used for [this tool](https://opensourcelibs.com/lib/dep_check):
> Checks your (Flutter) dependencies for the latest version
🤣

afoxman:
dep-check is becoming important an important tool outside of react-native projects. I expect it will move out to another tools repo in the near future (months), and that will be the time to give it a new name. The react-native profiles will still live here and will be paired with the-artist-formerly-known-as-dep-check through our tools and config.

afoxman:
[...]

PR #1757, opened 2022-07-21
Proposal for dep-check 2.0 (new config scheme + rename)

kelset:
### Description

RFC to discuss further the new  new configuration scheme for `dep-check` originally proposed here https://github.com/microsoft/rnx-kit/issues/1638

You can read the [formatted text here](https://github.com/microsoft/rnx-kit/blob/kelset/dep-check-v2-rfc/text/0000-dep-check-v2.md), please leave comments inline to discuss.

### DIFF
--- a/text/0000-dep-check-v2.md
+++ b/text/0000-dep-check-v2.md
@@ -1,0 +1,225 @@
+- Title: dep-check v2
+- Date: 2022-07-21
+- RFC PR: (leave this empty)
+- rnx-kit Issue: (leave this empty)
+
+# Summary
+
+This proposal wants to cover a set of substantial changes we want to apply to
+`dep-check` to expand and reshape its core logic and make it easier to use. We
+have set a few goals for ourselves:
+
+- To make it easier to relate to profiles, especially for web only consumers
+- To make it easier to provide custom presets (collection of
+  [profiles](https://microsoft.github.io/rnx-kit/docs/architecture/dependency-management#profiles))
+  - This used to be hidden from the public, but will be exposed in this
+    proposal. One example of a preset is the
+    [built-in one](https://github.com/microsoft/rnx-kit/blob/main/packages/dep-check/src/presets/microsoft/index.ts).
+- The old configuration scheme will still be supported and take precedence
+
+# Motivation
+
+While `dep-check` is a tech agnostic solution for managing dependencies, its
+configuration schema is currently very coupled with React Native versions. This
+makes the configuration look out of place when used in web only packages.
+Further, users have expressed a desire to centralize their configuration around
+arbitrary dependencies (e.g. `react` or `@fluentui/react` instead of
+`react-native`). They also want the ability to bring their own presets.
+
+# Guide-level explanation
+
+This is how the new config will live in a `package.json`:
+
+```json
+{
+  "rnx-kit": {
+    "kitType": "app",
+    "bundle":{
+        ...
+    },
+    "dep-check": {
+        "presets": [
+          "microsoft",
+          "/path/to/my-collection-of-multiple-profiles"
+        ],
+        "requirements": [
+          "react@^18.0"
+        ],
+        "capabilities": [
+          "core-android",
+          "core-ios",
+          "core-macos",
+          "core-windows",
+          "react",
+          "test-app"
+        ],
+        "exclude-packages":{
+            "folder-name"
+        },
+      }
+  }
+}
+```
+
+There are a few main differences with how the config exists currently:
+
+- the config livs within the `dep-check` section of the `rnx-kit` section (not
+  on its own on top level)
+- there's no more `reactNativeVersion` and `reactNativeDevVersion`, replaced by
+  more generic `requirements` (see details in the next section)
+- no `profiles` option - it would just add confusion.
+
+Moreover, we will enforce both `presets` and `requirements` to not be empty.
+
+As a final note, the current "default" preset existing in the repo, `microsoft`
+will be renamed `microsoft-react-native` to open the door to have a more
+generic/web based one such as `microsoft-web`.
+
+# Reference-level explanation
+
+Taking inspiration from
+[browserslist](https://github.com/browserslist/browserslist), we want to
+introduce queries as a way to pick profiles.
+
+For example, to ensure that all packages are aligned on React 18.x:
+
+```json
+{
+  "dep-check": {
+    "requirements": ["react@^18.0"]
+  }
+}
+```
+
+To ensure all packages declare compatibility with React >=16, and align on 18.x
+for development:
+
+```json
+{
+  "dep-check": {
+    "requirements": {
+      "development": ["react@^18.0"],
+      "production": ["react@>=16.13"]
+    }
+  }
+}
+```
+
+You can bring your own presets to enrich the built-in one:
+
+```json
+{
+  "dep-check": {
+    "presets": ["microsoft", "/path/to/my-collection-of-multiple-profiles"],
+    "requirements": ["react@^18.0"]
+  }
+}
+```
+
+Or replace the built-in preset entirely:
+
+```json
+{
+  "dep-check": {
+    "presets": ["/path/to/my-collection-of-multiple-profiles"],
+    "requirements": ["react@^18.0"]
+  }
+}
+```
+
+In this new configuration shape, we are removing the `profiles` entry: it would
+just add confusion. Also, what we've been calling "custom profile" so far is
+actually a preset, so the custom preset is expected as an extra element in the
+presets array.
+
+Note that this config is just a different "frontend" to the tool. Everything
+else still works the same. For instance, the following configurations are
+equivalent:
+
+```json
+{
+  "rnx-kit": {
+    "reactNativeVersion": "^0.66 || ^0.67 || ^0.68",
+    "reactNativeDevVersion": "0.66"
+  }
+}
+```
+
+```json
+{
+  "dep-check": {
+    "requirements": {
+      "development": ["core@0.66"],
+      "production": ["core@^0.66 || ^0.67 || ^0.68"]
+    }
+  }
+}
+```
+
+## A Note on Conflict Resolution
+
+What happens in the case of conflicts? For instance, React Native is not
+compatible with React 18 until 0.69. What happens if we require both React 18
+and React Native below 0.69:
+
+```json
+{
+  "dep-check": {
+    "requirements": ["core@<0.69", "react@^18.0"]
+  }
+}
+```
+
+In this case, `dep-check` will throw an error since the built-in preset cannot
+fulfill the requirements. To resolve the conflict, we recommend that users move
+one of the requirements to the relevant packages. For instance:
+
+```js
+// global config
+{
+  "dep-check": {
+    "requirements": [
+      "react@^18.0"
+    ]
+  }
+}
+```
+
+```js
+// package config
+{
+  "rnx-kit": {
+    "reactNativeVersion": "^0.66 || ^0.67 || ^0.68",
+    "reactNativeDevVersion": "0.66",
+  }
+}
+```
+
+In short, `dep-check` will not try to resolve conflicts. It is up to users to
+build presets/profiles that fit their needs.
+
+# Drawbacks
+
+Not many drawbacks, just the one-off cost for developers to migrate to the new
+shape. But given the footprint of the change, it's a few lines per each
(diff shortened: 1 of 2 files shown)


text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,225 @@
...
+  }
+}
+```
+
+In this case, `dep-check` will throw an error since the built-in preset cannot
+fulfill the requirements. To resolve the conflict, we recommend that users move
+one of the requirements to the relevant packages. For instance:
+
+```js
+// global config
```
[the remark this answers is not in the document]
kelset: Good question!

For consistency sake, we are thinking of removing `--vigilant` since it's a bit indirect; in 2.0 we'd have the option to directly pass `presets` and `requirements`, so it would look something like this:
```
yarn rnx-dep-check --presets /path/to/my-collection-of-multiple-profiles --requirements react@^18.0
```

(and if you don't explicitly pass presets, we'd just use the default ones)

kelset:
(sidenote: we'll also want to do the renaming bit as part of this effort: https://github.com/microsoft/rnx-kit/issues/484)

text/0000-dep-check-v2.md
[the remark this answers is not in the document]
AlekhyaYalla: Makes sense to me. Thanks for update  & this amazing work!

kelset:
gentle nudge @afoxman @dzearing @JasonVMo - would be great to get your feedback in so that we can start making a plan of action once we have agreement on the way forward.

text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+To ensure all packages declare compatibility with React >=16, and align on 18.x
+for development:
+
+```json
+{
+  "align-deps": {
+    "requirements": {
+      "development": ["react@^18.0"],
+      "production": ["react@>=16.13"]
```
afoxman, line 113: "development" and "production" aren't discussed specifically in the RFC. Are they considered to be query terms of some kind, like you talked about with browserlist? Can you explain how that query system will work for align-deps, and how the terms are defined & used in code, data, and on the command-line?

text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,262 @@
...
+      "development": ["core@0.66"],
+      "production": ["core@^0.66 || ^0.67 || ^0.68"]
+    }
+  }
+}
+```
+
+## About commands
+
+For consistency sake, we will remove the existing `--vigilant` command since
```
afoxman, line 172: In 1.0, vigilant would check ALL packages, even those that aren't configured with an rnx-kit data block in package.json.

Are you saying that in 2.0, you get the "check ALL packages" behavior once you specify `presets` and `requirements`? And if you don't, then it will only check packages that specify them in config?

I wonder if I have that right, because it is at odds with the line below:

> if the user doesn't explicitlt pass presets, it will use the default ones

That sort of hints that an empty command line still applies align-deps behavior to ALL packages, configured or not.

text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+## A Note on Conflict Resolution
+
+What happens in the case of conflicts? For instance, React Native is not
+compatible with React 18 until 0.69. What happens if we require both React 18
+and React Native below 0.69:
+
+```json
+{
+  "align-deps": {
```
afoxman, line 193: Nit: this config example doesn't show the containing rnx-kit dictionary.

afoxman: Nor does the one below

text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,262 @@
...
+      "react@^18.0"
+    ]
+  }
+}
+```
+
+```js
+// package config
+{
+  "rnx-kit": {
```
afoxman, line 217: Can the packages in a repo be split, where some use old config and some use new?

What happens when a package accidentally has both? Does one take precedence? It might be helpful to detect this as a migration error, and complain that align-deps doesn't understand what to do.

text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,262 @@
...
+We'll pair the rollout with a migration guide and a blogpost in the rnx-kit
+repo.
+
+Also, since we will rename the package, we should also do the following steps:
+
+1. Rename the `dep-check` to the new name
+2. Change the old `dep-check` package to wrap the new package with a warning
+   about the name change
+3. Delete `dep-check` as its usage number goes to zero
+
```
afoxman, line 259: In 2.0, can the wrapped dep-check offer to migrate config programmatically? Is it doable with a simple data transformation, or does it need to be a manual change?

text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,262 @@
...
+    },
+    "align-deps": {
+        "presets": [
+          "microsoft",
+          "/path/to/my-collection-of-multiple-profiles"
+        ],
+        "requirements": [
+          "react@^18.0"
+        ],
+        "capabilities": [
```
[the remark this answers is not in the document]
tido64: The order is important. Subsequent presets may overwrite previous ones.

text/0000-dep-check-v2.md
```diff
@@ -0,0 +1,262 @@
...
+    "presets": ["/path/to/my-collection-of-multiple-profiles"],
+    "requirements": ["react@^18.0"]
+  }
+}
+```
+
+In this new configuration shape, we are removing the `profiles` entry: it would
+just add confusion. Also, what we've been calling "custom profile" so far is
+actually a preset, so the custom preset is expected as an extra element in the
+presets array.
```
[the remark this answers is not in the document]
tido64: dep-check will ship with a couple of built-in presets. When "microsoft" is specified in the presets array, dep-check will look for built-in presets first before falling back to Node module resolution. Built-ins are tightly controlled by us/community.

text/0000-dep-check-v2.md
tido64: `production` and `development` are roughly equal to today's `reactNativeVersion` and `reactNativeDevVersion`. The former determines what should be installed by consumers, while the other is for local development.

text/0000-dep-check-v2.md
tido64: These are roughly equivalent to the old flags:

| Old | New |
| :-- | :-- |
| `--vigilant` | `--requirements` |
| `--custom-profiles` | `--presets` |

text/0000-dep-check-v2.md
tido64: This should be possible since that's we would be doing internally anyway.

AlekhyaYalla:
I have a few questions on dep-check 2.0 features & timelines. I would like to get your confirmations on the below. Thanks!

- What would the timelines be for the dep-check 2.0 release?
- To support `peerDependencies` by defining with the "production" key and defining more than one profile/presets for a dependency,
Can we have the functionality of giving a range to a package with a "peer" key, so that if that dependency is there in `peerDependencies` section, that range would be used for alignment in the `peerDependencies` section
Eg: Today we need a similar configuration like this
  ```
  {
    "dep-check": {
      "requirements": {
        "development": [
          "react@18"
        ],
        "production": [
          "react@>=16.13.0  < 18.0.0"
        ]
      }
    }
  }
  ```
Proposal: Can we have similar to the below in profile itself
  ```
  {
  "react":{
        "name":"react",
        "version":"17.0.1",
        "peerRange": ">=16.8.1 <18.0.0
  }
  ```

- Currently with dep-check 1.0, if a profile is given in the dep-check command, that would be used for alignment for the dependencies defined in the profile. But if dependencies like react/react-dom/react-native are missing in the profile, the global(Github) 0.64 profile is getting used for alignment. I think this would be changed in dep-check 2.0, as we provide which presets has to be used for alignment, isn't it?

- In dep-check 1.0 we cant mention more than one profile, but with dep-check 2.0 we can align with more than one profile with "requirements" as an array, isn't it?

- With dep-check 2.0, we should be able to call/define a profile with any string eg: "default-profile", right? Cause in RFC proposal I see every requirement is based out of a dependency with a version eg: `react@^18.0 `/ `core@0.66`. Can I define a  profile irrespective of a dependency, declare the dependencies of my interest to be aligned in a single profile and use it for alignment on all of the packages?

tido64:
> 1. What would the timelines be for the dep-check 2.0 release?

Depending on scheduling and any unforeseen circumstances, I don't expect the implementation to take more than a few weeks.

> 2. To support `peerDependencies` by defining with the "production" key and defining more than one profile/presets for a dependency,
>    Can we have the functionality of giving a range to a package with a "peer" key, so that if that dependency is there in `peerDependencies` section, that range would be used for alignment in the `peerDependencies` section
>    Eg: Today we need a similar configuration like this
> 
> ```
> {
>   "dep-check": {
>     "requirements": {
>       "development": [
>         "react@18"
>       ],
>       "production": [
>         "react@>=16.13.0  < 18.0.0"
>       ]
>     }
>   }
> }
> ```
> 
> Proposal: Can we have similar to the below in profile itself
> 
> ```
> {
> "react":{
>       "name":"react",
>       "version":"17.0.1",
>       "peerRange": ">=16.8.1 <18.0.0
> }
> ```

This seems at odds with the consistency checks that dep-check provides. You should be able to get the same effect by applying the following preset:

```
{
  "react-library-production": {
    "name": "react",
    "version": ">=16.8.1 <18.0.0"
  }
}
```

I don't think we want to increase complexity by explicitly supporting a `peerRange` property.

> 3. Currently with dep-check 1.0, if a profile is given in the dep-check command, that would be used for alignment for the dependencies defined in the profile. But if dependencies like react/react-dom/react-native are missing in the profile, the global(Github) 0.64 profile is getting used for alignment. I think this would be changed in dep-check 2.0, as we provide which presets has to be used for alignment, isn't it?

With 2.0, you can have either behaviour, depending on your configuration. If you don't set `presets`, you'll be using one of the built-in presets. This is similar to 1.0 without the use of custom profiles. `presets: ["microsoft", "my-custom-preset"]` is roughly equivalent to 1.0 with custom profiles. New with 2.0 is that you can provide multiple presets and/or drop the built-in preset entirely.

> 4. In dep-check 1.0 we cant mention more than one profile, but with dep-check 2.0 we can align with more than one profile with "requirements" as an array, isn't it?

`requirements` is what decides which profiles gets applied. It follows the same rules as `reactNativeVersion` and `reactNativeDevVersion` of 1.0 today. These two props currently assume you want react-native. `requirements` is the generic version of them.

> 5. With dep-check 2.0, we should be able to call/define a profile with any string eg: "default-profile", right? Cause in RFC proposal I see every requirement is based out of a dependency with a version eg: `react@^18.0 `/ `core@0.66`. Can I define a  profile irrespective of a dependency, declare the dependencies of my interest to be aligned in a single profile and use it for alignment on all of the packages?

Profile names no longer matter in 2.0. Not unless you are looking to overwrite profiles.

tido64:
@afoxman, @AlekhyaYalla, @dzearing, @JasonVMo: We've incorporated all your feedback and made some sections clearer. Please have a look and sign off if you have no further comments/questions.

afoxman (approved):
Thank you for all the hard work and responsiveness on this RFC. This RFC is approved.

Next step is to enter a 7-day comment period. Since the community has already been commenting for several weeks, I think approval from the ppl CC'd in the last comment is enough to constitute final acceptance.

After that, please create and bind a new "implementation" issue to it. That issue will be tagged in any PRs where code is submitted to the repo. When all PRs are done and the work is finished, you can close the issue and consider the RFC complete.

text/0000-dep-check-v2.md
```diff
@@ -1,7 +1,7 @@
 - Title: dep-check v2
 - Date: 2022-07-21
-- RFC PR: (leave this empty)
-- rnx-kit Issue: (leave this empty)
+- RFC PR: https://github.com/microsoft/rnx-kit/pull/1757
```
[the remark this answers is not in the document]
afoxman: It's been 7 days since approval, and I don't see followup comments from the members who initially raised concerns, which I'll take to mean tacit acceptance.

This RFC is now accepted. I'll create a tracking workitem soon and merge this PR.

afoxman:
Congratulations @tido64 and @kelset on having your RFC accepted and approved!

kelset:
the issue tracking existing work is here: https://github.com/microsoft/rnx-kit/issues/1890

PR #1758, opened 2022-07-21
Add folders to the gitignore for rfcs base branch

kelset:
### Description

Small PR to address the fact that the `.gitignore` in the `rfcs` branch should basically ignore everything that exists in the normal `main` branch - that way switching back and forth won't make `git` die because of the leftover files.

### Test plan

N/A


.gitignore
```diff
@@ -1 +1,5 @@
 .DS_Store
+node_modules
+docsite
+incubator
+packages
```
tido64, line 5: Can you add a newline here? That should get rid of the warning:
image

tido64 (approved):

PR #1793, opened 2022-07-29
chore: bump Node requirement so we can target ES2019

tido64:
### Description

According to https://kangax.github.io/compat-table/es2016plus/, we can target ES2019 with Node 12+.

### Test plan

CI should pass.

### DIFF
--- a/packages/metro-service/package.json
+++ b/packages/metro-service/package.json
@@ -21,7 +21,7 @@
     "test": "rnx-kit-scripts test"
   },
   "engines": {
-    "node": ">=10.12"
+    "node": ">=12.13"
   },
   "dependencies": {
     "@rnx-kit/tools-language": "^1.2.6",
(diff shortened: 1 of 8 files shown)


packages/metro-service/package.json
```diff
@@ -21,7 +21,7 @@
     "test": "rnx-kit-scripts test"
   },
   "engines": {
-    "node": ">=10.12"
+    "node": ">=12.13"
```
tido64, line 24: @afoxman: Can we bump this to 14?

kelset: I think we should

kelset (approved):

packages/metro-service/package.json
afoxman: Yes, definitely. It's probably time to start looking at node 18 for our CI loops, and talk about an end-date for node 14 support. later this year? early next year?

The recent flurry of issues we've had from supporting older node engines is something I'd like to avoid in the future, if possible.

afoxman (approved):

PR #1796, opened 2022-08-02
feat: @expo/config-plugins support module for react-native-test-app

tido64:
### Description

Adds `@expo/config-plugins` support module for `react-native-test-app`.

### Test plan

See https://github.com/microsoft/react-native-test-app/pull/1033.


tido64:
Moved to the code into RNTA.

PR #1797, opened 2022-08-05
Update docusaurus

afoxman:
### Description

Update to docusaurus to release 2.0. Fix MDX inline plugin to fit with updated docusaurus mdx loader.

### Test plan

Verify website renders. Review "tools" pages manually, since they use the MDX-include loader that was updated in this PR.


afoxman:
Turns out this is broken because the 2.0 release requires Node 16. We still support Node 14 and our CI runs Node 14 builds.

I could use --ignore-engines with yarn, and have CI skip docusaurus unless its node 16 or higher, but that would suppress other packages too. 

Let's leave this PR open for now and talk about what the plan is for Node 14. Node 16 is LTS now and Node 18 is out.

tido64:
> Turns out this is broken because the 2.0 release requires Node 16. We still support Node 14 and our CI runs Node 14 builds.
> 
> I could use --ignore-engines with yarn, and have CI skip docusaurus unless its node 16 or higher, but that would suppress other packages too.
> 
> Let's leave this PR open for now and talk about what the plan is for Node 14. Node 16 is LTS now and Node 18 is out.

Does `docsite` need to be part of the monorepo? Otherwise, we could also just "eject" it and make sure we always use Node 16+.

afoxman:
@kelset This is ready for review. I've ejected docusaurus from the monorepo workspace, making it a standalone project under the /docsite directory. That means no more renovate or changeset support, and I'm ok with that for now. Down the road when we drop node 14, we can re-inject /docsite into the repo again. Or come up with another plan.

tido64 (approved):
> That means no more renovate or changeset support, and I'm ok with that for now.

Not sure Changeset support is necessary for the website, but I think Renovate should be smart enough to pick up the second `yarn.lock`. We'll find out once this merges.

kelset (approved):

PR #1836, opened 2022-08-23
fix(metro-service): replace `cp-file`, `make-dir` with Node API

tido64:
### Description

Neither [`cp-file`](https://github.com/sindresorhus/cp-file/blob/a98bdff6d7dd7005370e05f84537c58d1caf1fd5/fs.js#L73) nor [`make-dir`](https://github.com/sindresorhus/make-dir/blob/55af5b8ca0808f83607fc04350299b531f4cdd0e/index.js#L113) are doing anything extra besides calling Node directly.

### Test plan

CI should pass.

### DIFF
--- a/.changeset/perfect-kangaroos-juggle.md
+++ b/.changeset/perfect-kangaroos-juggle.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-service": patch
+---
+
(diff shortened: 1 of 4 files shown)


kelset (approved):
me likes removing deps in favour of standard 👏

PR #1848, opened 2022-08-25
feat(jest-preset): use `@swc/jest` if it's available

tido64:
### Description

⚠️⚠️⚠️ WIP ⚠️⚠️⚠️

I'm just checking if `@swc/jest` makes a significant improvement in build times.

### Test plan

CI should pass.


tido64:
The answer is no. At least not in our repo.

PR #1850, opened 2022-08-25
feat(cli): add support for "Random Access Module" bundle

tido64:
### Description

Add support for "Random Access Module" bundle format.

### Test plan

Verify that `rnx-bundle` still works:

```
% yarn bundle --dev false

% file dist/main.android.bundle
dist/main.android.bundle: ASCII text, with very long lines (37843)

% file dist/main.ios.jsbundle
dist/main.ios.jsbundle: ASCII text, with very long lines (37843)

% file dist/main.windows.bundle
dist/main.windows.bundle: ASCII text, with very long lines (36299)
```

Run `rnx-ram-bundle`:

```
% yarn ram-bundle --dev false

% file dist/main.android.bundle
dist/main.android.bundle: ASCII text, with very long lines (5559)

% file dist/main.ios.jsbundle
dist/main.ios.jsbundle: data

% file dist/main.windows.jsbundle
dist/main.windows.bundle: data
```

The file format for the Android bundle is still text because it's in the [File RAM bundle format](https://facebook.github.io/metro/docs/bundling#file-ram-bundle). Note that the `dist` folder now also contains a bunch of `js-modules/${id}.js` files.

Verify that you can force the output bundle format for Android:

```
% yarn ram-bundle --dev false --indexed-ram-bundle --platform android

% file dist/main.android.bundle
dist/main.android.bundle: data
```

Tree shaking is not compatible with the RAM bundle format. Verify that `treeShake` gets disabled if enabled:

```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index 557c102b..09af6757 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -86,7 +86,7 @@
           ]
         },
         "typescriptValidation": true,
-        "treeShake": false,
+        "treeShake": true,
         "targets": [
           "android",
           "ios",
```

Run:

```diff
 % yarn ram-bundle --dev false --platform ios
 yarn run v1.22.19
 $ react-native rnx-ram-bundle --dev false --platform ios
+info `treeShake` was disabled because it is not compatible with the RAM bundle format
 info Bundling ios...
                     Welcome to Metro!
               Fast - Scalable - Integrated


 info start
 info finish
 info Writing unbundle output to:, dist/main.ios.jsbundle
 info Writing sourcemap output to:, dist/main.ios.jsbundle.map
 info Done writing unbundle output
 info Done writing sourcemap output
 info Copying 1 asset files
 info Done copying assets
 ✨  Done in 3.88s.
```


tido64:
fyi @crazyfraggle

packages/config/src/bundleConfig.ts
```diff
@@ -82,6 +82,11 @@ export type BundleParameters = BundlerPlugins & {
    * Either absolute, or relative to the package. If not given, assets are ignored.
    */
   assetsDest?: string;
+
+  /**
+   * Builds JavaScript as a "Random Access Module" bundle for offline use.
```
tido64, line 87: `ramBundle` is incompatible with `treeShake`

afoxman (commented):
Can you reorganize this to use its own CLI command `rnx-ram-bundle` so that it matches existing CLI convention? Right now, `rnx-bundle` and `rnx-start` both go the great lengths to mirror and be drop-in replacements for `bundle` and `start`. If we do RAM bundling, we should have a similar drop-in command for that as well.

tido64:
> Can you reorganize this to use its own CLI command `rnx-ram-bundle` so that it matches existing CLI convention? Right now, `rnx-bundle` and `rnx-start` both go the great lengths to mirror and be drop-in replacements for `bundle` and `start`. If we do RAM bundling, we should have a similar drop-in command for that as well.

Would it make sense to keep this as-is, then add the `rnx-ram-bundle`? If/when we're getting a third bundle format, e.g. Hermes bytecode, we would have to add a third command if we're following this pattern. It would be a lot simpler if we introduced the `bundleFormat` option. I'm thinking this is something we can upstream as well.

packages/metro-service/src/bundle.ts
```diff
@@ -32,12 +35,19 @@ type RequestOptions = {
   unstable_transformProfile?: TransformProfile;
 };
 
-export async function bundle(args: BundleArgs, config: ConfigT): Promise<void> {
+export async function bundle(
```
afoxman, line 38: For RAM bundling, this should be a different function which has `RamBundleArgs` instead of `BundleArgs`, since they are slightly different (see `react-native ram-bundle --help`). For example, RAM bundling supports `--indexed-ram-bundle` as an argument.

As a separate function, it would call `buildBundleWithConfig` passing the 3rd arg as RamBundle, causing serialization to be RAM instead of CJS.

afoxman: The "fallback" logic could be lifted out and shared between them, with the changes you've made.

afoxman: This change should ripple out to the CLI, leading to a separation of CJS and RAM bundling -- args/types, variations like indexed-RAM-bundling (RAM only) and tree shaking (CJS only), and top-level CLI commands (which you already have in place).

packages/cli/src/metro-config.ts
```diff
@@ -146,7 +145,14 @@ export function customizeMetroConfig(
     plugins.push(CyclicDependencies());
   }
 
-  if (treeShake) {
+  // Tree shaking is currently incompatible with RAM bundle.
+  if (treeShake && bundleFormat === "ram-bundle" && print) {
```
afoxman, line 149: With the changes I proposed in `metro-service` and `config`, this check shouldn't be necessary since tree shaking can't be enabled. The tree shaking option isn't available for RAM bundling in the CLI or config.

packages/config/src/bundleConfig.ts
```diff
@@ -57,10 +57,22 @@ export type BundleParameters = BundlerPlugins & {
...
+   * Specifies the format of the output bundle.
+   *
+   * Note that setting the format to `ram-bundle` will disable `treeShake`.
+   * Metro ignores custom serializers when RAM bundling.
+   *
+   * @see https://facebook.github.io/metro/docs/bundling
+   *
+   * @default "plain"
+   */
+  bundleFormat?: "plain" | "ram-bundle";
```
afoxman, line 74: Since RAM bundling has slightly different params from CJS bundling, and since it is a separate CLI command, it should have its own config -- e.g. `ramBundle` prop next to `bundle` prop. When you run `rnx-ram-bundle`, `ramBundle` goes into effect.

The two differences I see are that RAM bundling supports an indexed bundle format and doesn't support tree shaking.

packages/cli/src/ram-bundle.ts
```diff
@@ -0,0 +1,27 @@
+import type { Config as CLIConfig } from "@react-native-community/cli-types";
+import type { CLIBundleOptions } from "./bundle";
+import { rnxBundle, rnxBundleCommand } from "./bundle";
+
+type CLIRamBundleOptions = Omit<CLIBundleOptions, "bundleFormat">;
```
afoxman, line 5: This type should include a prop (boolean) for `--indexed-ram-bundle`.

packages/cli/src/ram-bundle.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export async function rnxRamBundle(
+  argv: Array<string>,
+  cliConfig: CLIConfig,
+  cliOptions: CLIRamBundleOptions
+): Promise<void> {
+  return rnxBundle(argv, cliConfig, {
```
afoxman, line 12: Since CLIRamBundleOptions is different from CLIBundleOptions, you'll need to expand rnxBundle into this function, and make the "convergence" point be `metroBundle`. The override method will be different, for example. You can reuse `loadMetroConfig`. For `getCliPlatformBundleConfigs`, you'll need a new API for reading RAM bundle config (which probably reuses nearly all of the same internal logic).

packages/cli/README.md
```diff
@@ -139,8 +139,7 @@ yarn react-native rnx-start --host localhost --port 8812
         "ignoredModules": ["react-is"],
         "throwOnError": false
       },
-      "typescriptValidation": true,
-      "treeShake": true
```
afoxman, line 143: 😢

packages/cli/src/bundle/metro.ts
```diff
@@ -1,8 +1,10 @@
 import { info } from "@rnx-kit/console";
-import { bundle, BundleArgs as MetroBundleArgs } from "@rnx-kit/metro-service";
+import type { BundleArgs as MetroBundleArgs } from "@rnx-kit/metro-service";
+import { bundle } from "@rnx-kit/metro-service";
 import { createDirectory } from "@rnx-kit/tools-node/fs";
+import Bundle from "metro/src/shared/output/bundle";
```
afoxman, line 5: Can you push this down into metro-service as either an export (alongside the ram-bundle equivalent), or better yet an enum like "commonjs" | "ram"? I don't want the super-ugly metro imports we do to escape metro-service, if possible.

packages/cli/src/bundle/overrides.ts
```diff
@@ -13,29 +13,34 @@ export type BundleConfigOverrides = {
   assetsDest?: string;
   treeShake?: boolean;
   unstableTransformProfile?: TransformProfile;
+  indexedRamBundle?: boolean;
 };
 
+export const overridableBundleOptions: readonly (keyof BundleConfigOverrides)[] =
```
afoxman, line 19:
```suggestion
export const overridableCommonBundleOptions: readonly (keyof BundleConfigOverrides)[] =
```

This lines it up with the naming you established with `CLICommonBundleOptions`.

packages/cli/src/bundle/overrides.ts
```diff
@@ -3,7 +3,7 @@ import type { BundleArgs } from "@rnx-kit/metro-service";
 import { pickValues } from "@rnx-kit/tools-language/properties";
 import type { CliPlatformBundleConfig } from "./types";
 
-export type BundleConfigOverrides = {
+type BundleConfigOverrides = {
```
afoxman, line 6:
```suggestion
type BundleConfigOverrides = { Pick<CLICommonBundleOptions, "entryFile" | "bundleOutput" ... >};
```

This should be based on the names/types found in `CLICommonBundleOptions`.

packages/cli/test/bundle/overrides.test.ts
```diff
@@ -16,7 +19,11 @@ describe("CLI > Bundle > Overrides > applyBundleConfigOverrides", () => {
 
   test("has no effect when no overrides are given", () => {
     const copy = { ...config };
-    applyBundleConfigOverrides({}, [copy]);
+    applyBundleConfigOverrides(
+      {},
+      [copy],
+      [...overridableBundleOptions, "treeShake"]
```
afoxman, line 25: Test for indexed ram bundle config?

afoxman (commented):
Thanks for making these changes to align the commands with the RN cli.

packages/cli/README.md
afoxman: Tree shaking is so good. I want it to be shown in the docs. I know it's not compatible with ram bundling, so it needed to go... and that made me sad.

afoxman (approved):

packages/cli/README.md
tido64: This is was under the `server` config though 😛

PR #1851, opened 2022-08-26
fix(metro-config): ensure `server.enhanceMiddleware` is consistent with `transformer.assetPlugins`

tido64:
### Description

Ensure `server.enhanceMiddleware` is only set if `transformer.assetPlugins` includes the asset plugin.

### Test plan

Tests were added.

### DIFF
--- a/packages/metro-config/src/index.js
+++ b/packages/metro-config/src/index.js
@@ -20,10 +20,12 @@ const UNIQUE_PACKAGES = ["react", "react-native"];
  */
 function verifyMiddlewareConsistency(config) {
   if (!config.server || !config.server.enhanceMiddleware) {
-    // If our middleware was removed, we should also remove the corresponding
-    // asset plugin. Note that we can only check whether `enhanceMiddleware` is
-    // unset as we cannot guarantee that our middleware isn't embedded
-    // somewhere.
+    // If our middleware was removed, the corresponding asset plugin should also
+    // be removed to avoid issues. We should notify the user of this so they can
+    // handle it accordingly.
+    //
+    // Note that we can only check whether `enhanceMiddleware` is unset as we
+    // cannot guarantee that our middleware isn't embedded somewhere.
     const monorepoPluginIndex =
       config.transformer &&
       config.transformer.assetPlugins &&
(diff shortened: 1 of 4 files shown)


packages/metro-config/src/index.js
```diff
@@ -12,6 +12,38 @@ const path = require("path");
...
+  const { enhanceMiddleware } = require("./assetPluginForMonorepos");
+
+  const assetPlugins = config.transformer && config.transformer.assetPlugins;
+  const monorepoPluginIndex = assetPlugins
+    ? assetPlugins.indexOf(require.resolve("./assetPluginForMonorepos"))
+    : -1;
+
+  if (
+    !config.server ||
+    (config.server && config.server.enhanceMiddleware !== enhanceMiddleware)
```
afoxman, line 31: Checking `=== enhanceMiddleware` isn't "deep" enough to know for sure if our middleware was removed. It may be wrapped in another function.

We do this in the CLI: https://github.com/microsoft/rnx-kit/blob/main/packages/cli/src/start.ts#L162

I don't think you can actually tell when middleware has been removed from the execution chain... :(

We need to trust that the dev doing config knows how to properly preserve our defaults, including assets and middleware.

afoxman (commented):
I'm not sure this change is workable (I left a comment as to why). Can you shed some more light on what is driving it? Maybe there's another approach that can be used.

packages/metro-config/src/index.js
afoxman: How about failing the Metro run with an exception, rather than quietly modifying the config?

Help them get their config in order, rather than quietly fix it for them... otherwise they won't understand or learn.

tido64: > Help them get their config in order, rather than quietly fix it for them... otherwise they won't understand or learn.

That's a really good point. I turned it into a warning instead. The dev server will still work in most cases regardless.

packages/metro-config/src/index.js
```diff
@@ -12,6 +12,35 @@ const path = require("path");
...
 
+/**
+ * Verifies `server.enhanceMiddleware` is only set if
+ * `transformer.assetPlugins` includes the asset plugin.
+ * @param {MetroConfig} config
+ * @returns {MetroConfig}
+ */
+function verifyMiddlewareConsistency(config) {
+  if (!config.server || !config.server.enhanceMiddleware) {
+    // If our middleware was removed, we should also remove the corresponding
```
afoxman, line 23: Nit: update this comment

afoxman (approved):

PR #1854, opened 2022-08-29
Change file not found error to warning

jmkitavi:
### Description
Currently when aligning dependencies and rnx-kit/dep-check finds a package without a package.json it exits with an error, blocking subsequent steps.
Changing this error to a warning to alert the user of missing file but still complete the runs.

### Test plan
Align dependencies for packages that has a missing package.json file.
This should show a warning and not error out

### DIFF
--- a/.changeset/olive-waves-sin.md
+++ b/.changeset/olive-waves-sin.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/dep-check": patch
+---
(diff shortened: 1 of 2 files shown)


tido64 (approved):

tido64:
Thanks for submitting this! Can you run `yarn change` and add a change log entry?

PR #1857, opened 2022-08-30
Update JS comments in an example

retyui:


tido64 (approved):
Thanks! That was a glaring mistake 😆

Issue #142, opened 2021-04-19
Metro: Also enable `disableImportExportTransform` when `experimentalImportSupport` is


tido64:
When enabling `experimentalImportSupport` in `metro.config.js`:

```js
module.exports = {
  transformer: {
    getTransformOptions: async () => ({
      transform: {
        experimentalImportSupport: true,
        inlineRequires: true,
      },
    }),
  },
};
```

`disableImportExportTransform` should also be enabled in `babel.config.js`:

```js
module.exports = {
  presets: [
    [
      "module:metro-react-native-babel-preset",
      { disableImportExportTransform: true },
    ],
  ],
};
```

When `babel.config.js` is missing, `disableImportExportTransform` is set to the value of `experimentalImportSupport` in [`metro-react-native-babel-transformer`](https://github.com/facebook/metro/blob/fa096594bcdeda8fe529aba94604016a1c3cbf93/packages/metro-react-native-babel-transformer/src/index.js#L93). Since we're recommending folks to use our preset, we need to find a way to keep both in sync somehow.

denissb:
Spent quite some time figuring this out today. I think enabling `experimentalImportSupport` needs to be explicitly mentioned in the docs herehttps://microsoft.github.io/rnx-kit/docs/guides/bundling#tree-shaking.

PR #1865, opened 2022-09-04
Update tree shaking docs

afoxman:
### Description

The tree shaking section of the `Bundling` guide omits an important setup step -- setting `experimentalImportSupport` in metro.config.js. This is called out in #142 as something we should do automatically, but we don't have a fix for that yet.

> Note: bug #142 calls out `experimentalImportSupport` but also sets `inlineRequires` to `true`, so I called out the need for both. If that's wrong, please let me know in a PR comment.

Thank you to @denissb for pointing this out (and sorry it caused you so much trouble).

### Test plan

Manually verify the docsite using a local server.

### DIFF
--- a/docsite/docs/guides/bundling.mdx
+++ b/docsite/docs/guides/bundling.mdx
@@ -261,36 +261,6 @@ Tree shaking is _off_ by default. You can control it with the
 }
 ```
 
-To make tree shaking to work properly, you'll also need to set
-`disableImportExportTransform` in your Babel config:
-
-```js title="babel.config.js"
-module.exports = {
-  presets: [
-    [
-      "@rnx-kit/babel-preset-metro-react-native",
-      { disableImportExportTransform: true },
-    ],
-  ],
-};
-```
-
-You'll also need to set `experimentalImportSupport` and `inlineRequires` in your
-Metro config:
-
-```js title="metro.config.js"
-module.exports = {
-  transformer: {
-    getTransformOptions: async () => ({
-      transform: {
-        experimentalImportSupport: true,
-        inlineRequires: true,
-      },
-    }),
-  },
-};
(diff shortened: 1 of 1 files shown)


docsite/docs/guides/bundling.mdx
```diff
@@ -275,6 +275,22 @@ module.exports = {
...
+```js title="metro.config.js"
+module.exports = {
+  transformer: {
+    getTransformOptions: async () => ({
+      transform: {
+        experimentalImportSupport: true,
+        inlineRequires: true,
+      },
+    }),
+  },
```
tido64, line 290: Both of these must be set to `false` for the tree shaking plugin to work:

https://github.com/microsoft/rnx-kit/blob/e56db00feda0cfbb58f9e609c6dacbc5c9ba0d82/packages/metro-serializer-esbuild/src/esbuildTransformerConfig.ts#L3-L21

In the [README](https://github.com/microsoft/rnx-kit/tree/main/packages/metro-serializer-esbuild#usage), we are recommending that people to use our config so we can control everything else that also need to be set:

```suggestion
  transformer: esbuildTransformerConfig,
```

tido64 (changes requested):

docsite/docs/guides/bundling.mdx
tido64: FYI: #1871

afoxman: Got it. In the bundling guide, I already have them using our babel preset, so it looks like with #1871 we no longer need ANY special babel or metro config.

I removed the unneeded `transform` flags, and I also removed the now-unneeded `esbuild` transform profile.

tido64 (approved):

PR #1871, opened 2022-09-05
fix: use `esbuild` transform profile when `metro-serializer-esbuild` is used

### LINKED ISSUE #142: Metro: Also enable `disableImportExportTransform` when `experimentalImportSupport` is

tido64:
### Description

Use `esbuild` transform profile when `metro-serializer-esbuild` is used.

Resolves #142.

### Test plan

```
cd packages/test-app
yarn build --dependencies
yarn bundle+esbuild --dev false --platform ios --reset-cache
```

| Before | After |
| :- | :- |
| 684314 | 585154 |

### DIFF
--- a/packages/metro-serializer-esbuild/src/index.ts
+++ b/packages/metro-serializer-esbuild/src/index.ts
@@ -123,6 +123,9 @@ export function MetroSerializer(
 ): SerializerConfigT["customSerializer"] {
   assertVersion(">=0.66.1");
 
+  // Signal to every plugin that we're using esbuild.
+  process.env["RNX_METRO_SERIALIZER_ESBUILD"] = "true";
+
   return (
     entryPoint: string,
     preModules: ReadonlyArray<Module>,
--- a/packages/metro-serializer-esbuild/src/index.ts
+++ b/packages/metro-serializer-esbuild/src/index.ts
@@ -124,7 +124,7 @@ export function MetroSerializer(
   assertVersion(">=0.66.1");
 
   // Signal to every plugin that we're using esbuild.
-  process.env[`RNX_METRO_SERIALIZER_ESBUILD_${process.pid}`] = "true";
+  process.env["RNX_METRO_SERIALIZER_ESBUILD"] = "true";
 
   return (
     entryPoint: string,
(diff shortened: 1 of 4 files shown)


packages/metro-serializer-esbuild/src/index.ts
```diff
@@ -123,6 +123,9 @@ export function MetroSerializer(
 ): SerializerConfigT["customSerializer"] {
   assertVersion(">=0.66.1");
 
+  // Signal to every plugin that we're using esbuild.
+  process.env[`RNX_METRO_SERIALIZER_ESBUILD_${process.pid}`] = "true";
```
afoxman, line 127: Why include the PID? It will only propagate to child processes. And since it's a signal to all of them, do you need the PID? I feel like I'm missing something.

afoxman (approved):

packages/metro-serializer-esbuild/src/index.ts
tido64: It's for the case where you are running two bundlers at the same time. We can't guarantee that both are using this serializer.

tido64: Actually, I may have overthought this. The environment variables shouldn't be shared between two separate processes.

PR #1874, opened 2022-09-06
fix(build): move some build scripts to a separate file

tido64:
### Description

Move some build scripts to a separate file so they can be shared.

### Test plan

```
cd incubator/build
yarn build --dependencies

# macOS
yarn rnx-build -p macos ../../packages/test-app

# iOS simulator
yarn rnx-build -p ios ../../packages/test-app

# iOS device
yarn rnx-build -p ios --deploy local-only --device-type device ../../packages/test-app

### DIFF
--- a/incubator/build/scripts/build-apple.sh
+++ b/incubator/build/scripts/build-apple.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -eo pipefail
+set -eox pipefail
 
 case ${1-} in
   archive)
@@ -30,7 +30,7 @@ case ${1-} in
           ;;
         *)
           echo "Unknown option: $1"
-          shift
+          exit 1
           ;;
       esac
     done
@@ -71,7 +71,7 @@ case ${1-} in
           ;;
         *)
           echo "Unknown option: $1"
-          shift
+          exit 1
           ;;
       esac
     done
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: incubator/build/scripts/build-apple.sh [hunk 1]
```diff
@@ -0,0 +1,115 @@
...
+        --device-type)
+          device_type="$2"
+          shift 2
+          ;;
+        --scheme)
+          scheme="$2"
+          shift 2
+          ;;
+        *)
+          echo "Unknown option: $1"
```
afoxman, line 32: Should this fail the script? I see it as an "internal error" -- e.g. "if you got here, something is wrong with the code".

incubator/build/scripts/build-apple.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+set -eo pipefail
+
```
afoxman, line 3: Will the individual shell commands show up in the dev terminal? how about in the CI log? It's important context when reviewing the stdout/stderr text.

afoxman (approved):

incubator/build/scripts/build-apple.sh
tido64: This should be output now:

```
% npx --prefix . rnx-build-apple build-macos --scheme ReactTestApp --test
+ case ${1-} in
+ shift
+ [[ 3 -gt 0 ]]
+ case $1 in
+ scheme=ReactTestApp
+ shift 2
+ [[ 1 -gt 0 ]]
+ case $1 in
+ echo 'Unknown option: --test'
Unknown option: --test
+ exit 1
```

PR #1877, opened 2022-09-08
feat(dep-check): introduce preset with New Arch packages

tido64:
### Description

Introduce a `@rnx-kit/dep-check` preset containing only packages that are compatible with New Architecture.

### Test plan

TODO


kelset:
(most likely not needed anymore thanks to the interop layer)

PR #1881, opened 2022-09-09
fix(metro-config): fix asset plugin breaking bundling

### LINKED ISSUE #1879: metro-config: `@@` in asset paths when bundling
### What happened?

`transformer.assetPlugins` is always applied whether we are bundling or serving, but `server.enhanceMiddleware` is ignored during bundling. This causes `@@` to show up in asset paths because they are not reverted.

### Affected Package

@rnx-kit/metro-config

### Version

1.2.38

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
n/a
```

### Steps to Reproduce

n/a

### Code of Conduct

tido64:
### Description

The asset plugin breaks bundling because it replace `..` in paths with `@@`. This plugin is only intended as a workaround for a Metro issue with serving assets from relative paths, and should not be used when bundling.

Resolves #1879.

### Test plan

Tested in an internal app.

To test this locally, you can modify the plugin to output something, e.g.:

```diff
diff --git a/packages/metro-config/src/assetPluginForMonorepos.js b/packages/metro-config/src/assetPluginForMonorepos.js
index b6186449..c1808e3c 100644
--- a/packages/metro-config/src/assetPluginForMonorepos.js
+++ b/packages/metro-config/src/assetPluginForMonorepos.js
@@ -8,6 +8,8 @@
  * }} Server;
  */

+let didCallAssetPlugin = false;
+
 /**
  * Metro doesn't support assets in a monorepo setup. When the app requests
  * assets at URLs such as `/assets/../../../node_modules/react-native/<...>`,
@@ -22,6 +24,10 @@
  * @returns {AssetData}
  */
 function assetPlugin(assetData) {
+  if (!didCallAssetPlugin) {
+    didCallAssetPlugin = true;
+    console.log("Asset plugin used in worker", process.pid);
+  }
   assetData.httpServerLocation = assetData.httpServerLocation.replace(
     /\.\.\//g,
     "@@/"
```

Then try comparing `yarn bundle --platform ios --reset-cache` with `yarn start` in `packages/test-app`. Only the latter should hit the added lines.

### DIFF
--- a/.changeset/dull-eggs-mate.md
+++ b/.changeset/dull-eggs-mate.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-config": patch
+---
+
+Fix monorepo asset plugin breaking bundling
--- a/packages/metro-config/src/assetPluginForMonorepos.js
+++ b/packages/metro-config/src/assetPluginForMonorepos.js
@@ -1,6 +1,11 @@
 /**
  * @typedef {import("metro").AssetData} AssetData;
+ * @typedef {import("metro-config").ConfigT} ConfigT;
  * @typedef {import("metro-config").Middleware} Middleware;
+ *
+ * @typedef {import("metro/src/Server") & {
+ *   _config?: ConfigT;
+ * }} Server;
  */
 
 /**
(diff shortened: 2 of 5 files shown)


afoxman (approved):
Dangerous yet really clever 😛. 50 points to Gryffindor.

PR #1886, opened 2022-09-13
feat(metro-plugin-duplicates-checker): differentiate between banned and duplicate modules

tido64:
### Description

The error message today is not clear when banned modules are encountered, causing confusion.

BREAKING CHANGES: Signature of `checkForDuplicatePackages` changed to return an object to differentiate between banned and duplicate modules.

### Test plan

Tests were updated accordingly.


tido64:
cc @sverrejoh

afoxman (approved):

PR #1888, opened 2022-09-14
fix(metro-config): add "react-native" to resolverMainFields

crazyfraggle:
### Description

For packages that include both `react-dom` and `react-native` code and differentiate by using the `"react-native"` directive in `package.json`,  the current list of mainFields will pick the `react-dom` module.

Seen in the wild with projects using recoiljs: https://github.com/facebookexperimental/Recoil

This change adds "react-native" as the preferred main field.

### Test plan

Include recoiljs (or other package using split module declarations) in a project and build.

### DIFF
--- a/.changeset/proud-rice-bake.md
+++ b/.changeset/proud-rice-bake.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-config": minor
+---
+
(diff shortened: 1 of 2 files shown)


tido64 (approved):
Thanks :heart:

PR #1896, opened 2022-09-21
Quick security and Windows-compatibility fix

warmsea:
This is a one-line quick fix for 2 issues:

1. The small issue is for **Windows compatibility**. Assuming `androidProject` is `D:\Projects\abc`. Notice that the path separator is backslash in Windows and slash in Linux/MacOS. The original code is write `file("D:\Projects\abc")` in the files which is not valid. After the fix it will write `file("D:\\Projects\\abc")` that the backslashes are correctly escaped.
2. The big issue is for **security**. Linux/MacOS allows a lot of special characters in file name. So hackers can potentially name there project `");doSomethingBad();("` and hack whoever uses it. This fix will correctly escape it and make sure it's just a string.

I don't have the knowledge to decide whether I should update `targetName` in the line as well so I'm not touching it and assuming it's safe.

### DIFF
--- a/packages/cli/test/copy-assets/assembleAarBundle.test.ts
+++ b/packages/cli/test/copy-assets/assembleAarBundle.test.ts
@@ -168,7 +168,7 @@ describe("assembleAarBundle", () => {
           /[/\\]node_modules[/\\].rnx-gradle-build[/\\]rnx-kit_react-native-auth[/\\]settings.gradle$/
         ),
         expect.stringMatching(
-          /include\(":rnx-kit_react-native-auth"\)\nproject\(":rnx-kit_react-native-auth"\).projectDir = file\(".*?[/\\\\]node_modules[/\\\\]@rnx-kit[/\\\\]react-native-auth[/\\\\]android"\)/
+          /include\(":rnx-kit_react-native-auth"\)\nproject\(":rnx-kit_react-native-auth"\).projectDir = file\(".*?(\/|\\\\)node_modules(\/|\\\\)@rnx-kit(\/|\\\\)react-native-auth(\/|\\\\)android"\)/
         ),
       ],
       [
--- a/packages/cli/test/copy-assets/assembleAarBundle.test.ts
+++ b/packages/cli/test/copy-assets/assembleAarBundle.test.ts
@@ -168,7 +168,7 @@ describe("assembleAarBundle", () => {
           /[/\\]node_modules[/\\].rnx-gradle-build[/\\]rnx-kit_react-native-auth[/\\]settings.gradle$/
         ),
         expect.stringMatching(
-          /include\(":rnx-kit_react-native-auth"\)\nproject\(":rnx-kit_react-native-auth"\).projectDir = file\(".*?[/\\]node_modules[/\\]@rnx-kit[/\\]react-native-auth[/\\]android"\)/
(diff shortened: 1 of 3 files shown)


tido64 (changes requested):
@warmsea: Thanks for submitting this. I'm not sure I fully understand this attack vector. `androidProject` is set by [findGradleProject](https://github.com/microsoft/rnx-kit/blob/dea2d815f9175c403c41e88367df969a431753a2/packages/cli/src/copy-assets.ts#L76), and is not set by users. How do we end up in the scenario where 1) the Windows path is invalid and 2) people can inject potentially harmful code?

warmsea:
Hi @tido64 

For (1) it is generating the following file in my project:

```
include(":react-native-svg")
project(":react-native-svg").projectDir = file("D:\Projects\client-android\ORK\node_modules\react-native-svg\android")
```

I'm expecting it to generate:

```
include(":react-native-svg")
project(":react-native-svg").projectDir = file("D:\\Projects\\client-android\\ORK\\node_modules\\react-native-svg\\android")
```

For (2), I'm simply considering file path is a kind of "user input" and in principle I prefer not to trust any user input and put it directly into something that will be executed.

tido64 (approved):
Thanks for the explanation. I haven't been in here for a while 😄 

Could you run `yarn change` at the repo root to generate a change log entry, and fix the test? I think we should be good to go then.

packages/cli/test/copy-assets/assembleAarBundle.test.ts
```diff
@@ -168,7 +168,7 @@ describe("assembleAarBundle", () => {
           /[/\\]node_modules[/\\].rnx-gradle-build[/\\]rnx-kit_react-native-auth[/\\]settings.gradle$/
         ),
         expect.stringMatching(
-          /include\(":rnx-kit_react-native-auth"\)\nproject\(":rnx-kit_react-native-auth"\).projectDir = file\(".*?[/\\]node_modules[/\\]@rnx-kit[/\\]react-native-auth[/\\]android"\)/
+          /include\(":rnx-kit_react-native-auth"\)\nproject\(":rnx-kit_react-native-auth"\).projectDir = file\(".*?[/\\\\]node_modules[/\\\\]@rnx-kit[/\\\\]react-native-auth[/\\\\]android"\)/
```
tido64, line 171: To allow multiple backslashes, we want to add `+`: 
```suggestion
          /include\(":rnx-kit_react-native-auth"\)\nproject\(":rnx-kit_react-native-auth"\).projectDir = file\(".*?[/\\]+node_modules[/\\]+@rnx-kit[/\\]+react-native-auth[/\\]+android"\)/
```

tido64: Alternatively, I think `{1,2}` should also work.

warmsea: Oh just saw this. Sorry I haven't wrote RegEx for a while. Kind of forgetting things. I actually want to disallow `\` which is invalid, and allow `\\` only. I'm hoping my next try works :)

warmsea: @tido64 I think it's ready now. Thanks for your quick response and help on this one!

tido64:

PR #1901, opened 2022-09-23
Fix bug in workspace detection

scottmas:
### Description
The algorithm for detecting workspace roots was flawed for npm and yarn. Per the spec (see https://classic.yarnpkg.com/lang/en/docs/workspaces/) , yarn and npm declare their workspaces by setting a `workspaces` property on the `package.json`. I was experiencing this in my RN repo and banging my head for a long time before I realized the root cause. 

### Test plan
Tests have been added. 

But to manually reproduce, inside a mono repo, set a yarn.lock or package-lock.json in one of the packages. The package will be incorrectly identified as the workspace root.

### DIFF
--- a/packages/tools-workspaces/src/common.ts
+++ b/packages/tools-workspaces/src/common.ts
@@ -1,4 +1,5 @@
 import type { Options } from "fast-glob";
+import fs from "fs";
 import fg from "fast-glob";
 import findUp from "find-up";
 import * as path from "path";
@@ -9,28 +10,18 @@ type PackageManager = {
 };
 
 export const LERNA_JSON = "lerna.json";
-export const PACKAGE_LOCK_JSON = "package-lock.json";
 export const PNPM_WORKSPACE_YAML = "pnpm-workspace.yaml";
 export const RUSH_JSON = "rush.json";
-export const YARN_LOCK = "yarn.lock";
+export const PACKAGE_JSON = "package.json";
 
 export const WORKSPACE_ROOT_SENTINELS = [
   LERNA_JSON,
   RUSH_JSON,
-  YARN_LOCK,
-  PACKAGE_LOCK_JSON,
   PNPM_WORKSPACE_YAML,
+  PACKAGE_JSON,
 ];
 
-function makeFindSentinel<R>(finder: (name: string[]) => R) {
-  let result: R | undefined;
-  return () => {
-    if (process.env.JEST_WORKER_ID || !result) {
-      result = finder(WORKSPACE_ROOT_SENTINELS);
-    }
-    return result;
-  };
-}
+const PACKAGE_JSON_REGEX = new RegExp(PACKAGE_JSON + "$", "i");
 
 function makeFindPackages<R>(
   glob: (patterns: string[], options: Options) => R,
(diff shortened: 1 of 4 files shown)


scottmas:
@microsoft-github-policy-service agree RecTech Holdings

tido64:
@scottmas: Thanks for submitting this PR. I'm not sure I understand the scenario you're describing. Does the package that contain the lock file belong to the root of workspaces? If so, why does it have its own lock file? The existence of the lock file indicates that it is its own workspace, i.e. it does not share `node_modules` folder, nor can it reference other packages by name.

scottmas:
Any updates on this? Would be nice to get merged in so I don't have to patch the package locally

afoxman:
> Any updates on this? Would be nice to get merged in so I don't have to patch the package locally

Hi @scottmas, can you answer tido64's question above?

scottmas:
Oh so sorry, for some reason I didn't see that comment. There's actually lots of reasons why a yarn.lock or package-lock.json file could exist inside a workspace dependency. For instance, when cloning a boilerplate repo, or maybe the developer is just messing around with package managers and left a gitignored lock file inside a dependency. There could be any million of reasons. It's also important to recognize that project dependencies may very well contain their own node_modules AND they very well may contain their own lock file (E.g. pnpm has a setting `shared-workspace-lockfile = false` that makes each project dependency have their own lock file).

And probably even more important in my mind, is that the mere existence of the lock file does NOT in itself prove that the folder is a workspace root. Each package manager has it's own rules for determining what is a workspace. As is, rnx-kit is essentially relying on a heuristic rather than the specifications themselves. This has already lead to trouble for the reasons I've mentioned and it's bound to lead to more trouble.

tido64:
If the reason is that someone accidentally left a lock file, then I'd argue that those files should be cleaned up instead. If you look in `rnx-kit`, under `docsite`. This is a separate package within our monorepo that is not part of the root workspaces. We have this kind of setup elsewhere as well, usually to isolate the dependencies of the package from the rest of the repo, either to speed up installs or prevent hoisting issues. If you run `findWorkspacePackages()` here, this change now returns the list packages in the root workspaces when it should return an _empty_ array. If you also run `yarn workspaces info` here, Yarn will tell you that it cannot find the root of our workspace. In code, I found that the workspace root (if found) is necessarily the same location as the lockfile: https://github.com/yarnpkg/yarn/blob/158d96dce95313d9a00218302631cd263877d164/src/config.js#L255-L256

npm doesn't have an equivalent command, but I ran the following experiment:

```
# From repo root:
npm run -w @rnx-kit/scripts build

# This should succeed. Now try running inside `packages/test-app`:
cd packages/test-app
npm run -w @rnx-kit/scripts build

# Next, try running inside `docsite`:
cd ../../docsite
npm run -w @rnx-kit/scripts build
```

The last command fails. npm obviously cannot have any knowledge of `yarn.lock`, so I'm guessing that it checks the workspaces configuration and determined that this package is not part of it. To me, it sounds like we should do the same (and take the perf hit) if we want to be 100% accurate. If we decide to do this, we will also have to fix upstream as we don't want to deviate from it: https://github.com/microsoft/workspace-tools/blob/5c182740a1fb88630c8df788070bf222dce25764/packages/workspace-tools/src/workspaces/implementations/index.ts

Thoughts? @afoxman? @kelset?

> It's also important to recognize that project dependencies may very well contain their own node_modules AND they very well may contain their own lock file (E.g. pnpm has a setting `shared-workspace-lockfile = false` that makes each project dependency have their own lock file).

If the issue is that we don't support `shared-workspace-lockfile = false`, then we should get that fixed.

kelset:
I agree with Tommy here.

The "reasons" mentioned seems to all fall into the bucket of "something went wrong accidentally" and I don't think we should support them.

I don't know in details how pnpm's `shared-workspace-lockfile = false` works, let's maybe start there - submitting a PR tailored to supporting that use case with specific tests for that.

scottmas:
Guys, the fact remains that the workspace detection method is not according to the specifications. If the workspace specs are silent on the presence of lock files, then it isn't technically "incorrect" or "user error" to have lock files. Who are we to say they "shouldn't" have a lock file on _their_ file system?

The breaking test you describe is actually a good example of this. This PR actually is **NOT** yet fully spec compliant. Technically, we need to add additional logic after we detect a candidate workspace root to see if the cwd is actually _included_ in the workspace packages glob. The pseudo code for fully spec compliant workspace detection looks something like the following:

```
1. Detect the nearest upwards package.json and get it's absolute path. This the requesting package.
2. Go up until we find a candidate workspace and get the workspace `packages` glob
3. Get a list of packages in the candidate workspace using the glob
4. If the requesting package exists in the list of packages found in the candidate workspace, this is the workspace root! If not, return to step 2
```

tido64:
I think it makes sense to have a spec compliant algorithm, but we also want to make sure that we don't deviate too much from [upstream](https://github.com/microsoft/workspace-tools/blob/5c182740a1fb88630c8df788070bf222dce25764/packages/workspace-tools/src/workspaces/implementations/index.ts) because a lot of tools are expecting similar behaviour and performance characteristics. Upstream changed to lockfiles because it is significantly faster in big repos (albeit at the expense of not being spec compliant), and we don't want to lose that.

Would it make sense to add the spec compliant implementation behind a flag, like an environment variable or similar? I don't know how this is blocking you: Are you using this library directly, or did you discover this through one of our tools?

scottmas:
It honestly doesn't matter too much to me, but it is a persistent paper cut when using rnx-kit. I'll close this PR (since it's not spec compliant either) and you guys can decide what you want to do next 🤷

Issue #1890, opened 2022-09-15
RFC 0001: dep-check v2


afoxman:
This issue tracks commits associated with [RFC 0001: dep-check v2](https://github.com/microsoft/rnx-kit/pull/1757).

Please reference it when making PRs against the repo.

- Migration:
  - [x] https://github.com/microsoft/rnx-kit/pull/1927
  - [x] https://github.com/microsoft/rnx-kit/pull/1931
  - [x] https://github.com/microsoft/rnx-kit/pull/1937
  - [x] https://github.com/microsoft/rnx-kit/pull/1940
  - [x] https://github.com/microsoft/rnx-kit/pull/1911
  - [x] https://github.com/microsoft/rnx-kit/pull/1930
  - [x] https://github.com/microsoft/rnx-kit/pull/1938
  - [x] https://github.com/microsoft/rnx-kit/pull/1948
  - [x] https://github.com/microsoft/rnx-kit/pull/1949
  - [x] https://github.com/microsoft/rnx-kit/pull/1950
  - [x] https://github.com/microsoft/rnx-kit/pull/1951
  - [x] https://github.com/microsoft/rnx-kit/pull/1953
  - [x] https://github.com/microsoft/rnx-kit/pull/1958
  - [x] https://github.com/microsoft/rnx-kit/pull/1972
  - [x] https://github.com/microsoft/rnx-kit/pull/1973
  - [x] https://github.com/microsoft/rnx-kit/pull/1975
  - [x] https://github.com/microsoft/rnx-kit/pull/1959
  - [x] https://github.com/microsoft/rnx-kit/pull/1960
  - [x] https://github.com/microsoft/rnx-kit/pull/1977
  - [x] https://github.com/microsoft/rnx-kit/pull/1979
  - [x] https://github.com/microsoft/rnx-kit/pull/1980
  - [x] https://github.com/microsoft/rnx-kit/pull/2023
- Release:
  - [x] https://github.com/microsoft/rnx-kit/pull/1981
  - [x] https://github.com/microsoft/rnx-kit/pull/1978
- Post-release:
  - [ ] https://github.com/microsoft/rnx-kit/pull/1976
  - [x] https://github.com/microsoft/rnx-kit/pull/1941

tido64:
I've created https://github.com/microsoft/rnx-kit/issues/2031 to handle post-release tasks. Closing this issue as completed. Thanks again, everyone!

PR #1911, opened 2022-09-30
refactor(dep-check): introduce the new configuration schema

tido64:
### Description

Introduces the new config schema. In particular, the old -> new config migration is implemented along with the check command. This can be flipped on with an env variable.

### Test plan

Build align-deps:

```
cd packages/align-deps
yarn build --dependencies
cd ../test-app
```

Change the test app's target prod version to 0.70:

```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index 43107f04..3a138093 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -61,7 +61,7 @@
     "preset": "@rnx-kit/scripts"
   },
   "rnx-kit": {
-    "reactNativeVersion": "^0.68",
+    "reactNativeVersion": "^0.70",
     "kitType": "app",
     "build": {
       "distribution": [
```

Run align-deps:

```
node ../align-deps/lib/cli.js
```

Follow the instructions on screen, and repeat the command. The warnings should go away.


packages/dep-check/src/check.ts
```diff
@@ -151,8 +161,183 @@ export function checkPackageManifest(
...
+}
+
+function resolve(
+  { kitType, alignDeps, manifest }: AlignDepsConfig,
+  projectRoot: string,
+  options: CheckOptions
+) {
+  const { capabilities, presets, requirements } = alignDeps;
+
+  const initialProdPreset = v2_profilesSatisfying(
```
tido64, line 231: Missing handling of empty preset being returned here.

packages/dep-check/src/compatibility/config.ts
```diff
@@ -0,0 +1,85 @@
...
+  );
+  return {
+    kitType,
+    alignDeps: {
+      presets: [
+        "microsoft/react-native",
+        ...(customProfilesPath ? [customProfilesPath] : []),
+      ],
+      requirements:
+        kitType === "app"
```
afoxman, line 49: Can this be simplified down to just the object? I thought I saw in other areas, where this is parsed, that if it's an object, `production` will be used over `development` as appropriate.

packages/dep-check/src/compatibility/config.ts
```diff
@@ -0,0 +1,85 @@
...
+      capabilities,
+    },
+    manifest,
+  };
+}
+
+export function migrateConfig(
+  config: AlignDepsConfig | CheckConfig
+): AlignDepsConfig {
+  if ("alignDeps" in config) {
```
afoxman, line 64: Should this be `hasOwnProperty` instead, which ignores inherited props?

packages/dep-check/src/compatibility/config.ts
```diff
@@ -0,0 +1,85 @@
...
+        .join(", ");
+      warn(`The following keys are no longer supported: ${unsupportedKeys}`);
+    }
+    return config;
+  }
+
+  const newConfig = transformConfig(config);
+  const { manifest, ...configOnly } = newConfig;
+
+  warn(`The config schema has changed. Please update your config to the following:
```
afoxman, line 78: Consider adding an example command they can use which will update it for them via --write? And maybe a link to the docs on the new schema so they can understand what they are looking at.

packages/dep-check/src/dependencies.ts
```diff
@@ -68,6 +70,114 @@ export function visitDependencies(
...
+  const trace: Trace[] = [
+    {
+      module: manifest.name,
+      profiles: Object.keys(preset),
+    },
+  ];
+
+  visitDependencies(manifest, projectRoot, (module, modulePath) => {
+    const kitConfig = getKitConfig({ cwd: modulePath });
+    if (!kitConfig) {
```
afoxman, line 90: Here's another place where we return early without a message when encountering a package without config.

packages/dep-check/src/presets/microsoft/react-native.ts
```diff
@@ -0,0 +1,29 @@
+import type { ProfileMap } from "../../types";
```
afoxman, line 1: Should this type change to `Preset`?

packages/dep-check/src/profiles.ts
```diff
@@ -184,6 +185,82 @@ export function profilesSatisfying(
   return profiles.filter((v) => versions.includes(v));
 }
 
+function v2_loadPreset(preset: string, projectRoot: string): Preset {
+  try {
+    return require("./presets/" + preset).default;
```
afoxman, line 190: Do we allow people to use align-deps via an API, or is it CLI-only? If there's an API, this `require` statement is going to cause problems for ESM projects.

packages/dep-check/src/profiles.ts
```diff
@@ -184,6 +185,82 @@ export function profilesSatisfying(
...
+function v2_compileRequirements(
+  requirements: string[]
+): ((pkg: MetaPackage | Package) => boolean)[] {
+  const includePrerelease = { includePrerelease: true };
+  return requirements.map((req) => {
+    const [requiredPackage, requiredVersionRange] = req.split("@");
+    return (pkg: MetaPackage | Package) => {
+      return (
+        pkg.name === requiredPackage &&
+        "version" in pkg &&
```
afoxman, line 205: `hasOwnProperty` instead of `in`?

afoxman (commented):
It's exciting to see align-deps coming along! Sorry this took me so long to get to. 

I left comments throughout the code, and had a few PR-wide comments as well:

1. The package is getting complex, and so the individual functions and code segments are going to be harder to "parse" and understand. 

One thing that would help is comments, both at the function level and in lengthy blocks that are doing key data transformations. 

Another is function names that express the data query or transform that is happening. For example, v2_profilesSatisfying. As an outsider who isn't immersed in dependency management/alignment challenges, I'm not sure what this name is telling me. Also, in this case, the function operates on requirements and presets, not profiles. So maybe "v2_filterAndMergePresetsSatisfyingRequirements". And this name is ugly, and leads me to think there are probably two or three functions it should split into.

2. Old vs New config. It seems like there's a straightforward, automatic way to internally transform from old to new. Internally, I see hints of the old config here and there. I'm wondering if you can localize these, pushing up to the CLI/API entry point? When an old config comes in, do a single transform, and then the other 95% of align-deps operates only on the new config. I realize this muddies error-reporting. e.g. telling users about 'preset X' and 'requirement R' in a message could confuse them if they're on the old config. They may still have enough info to puzzle it out, though, and you could append a link to new config in the error message if they want to learn more. There is value in both customer happiness and tool usability.

packages/dep-check/src/check.ts
```diff
@@ -151,8 +163,232 @@ export function checkPackageManifest(
...
+export function v2_getConfig(
+  manifestPath: string,
+  { uncheckedReturnCode }: CheckOptions
+): number | AlignDepsConfig | CheckConfig {
+  const manifest = readPackage(manifestPath);
+  if (!isPackageManifest(manifest)) {
+    error(
+      `'${manifestPath}' does not contain a valid package manifest — please make sure it's not missing 'name' or 'version'`
+    );
+    return 1;
```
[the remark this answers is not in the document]
tido64: Here we are returning if `package.json` is malformed, i.e. missing `name` and `version`. `uncheckedReturnCode` is only used in vigilant mode, but it's confusing as you pointed out. When I get to the vigilant mode, I will find a way to get rid of it. For now, I want to isolate changes to the check command only, and slowly migrate other parts.

packages/dep-check/src/check.ts
```diff
@@ -151,8 +163,232 @@ export function checkPackageManifest(
...
+    warn(
+      `Known bad packages are found in '${manifest.name}':\n` +
+        badPackages
+          .map((pkg) => `    ${pkg.name}@${pkg.version}: ${pkg.reason}`)
+          .join("\n")
+    );
+  }
+
+  const projectRoot = path.dirname(manifestPath);
+  const kitConfig = getKitConfig({ cwd: projectRoot });
```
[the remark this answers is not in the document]
tido64: The reason we don't print any warnings here is because we cannot expect all packages in a monorepo to be configured. Some packages may be web-only, or is a test/utility package that don't need management. That's why we have vigilant mode.

packages/dep-check/src/check.ts
```diff
@@ -151,8 +163,232 @@ export function checkPackageManifest(
...
+  }
+
+  const { devPreset, prodPreset, capabilities } = resolve(
+    config,
+    path.dirname(manifestPath),
+    options
+  );
+  const { kitType, manifest } = config;
+
+  info("Aligning dependencies according to the following profiles:");
```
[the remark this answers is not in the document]
tido64: I can distinguish the two by printing slightly different information, but the "mode" is hard-coded in your config. It's not something you would change on a regular basis.

packages/dep-check/src/check.ts
```diff
@@ -151,8 +163,232 @@ export function checkPackageManifest(
...
+
+  return 0;
+}
+
 export function makeCheckCommand(options: CheckOptions): Command {
-  return (manifest: string) => {
-    return checkPackageManifest(manifest, options);
-  };
+  const check =
+    process.env["DEP_CHECK_VERSION"] === "2"
```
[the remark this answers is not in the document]
tido64: I don't expect `DEP_CHECK_VERSION` to be long lived. It only exists now while I implement v2. Mostly for making testing easier. It's not meant to be public.

packages/dep-check/src/compatibility/config.ts
tido64: This config is also stringified and printed to the user further down. Added a comment here.

packages/dep-check/src/compatibility/config.ts
tido64: `hasOwnProperty` is problematic in that it can be overwritten. You would have to do something like `Object.prototype.hasOwnProperty(config, "alignDeps")`. Since `config` is just an object read from `package.json`, I think using `in` here is more legible.

packages/dep-check/src/compatibility/config.ts
tido64: Yes, I think it is mentioned in the RFC, but it won't be part of this PR. Added a TODO as a reminder.

packages/dep-check/src/dependencies.ts
tido64: This is when we visit transitive dependencies. We cannot expect random third party libraries to be configured.

packages/dep-check/src/presets/microsoft/react-native.ts
tido64: Yes, but changing this now will break v1. We will remove v1 once all commands have been migrated..

packages/dep-check/src/profiles.ts
tido64: Not really. This package is currently transformed to CJS, so the `require` statement here will be correctly handled.

packages/dep-check/src/profiles.ts
tido64: See reason above.

tido64 (commented):
I should have made this clearer in the description. The scope of this PR is limited to the new config schema, and the check command. As such, dep-check will be in a state of messy mash of both v1 and v2 until we've migrated everything.

Currently, the plan is to:

1. Migrate commands to the new config schema
2. Copy all code to a new package, align-deps
3. Remove all v1 code
4. Publish align-deps
5. Have dep-check wrap align-deps with a warning about the name change

But if this is confusing, maybe it makes more sense to swap 1 and 2 here. What do you think?

packages/dep-check/src/check.ts
```diff
@@ -151,8 +163,232 @@ export function checkPackageManifest(
...
+          bColor: chalk.green,
+        }
+      );
+      console.log(diff);
+
+      error(
+        "Changes are needed to satisfy all requirements. Re-run with `--write` to apply them."
+      );
+
+      const url = chalk.bold("https://aka.ms/align-deps");
```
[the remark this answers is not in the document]
tido64: It should be reserved now. 😄

tido64:
I will rebase this PR when https://github.com/microsoft/rnx-kit/pull/1927 lands.

packages/dep-check/src/check.ts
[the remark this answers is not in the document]
tido64: I changed my mind. It makes sense to output an error/warning here.

packages/align-deps/src/check.ts
```diff
@@ -87,33 +128,159 @@ export function getCheckConfig(
...
+  const prodRequirements = Array.isArray(requirements)
+    ? requirements
+    : requirements.production;
+  const mergedPreset = mergePresets(presets, projectRoot);
+  const initialProdPreset = filterPreset(prodRequirements, mergedPreset);
+  ensurePreset(initialProdPreset, prodRequirements);
+
+  const devPreset = (() => {
+    if (kitType === "app") {
+      // Preset for development is unused when the package is an app.
```
kelset, line 216: sidenote: we need to make sure this is clear in the docs - I can see people not be aware

packages/align-deps/src/presets/microsoft/react-native.ts
```diff
@@ -0,0 +1,29 @@
+import type { ProfileMap } from "../../types";
+import profile_0_61 from "./react-native/profile-0.61";
+import profile_0_62 from "./react-native/profile-0.62";
+import profile_0_63 from "./react-native/profile-0.63";
```
kelset, line 4: do we want to keep supporting versions older than what RNTA supports? Should we try to keep them aligned?

packages/test-app/package.json
```diff
@@ -155,6 +155,7 @@
       "core-ios",
       "core-macos",
       "core-windows",
+      "babel-preset-react-native",
```
kelset, line 158: is this change expected? leftover from some other change?

kelset (approved):
overall, given that the logic changes are already happening in the new package, I think we can go ahead and merge this.

Thanks for all your work Tommy, really appreciate that you keep the testing side up to date and enhanced with more tests.

I've left a few comments but they are all things not strictly related to the code

packages/align-deps/src/presets/microsoft/react-native.ts
tido64: We should probably prune some of the older versions. It's a bit cumbersome because they all depend on each other. Definitely should be done separately.

packages/test-app/package.json
tido64: Yes, we are using the capability, but it's currently unmanaged.

packages/dep-check/src/check.ts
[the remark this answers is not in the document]
afoxman: How are you tracking future work like this?

packages/dep-check/src/check.ts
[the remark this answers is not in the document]
afoxman: Got it, but the output would be more valuable if it mentioned your current mode. Especially when forensically reviewing CI logs. Having all of the data in there makes debugging easier.

packages/dep-check/src/check.ts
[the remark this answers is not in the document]
afoxman: It it's testing only and shouldn't be public, maybe take it out of the PR, or rename it to "TESTING_ONLY_DEP_CHECK_VERSION" or something ridiculous/obvious.

packages/dep-check/src/compatibility/config.ts
afoxman: Sorry, I don't follow. The user shouldn't have an expectation of array vs object since you're generating it for them. Why not always generate the "more complete" (?) object?

packages/dep-check/src/compatibility/config.ts
afoxman: There are one or two other TODOs from earlier comments. Maybe track these in an issue instead?

packages/dep-check/src/dependencies.ts
afoxman: That makes sense.

packages/dep-check/src/check.ts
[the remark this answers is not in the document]
tido64: I've already changed it in the next batch of changes 😛   I've changed `Command` to return an error code (string) instead of an arbitrary number. Here, we will return `invalid-manifest` instead of `1`. Instead of `uncheckedReturnCode || 1`, we will return `not-configured` and handle it accordingly.

I've added a list of future work in the description of https://github.com/microsoft/rnx-kit/issues/1890. Currently working migrating "vigilant" mode.

packages/dep-check/src/check.ts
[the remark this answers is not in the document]
tido64: This will be part of the next PR.

packages/dep-check/src/check.ts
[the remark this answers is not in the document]
tido64: It's already removed since we are doing work in `align-deps` now.

packages/dep-check/src/compatibility/config.ts
tido64: The `development` key is unused if `kitType == "library"`. Having it there doesn't make any sense.

packages/align-deps/src/preset.ts
```diff
@@ -0,0 +1,62 @@
...
+): ((pkg: MetaPackage | Package) => boolean)[] {
+  const includePrerelease = { includePrerelease: true };
+  return requirements.map((req) => {
+    const [requiredPackage, requiredVersionRange] = req.split("@");
+    return (pkg: MetaPackage | Package) => {
+      return (
+        pkg.name === requiredPackage &&
+        "version" in pkg &&
+        semverSatisfies(
+          semverCoerce(pkg.version) || "0.0.1",
```
afoxman, line 16: Is there an example of when `semverCoerce` would legitimately fail, or is this the "0.0.1" an unlikely fallback? I'm not sure `align-deps` should try to run if it encounters a package with an invalid manifest. I think it's reasonable to fail and bail, rather than risk operating on other potentially bad data.

afoxman (approved):
Seeing `align-deps` separate from `dep-check` makes things easier to understand. I have some new questions from this round of feedback, and a few unresolved comments from the last round.

I'll leave it to you to decide how and when to submit this PR and track the open items. I don't want to hold this up, but I would like those items addressed at some point.

If you submit this PR as-is, can you mark the align-deps version as 0.x or 0.0.x and use an '-alpha' prerelease tag for now?

packages/align-deps/src/cli.ts
```diff
@@ -112,124 +78,90 @@ async function makeCommand(args: Args): Promise<Command | undefined> {
...
 
-  process.exit(exitCode);
+  process.exitCode = errors;
 }
 
 if (require.main === module) {
   require("yargs").usage(
     "$0 [packages...]",
-    "Dependency checker for React Native apps",
+    "Dependency checker for npm packages",
```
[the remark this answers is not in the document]
tido64: Do you have any suggestions? This package is still private so there's still time to change it. I'll make it part of the docs update PR.

packages/align-deps/src/dependencies.ts
```diff
@@ -68,6 +70,114 @@ export function visitDependencies(
...
+    const capabilities =
+      kitConfig.alignDeps?.capabilities || kitConfig.capabilities;
+    if (Array.isArray(capabilities)) {
+      for (const capability of capabilities) {
+        allCapabilities.add(capability);
+      }
+    }
+
+    const filteredPreset = filterPreset(requirements, preset);
+    const filteredNames = Object.keys(filteredPreset);
```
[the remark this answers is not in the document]
tido64: I've added more comments around this in a separate PR, but what happens here is that we filter out profiles that do not fulfill the requirements. So `filteredNames` is the list of profile names that do satisfy the requirements.

packages/align-deps/src/dependencies.ts
```diff
@@ -68,6 +70,114 @@ export function visitDependencies(
...
+
+    const requirements = (() => {
+      const requirements = kitConfig.alignDeps?.requirements;
+      if (requirements) {
+        return Array.isArray(requirements)
+          ? requirements
+          : requirements.production;
+      }
+
+      if (kitConfig.reactNativeVersion) {
```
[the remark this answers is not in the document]
tido64: This for detecting old config in third-parties. I don't know if it makes sense to output warnings here as it may be too noisy.

packages/align-deps/src/compatibility/config.ts
```diff
@@ -1,11 +1,15 @@
...
+import semverCoerce from "semver/functions/coerce";
 import type { AlignDepsConfig, CheckConfig } from "../types";
 
 function dropPatchFromVersion(version: string): string {
   return version
     .split("||")
-    .map((v) => v.trim().split(".").slice(0, 2).join("."))
+    .map((v) => {
+      const coerced = semverCoerce(v);
+      return coerced ? `${coerced.major}.${coerced.minor}` : "0.0";
```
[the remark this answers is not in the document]
tido64: This function was moved in more recent changes. I'll make this throw in a later PR.

packages/align-deps/src/preset.ts
tido64: I think this would fail if you fed it random strings. I don't expect this to fail often, but I've changed it to throw instead. I will also change the other sites, but in separate PRs since I've moved stuff around a bit.

tido64:
@afoxman, @kelset: Thanks for reviewing! I'm going to merge this now and will try to address all your concerns in upcoming PRs. I have five lined up already and they all depend on another.

PR #1914, opened 2022-10-03
fix(metro-serializer-esbuild): resolve all files in a namespace

tido64:
### Description

We don't resolve files from disk, and it is required for certain functionalities such as app bundle preludes.

### Test plan

CI should pass.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1356,10 +1356,15 @@
     enabled "2.0.x"
     kuler "^2.0.0"
 
-"@esbuild/linux-loong64@0.15.5":
-  version "0.15.5"
-  resolved "https://registry.yarnpkg.com/@esbuild/linux-loong64/-/linux-loong64-0.15.5.tgz#91aef76d332cdc7c8942b600fa2307f3387e6f82"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -1356,10 +1356,15 @@
     enabled "2.0.x"
     kuler "^2.0.0"
 
-"@esbuild/linux-loong64@0.15.5":
-  version "0.15.5"
-  resolved "https://registry.yarnpkg.com/@esbuild/linux-loong64/-/linux-loong64-0.15.5.tgz#91aef76d332cdc7c8942b600fa2307f3387e6f82"
-  integrity sha512-UHkDFCfSGTuXq08oQltXxSZmH1TXyWsL+4QhZDWvvLl6mEJQqk3u7/wq1LjhrrAXYIllaTtRSzUXl4Olkf2J8A==
+"@esbuild/android-arm@0.15.10":
```
kelset, line 1359: where are these yarn.lock changes coming from? there's no package.json change that I can see here - leftover from a different PR?

kelset (approved):

yarn.lock
tido64: Oh, I bumped esbuild because I needed to make sure the error I was seeing internally wasn't because of the latest version.

PR #1924, opened 2022-10-07
fix(react-native-auth): `redirectURI` should be nullable

tido64:
### Description

`redirectUri` is already optional in TS, but not in the native projections.

### Test plan

CI should pass.


tido64:
cc @sweggersen

PR #1927, opened 2022-10-10
feat(align-deps): copy from dep-check

tido64:
### Description

This is a plain copy of dep-check.

### Test plan

CI should pass.


kelset (approved):
as a followup PR can you add in the readme something like "this package used to be called dep-check, you can read more about why the change happened here" and link to the RFC?

tido64:
> as a followup PR can you add in the readme something like "this package used to be called dep-check, you can read more about why the change happened here" and link to the RFC?

Yeah, we will have to rewrite the README. I'll make sure to include the RFC as well.

PR #1930, opened 2022-10-13
refactor(align-deps): migrate vigilant mode

tido64:
### Description

Migrates vigilant mode to the new config schema.

### Test plan

Build align-deps:

```
cd packages/align-deps
yarn build --dependencies
cd ../test-app
```

Remove the `test-app` capability from test app's `package.json`:

```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index 43107f04..aa7fc14c 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -50,7 +50,7 @@
     "@types/react-native": "^0.68.0",
     "jest-cli": "^27.5.1",
     "metro-react-native-babel-preset": "^0.67.0",
-    "react-native-test-app": "^1.3.5",
+    "react-native-test-app": "^1.0.0",
     "react-test-renderer": "17.0.2",
     "typescript": "^4.0.0"
   },
@@ -156,8 +156,7 @@
       "core-macos",
       "core-windows",
       "babel-preset-react-native",
-      "react",
-      "test-app"
+      "react"
     ]
   }
 }
```

Run align-deps:

```
node ../align-deps/lib/cli.js --requirements react-native@0.68
```

align-deps should complain about react-native-test-app even though it's not managed.

Tests were migrated as well.


packages/align-deps/src/commands/check.ts
```diff
@@ -0,0 +1,221 @@
...
+  const updatedManifest = updatePackageManifest(
+    manifest,
+    capabilities,
+    Object.values(prodPreset),
+    Object.values(devPreset),
+    kitType
+  );
+
+  // Don't fail when manifests only have whitespace differences.
+  const updatedManifestJson = JSON.stringify(updatedManifest, undefined, 2);
```
afoxman, line 160: For later - Have you every looked at what happens across devbox platforms with this diffing strategy? Do Windows CRLFs survive JSON.stringify (or Git checkout) when coming down on a Mac, and then show up here as diffs because !== is a strict char-by-char equivalence test?

packages/align-deps/src/commands/check.ts
```diff
@@ -0,0 +1,221 @@
...
+}
+
+export function makeCheckCommand(options: Options): Command {
+  const { presets = defaultConfig.presets, requirements } = options;
+  if (!requirements) {
+    return (manifest: string) => checkPackageManifest(manifest, options);
+  }
+
+  return (manifest: string) => {
+    const inputConfig = getConfig(manifest);
```
afoxman, line 192: This took some time to puzzle out... Why would config come back as a string? When I saw the error-handling below, I looked at getConfig() to see what possible errors could appear.

The API will be easier to read (and will more closely match a popular JS convention) if you separate error from result:

```
const { configError, inputConfig } = getConfig(manifest);
```

packages/align-deps/src/commands/vigilant.ts
```diff
@@ -0,0 +1,149 @@
...
+    "development"
+  );
+
+  const peerDependencies = updateDependencies(
+    {},
+    resolveCapabilities(unmanagedCapabilities, Object.values(supportPreset)),
+    "peer"
+  );
+
+  const { name, version } = require("../../package.json");
```
afoxman, line 74: This seems odd. Why pull in the `align-deps` package info to be used in the output manifest? Why not use the name/version from `manifestPath`?

afoxman (commented):
Like the last PR, I continue to struggle to make sense of what is going on in most of the code. I know the code is well-written, and well-tested. It's just doing a lot, and I am not as deeply in the mindset/space as you are. I'm bouncing off the surface in most cases, rather than going deeper and seeing the logic in my head.

So this review is more of a JS/TS examination rather than an `align-deps` correctness review.

What would help me are smaller functions, thorough comments demonstrating what data looks like going in/out of functions, why the transformations are happening, etc. Some of this could be encoded in more descriptive function/variable names, as well.

packages/align-deps/src/commands/check.ts
```diff
@@ -0,0 +1,221 @@
...
+      }
+      return "invalid-configuration";
+    }
+    return {
+      kitType,
+      alignDeps: {
+        ...defaultConfig,
+        ...alignDeps,
+      },
+      ...config,
```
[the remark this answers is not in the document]
tido64: In `migrateConfig` we check for old keys and ask people to remove them.

packages/align-deps/src/commands/check.ts
```diff
@@ -0,0 +1,221 @@
...
+    reactNativeDevVersion,
+    reactNativeVersion,
+  } = getKitCapabilities(config);
+
+  return {
+    kitType,
+    reactNativeVersion,
+    ...(config.reactNativeDevVersion ? { reactNativeDevVersion } : undefined),
+    capabilities,
+    customProfiles,
```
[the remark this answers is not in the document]
tido64: `customProfilesPath` was the "resolved" path of `customProfiles`. The types will be fixed here: https://github.com/tido64/rnx-kit/compare/tido/align-deps/init...tido64:rnx-kit:tido/align-deps/set-version?expand=1

packages/align-deps/src/commands/check.ts
tido64: Last I checked, `JSON.stringify` always uses Unix newlines. Regardless, since we are stringifying both objects here, we are guaranteed to be using the same newlines.

packages/align-deps/src/commands/check.ts
```diff
@@ -0,0 +1,221 @@
...
+    reactNativeDevVersion,
+    reactNativeVersion,
+  } = getKitCapabilities(config);
+
+  return {
+    kitType,
+    reactNativeVersion,
+    ...(config.reactNativeDevVersion ? { reactNativeDevVersion } : undefined),
```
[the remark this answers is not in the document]
tido64: It will be made optional in the "migrate `set-version`" PR: https://github.com/tido64/rnx-kit/compare/tido/align-deps/init...tido64:rnx-kit:tido/align-deps/set-version?expand=1

packages/align-deps/src/commands/check.ts
tido64: Would it help to have an `isError` helper function? I tried changing it, but I think it makes `checkPackageManifest` less elegant because we no longer can pass in just the config object:

image

Here I would have to do something like: `checkPackageManifest(manifest, options, { config, error: null })`

tido64:
> Like the last PR, I continue to struggle to make sense of what is going on in most of the code. I know the code is well-written, and well-tested. It's just doing a lot, and I am not as deeply in the mindset/space as you are. I'm bouncing off the surface in most cases, rather than going deeper and seeing the logic in my head.
> 
> So this review is more of a JS/TS examination rather than an `align-deps` correctness review.
> 
> What would help me are smaller functions, thorough comments demonstrating what data looks like going in/out of functions, why the transformations are happening, etc. Some of this could be encoded in more descriptive function/variable names, as well.
> 
> I'm voicing this because other engineers will be less likely to want to work on the code, or will struggle to get it correct, without a large up-front investment to understand it top-to-bottom.

I think this is a good call-out. In this phase, a lot of things are moving around still, and about half of the code (v1) will go away by the fourth/fifth PR. If you're up for it, we can walk through the code by the end and I will make a note of all the confusing parts that needs further refactoring/more documentation.

packages/align-deps/src/commands/vigilant.ts
afoxman: The change seems just as weird. You've given a fixed name and version. Will those turn into a real preset package? Are they a placeholder that will always be in the code?

afoxman (approved):
I still have on open question about the vigilant preset package...

And a request - please avoid rebasing as it detaches all comment histories and breaks github's PR continuity. We always use squash-merge so your changes will all go into one single PR at the end of the day.

packages/align-deps/src/commands/vigilant.ts
tido64: They're just there because the type definition requires it. They're actually unused.

PR #1931, opened 2022-10-13
docs(align-deps): update README

tido64:
### Description

Update README.

### Test plan

n/a


packages/align-deps/README.md
```diff
@@ -7,13 +7,18 @@
 
 <!--remove-block end-->
 
-`@rnx-kit/align-deps` manages React Native dependencies for a package, based on
-its needs and requirements.
+`@rnx-kit/align-deps` manages dependencies for an npm package, based on its
+needs and requirements.
```
afoxman, line 11: Consider expanding on this, explaining what `align-deps` does in some high-level detail, using words rather than code/bullets. A few sentences to contextualize it and introduce concepts like:

- aligning dependencies is important for compatibility
- built around preset, requirements, and capabilities (and briefly what these are and how they work together)
- ships with presets that capture real-world compatibility data, and is extensible so you can add your own
- you can use the command line, or get a richer experience through package configuration
- works well in monorepos and across them

afoxman: Then, below, when they read your examples and more technical info, they'll have a picture in mind of what it means and why it is the way it is.

afoxman (approved):

packages/align-deps/README.md
afoxman: Sounds ok to me. Maybe put it at the end of the intro paragraph?

Then also update that guide so it has the right terminology :)

PR #1941, opened 2022-10-18
feat(dep-check): add migration message

tido64:
### Description

Tell users to migrate to `@rnx-kit/align-deps`.

### Test plan

```
cd packages/dep-check
yarn build --dependencies
```


afoxman (approved):
What are your plans for the dep-check package going forward?

tido64:
> What are your plans for the dep-check package going forward?

I'm thinking we should remove it and add a notice on npm that this package has been deprecated. I will check on usage numbers some time next month (or after new years) before proceeding. We can probably already add the deprecation notice after this change has been published. What are your thoughts?

afoxman:
> > What are your plans for the dep-check package going forward?
> 
> I'm thinking we should remove it and add a notice on npm that this package has been deprecated. I will check on usage numbers some time next month (or after new years) before proceeding. We can probably already add the deprecation notice after this change has been published. What are your thoughts?

This makes sense. 

People will start seeing this warning message when they run it, and soon after, when you mark is deprecated on npm, they'll see warnings when they install it too. 

After that, I think you can delete the package from the repo. We won't service it anymore. `align-deps` was done very carefully so that it can support migrating users easily. I don't think we should try to continue maintaining `dep-check` at all. People can always stay on the last working copy of it until they are ready to upgrade.

Great work!

afoxman:
> After that, I think you can delete the package from the repo. We won't service it anymore. `align-deps` was done very carefully so that it can support migrating users easily. I don't think we should try to continue maintaining `dep-check` at all. People can always stay on the last working copy of it until they are ready to upgrade.

This may be too aggressive, actually. There are links out there which point to the `dep-check` package. So maybe just reduce it down to a mostly-empty placeholder package.json file and a README, since I suspect most links are to the root of the package or the README.

PR #1948, opened 2022-10-20
refactor(align-deps): migrate set-version command

tido64:
### Description

Migrates the `set-version` command.

### Test plan

As with the `init` command, the `set-version` is currently also only available for React Native packages.

- Verify that `align-deps` fails if the configuration is missing:
  ```
  cd packages/babel-preset-metro-react-native
  node ../align-deps/lib/cli.js --set-version 0.68
  ```
- Verify that `align-deps` fails if the configuration is incomplete:
  ```
  cd packages/metro-config
  node ../align-deps/lib/cli.js --set-version 0.68
  ```
- `align-deps` will migrate while updating requirements:
  ```
  cd packages/test-app
  node ../align-deps/lib/cli.js --set-version 0.68
  ```


packages/align-deps/src/helpers.ts
```diff
@@ -14,23 +15,41 @@ export function compare<T>(lhs: T, rhs: T): -1 | 0 | 1 {
...
+        .map((v) => {
+          if (v === "*" || v === "-") {
+            // No need to manipulate `*` or hyphen ranges, e.g. `1.0 - 2.0`
+            return v;
+          }
+
+          const [major, minor = "0"] = v.split(".");
+          return major === "^0" || major === "~0"
+            ? `0.${minor}`
+            : `${major}.${minor}`;
```
afoxman, line 42: Are there semver routines you can use instead of custom string parsing/handling logic here?

packages/align-deps/src/preset.ts
```diff
@@ -53,11 +53,19 @@ function loadPreset(preset: string, projectRoot: string): Preset {
   try {
     return require("./presets/" + preset).default;
   } catch (_) {
-    return require(require.resolve(preset, { paths: [projectRoot] }));
+    return process.env.JEST_WORKER_ID
```
afoxman, line 56: Adding special handling for Jest seems wrong. If this isn't here, is there some Jest config or some other organization of the preset files within the package that make Jest work?

packages/align-deps/src/preset.ts
```diff
@@ -53,11 +53,19 @@ function loadPreset(preset: string, projectRoot: string): Preset {
...
 }
 
-export function filterPreset(requirements: string[], preset: Preset): Preset {
+/**
+ * Filters out any profiles that do not satisfy the specified requirements.
+ * @param preset The preset to filter
+ * @param requirements The requirements that a profile must satisfy
+ * @returns Preset with only profiles that satisfy the requirements
+ */
+export function filterPreset(preset: Preset, requirements: string[]): Preset {
```
afoxman, line 68: There's a FIXME in this function. Is that tracked somewhere?

packages/align-deps/src/preset.ts
```diff
@@ -74,6 +82,12 @@ export function filterPreset(requirements: string[], preset: Preset): Preset {
   return filteredPreset;
 }
 
+/**
+ * Loads and merges specified presets.
```
afoxman, line 86: This is good to see - thanks for adding docs. Maybe expand on how collisions are handled and how that relates to ordering?

packages/align-deps/src/helpers.ts
afoxman: The `Range` class seems in semver may be able to do some of this, with utility functions being used on the specific components of the range.

packages/align-deps/src/commands/setVersion.ts
```diff
@@ -0,0 +1,173 @@
...
+  return { title: version, value: version };
+}
+
+async function parseInput(versions: string | number): Promise<{
+  supportedVersions: string[];
+  targetVersion: string;
+} | null> {
+  // When `--set-version` is without a value, `versions` is an empty string if
+  // invoked directly. When invoked via `@react-native-community/cli`,
+  // `versions` is `true` instead.
```
afoxman, line 45: If `versions` is an empty string or `true`, why is the type `string | number` instead of `string | boolean` (or just `string` if you meant "true" instead of `true`)?

afoxman: Any why not normalize at the call site to one convention or the other?

packages/align-deps/src/commands/setVersion.ts
```diff
@@ -0,0 +1,173 @@
...
+function updateRequirements(
+  { requirements }: AlignDepsConfig["alignDeps"],
+  prodVersion: string,
+  devVersion = prodVersion
+): void {
+  if (Array.isArray(requirements)) {
+    setRequirement(requirements, prodVersion);
+  } else {
+    setRequirement(requirements.production, prodVersion);
+    setRequirement(requirements.development, devVersion);
```
afoxman, line 98: Is this input validated by the caller? In `setRequirement`, if the property isn't set or is a non-array value, then the code will crash (which is good, but not as good as a nice message about a bad config/cmd-line).

packages/align-deps/src/commands/setVersion.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+}
+
+function setVersion(
+  config: AlignDepsConfig | LegacyCheckConfig,
+  targetVersion: string,
+  supportedVersions: string[]
+): ErrorCode | PackageManifest {
```
afoxman: `ErrorCode` isn't needed because nothing in here fails with an error. Maybe take it out until there's a place where a graceful failure needs to be handled?

packages/align-deps/src/commands/setVersion.ts
tido64: This input is guaranteed to be valid since it is returned from `loadConfig`.

packages/align-deps/src/preset.ts
tido64: It's tracked in https://github.com/microsoft/rnx-kit/issues/1890. I already have a fix ready, but all my changes are stacked on each other so I have to wait till the first one lands before the next one can be submitted for review.

packages/align-deps/src/helpers.ts
tido64: Unfortunately, I tried all of them and they are all lossy for some reason. I didn't want to implement custom logic for this either.

packages/align-deps/src/commands/setVersion.ts
tido64: Yeah, it's removed in the latest iteration. Sorry, I was still reviewing the changes after I had submitted the PR 😛

packages/align-deps/src/preset.ts
tido64: Jest doesn't allow you to mock `require.resolve`. At least I couldn't find a way to do it. It only mocks `resolve`. This is a bit of a workaround here because we do need to call the former to get the correct starting point for module resolution in monorepos, in particular the ones that use pnpm-like solutions. I could look into mocking `loadPreset` itself. If there's a solution that doesn't need mocking at all, I'm all ears.

afoxman (approved):

Issue #1945, opened 2022-10-19
rnx-kit should use `moduleSuffixes` when working against TS 4.7 or newer


afoxman:
The rnx-kit CLI commands `rnx-start` and `rnx-bundle` use a custom module resolver with TypeScript. The resolver replaces TS's default resolver. The custom resolver adds support for react-native platform file extensions (.ios.ts, .native.ts).

The custom resolver _does not_ support TypeScript's path-remapping features: see `baseUrl` and `paths`.

As a result, bundling fails in odd ways when encountering projects which use TS-remapping.

When working against TS 4.7 or newer, we should not run any custom resolver code. Instead, we should dynamically hook in and add/update `moduleSuffixes` for every tsconfig we encounter. The idea is to let the original TS resolver run, including support for path-remapping, while making sure RN file resolution works.

Keep in mind `moduleSuffixes` can be used for non-RN purposes, and updating it needs to be done carefully.

Original issue: #1900.

Issue #1944, opened 2022-10-19
rnx-kit CLI should fail when it encounters tsconfig with path-remapping (TS < 4.7)


afoxman:
The rnx-kit CLI commands `rnx-start` and `rnx-bundle` use a custom module resolver with TypeScript. The resolver replaces TS's default resolver. The custom resolver adds support for react-native platform file extensions (.ios.ts, .native.ts).

The custom resolver _does not_ support TypeScript's path-remapping features: see `baseUrl` and `paths`.

As a result, bundling fails in odd ways when encountering projects which use TS-remapping.

We should detect the `baseUrl` and `paths` props in tsconfig, and fail the rnx-kit CLI command with a specific message about these not being supported.

Only do this for TS versions older than 4.7. In a separate bug, rnx-kit will handle TS 4.7 and newer using `moduleSuffixes` instead of a custom resolver.

Original issue: #1900.

Issue #1910, opened 2022-09-29
Aliases not working


johnf:
### What happened?

In my react project I use aliases to import from ```src``` e.g.

```typescript
import { Section } from './src/Section';
// versus
import { Section } from '@demo/Section';
```

This doesn't seem to be supported by rnx-start, I get the following error

```
Reloading app...
App.tsx:29:31 - error TS2307: Cannot find module '@demo/Section' or its corresponding type declarations.

29 import {Section, styles} from '@demo/Section';
                                 ~~~~~~~~~~~~~~~
```

I'm guessing I probably need a slight config change but haven't been able to work it out.

I'm happy to provide a PR for documentation if someone can point me in the right direction.

### Affected Package

@rnx-kit/cli

### Version

0.14.8

### Which platforms are you seeing this issue on?

- [X] Android
- [X] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
info Fetching system and libraries information...
System:
    OS: Linux 5.15 Ubuntu 20.04.5 LTS (Focal Fossa)
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 10.86 GB / 30.99 GB
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.18 - ~/.nvm/versions/node/v16.15.0/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  SDKs:
    Android SDK:
      API Levels: 30, 31, 32, 33
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.2, 30.0.3, 31.0.0, 31.0.0, 32.1.0, 33.0.0, 33.0.0
      System Images: android-30 | Google APIs Intel x86 Atom, android-30 | Google Play Intel x86 Atom, android-31 | Google APIs Intel x86 Atom_64, android-33 | Google APIs Intel x86 Atom_64
      Android NDK: Not Found
  IDEs:
    Android Studio: AI-213.7172.25.2113.9014738
  Languages:
    Java: 11.0.16 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 18.1.0 => 18.1.0 
    react-native: 0.70.1 => 0.70.1 
[...]

PR #1956, opened 2022-10-22
Typescript alias support

johnf:
### Description

Switch to using Typescript name resolver when using TS > 4.7, now that it has moduleSuffix support.
For <4.7, error if ```baseUrl``` or ```paths``` is being used.

Resolves:
* #1944 
* #1945

### Test plan

See example repo https://github.com/johnf/rnk-kit-aliases-demo


packages/cli/src/typescript/project-cache.ts
```diff
@@ -114,6 +114,20 @@ export function createProjectCache(
...
   }
 
+  function checkForAliasSupport(options: ts.CompilerOptions) {
+    const { baseUrl, paths } = options;
+    if (!baseUrl && !paths) {
+      return;
+    }
+
+    const [major, minor] = ts.version.split(".").map((v) => parseInt(v, 10));
+    if (major < 4 || (major === 4 && minor < 7)) {
```
tido64, line 124: Should we use `semver` to parse the version number? For reference, we have this Metro version check elsewhere:
https://github.com/microsoft/rnx-kit/blob/19b7decc611951b3e3a87ccb130f7e552bfd0d9c/packages/metro-serializer/src/index.ts#L33

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -120,6 +120,29 @@ export function resolveModuleName(
 ): ts.ResolvedModuleFull | undefined {
   let module: ts.ResolvedModuleFull | undefined = undefined;
 
+  // TODO: Should we optimise by moving this check out of the function?
```
tido64, line 123: I would definitely create a separate function that just returns a boolean, e.g.:

```ts
function supportsModuleSuffixes(): boolean {
  return semver.satisfies(ts.version, ">=4.7.0");
}
```

Then use `supportsModuleSuffixes()` where needed.

packages/cli/src/typescript/project-cache.ts
tido64: Also, I think we should change the error message to say something like "@rnx-kit/cli requires TypeScript 4.7 or later for path mapping support."

Path mapping has existed for some time. We should be clear that it's just our custom resolver that has this limitation.

tido64 (commented):
Thanks for working on this ❤️

I left a few comments that I think will help improve maintainability.

packages/cli/src/typescript/project-cache.ts
afoxman: Also, in the error message, mention that `paths` and/or `baseUrl` fields were set, which is why this error was hit. This tells them the root of the problem so they can work around by either removing path-remapping or by turning off typescriptValidation.

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -120,6 +120,29 @@ export function resolveModuleName(
...
+  let typeScriptSupportsPaths = true;
+  const [major, minor] = ts.version.split(".").map((v) => parseInt(v, 10));
+  if (major < 4 || (major === 4 && minor < 7)) {
+    typeScriptSupportsPaths = false;
+  }
+
+  if (typeScriptSupportsPaths) {
+    const optionsWithSuffixes = {
+      ...context.options,
+      moduleSuffixes: [...context.platformExtensions, ""],
```
afoxman, line 133: I had some thoughts about replacing `moduleSuffixes`. Details are here: https://github.com/microsoft/rnx-kit/issues/1910#issuecomment-1289439344

packages/cli/src/typescript/project-cache.ts
johnf: I've switched to semver. Decidied to define the helper function in the resolve package and export it so cli can use it

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -120,6 +125,22 @@ export function resolveModuleName(
...
+    };
+
+    module = ts.resolveModuleName(
+      moduleName,
+      containingFile,
+      optionsWithSuffixes,
+      context.host
+    ).resolvedModule;
```
afoxman, line 139: I think some more work is needed here. Note the two API signatures below:

```ts
// Hooked function in the `host` object we pass to TS
function resolveModuleNames(
   moduleNames: string[], 
   containingFile: string, 
   reusedNames: string[] | undefined, 
   redirectedReference: ResolvedProjectReference | undefined, 
   options: CompilerOptions, 
   containingSourceFile?: SourceFile): 
(ResolvedModule | undefined)[];

// TS function we invoke to do the resolution work, after modifying TSConfig.
function resolveModuleName(
    moduleName: string, 
    containingFile: string, 
    compilerOptions: CompilerOptions, 
    host: ModuleResolutionHost, 
    cache?: ModuleResolutionCache, 
    redirectedReference?: ResolvedProjectReference, 
    resolutionMode?: ModuleKind.CommonJS | ModuleKind.ESNext): ResolvedModuleWithFailedLookupLocations;
```

With 4.7, we're calling ts.resolveModuleName, and should be passing through fields like `redirectedReference` and `options` (as `compilerOptions`, after modifying `moduleSuffixes`). 

We should also be creating a module resolution cache, just as TS does when it invokes its own internal loader:

```ts
moduleResolutionCache = createModuleResolutionCache(currentDirectory, getCanonicalFileName, options);
```

afoxman: This should all be in an "if 4.7" block to avoid the need for backward-compat testing.

packages/typescript-react-native-resolver/src/host.ts
johnf: Implemented as per last comments in https://github.com/microsoft/rnx-kit/issues/1910#issuecomment-1289439344

packages/cli/src/typescript/project-cache.ts
```diff
@@ -114,6 +114,19 @@ export function createProjectCache(
     return cmdLine;
   }
 
+  function checkForAliasSupport(options: ts.CompilerOptions) {
+    const { baseUrl, paths } = options;
```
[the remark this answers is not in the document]
johnf: Done

packages/typescript-react-native-resolver/src/host.ts
johnf: Added redirect4edReferences, options we were already pulling out of context

johnf:
@afoxman At the moment the ```host.ts``` tests are failing, because the tests are taking the new code path as typescript4.7 is installed.
I tried a couple of things but I couldn't work out how to mock ```supportsModuleSuffixes``` to return false.

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -116,10 +126,45 @@ export function resolveModuleName(
...
+    const optionsWithSuffixes = {
+      ...context.options,
+    };
+
+    if (!optionsWithSuffixes.moduleSuffixes) {
+      optionsWithSuffixes.moduleSuffixes = moduleSuffixes;
+    } else if (
+      !optionsWithSuffixes.moduleSuffixes
+        .join(",")
+        .includes(moduleSuffixes.join(","))
```
afoxman, line 146: There could be other suffixes in there, so you need to check for ours while ignoring the others. And you need to ensure ordering.

Maybe something like this:
```ts
const intersection = require("lodash/intersection");
const ok = intersection(optionsWithSuffixes.moduleSuffixes, moduleSuffixes) === moduleSuffixes;
if (!ok) { ... throw ... }
```

You'll need to add lodash to the dependency list, too.

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -116,10 +126,45 @@ export function resolveModuleName(
...
+    ) {
+      throw new Error(
+        `When using Typescript >=4.7, @rnx-kit/cli sets modulePrefixes.\n` +
+          `It does not currently support setting of modulePrefixes unless it exactly matches [".android", ".native", ""] or [".ios", ".native", ""].\n` +
+          `We would like to understand any use cases where you are setting modulePrefixes tosomething else.\n` +
+          `Please comment at https://github.com/microsoft/rnx-kit/discussions/1971`
```
afoxman, line 152:
```suggestion
        `Failed to resolve module reference '${moduleName}' in source file '${containingFile}'. The parent package has a TypeScript configuration which sets 'moduleSuffixes' to '${optionsWithSuffixes.moduleSuffixes.join(",")}'. This is incompatible with the target platform '${context.platform}', which requires moduleSuffixes to contain at least '${moduleSuffixes.join(",")}', in that order.\n` +
          `We would like to understand any use cases where you this error occurs, as there may be room to make improvements.\n` +
          `Please add a comment about your scenario, and include this error message: https://github.com/microsoft/rnx-kit/discussions/1971.`
```

packages/typescript-react-native-resolver/src/host.ts
afoxman: In this PR, we should stop loading the tsconfig entirely, and just use what is coming from TypeScript itself (the resolveModuleNames() call that we hooked).

This gets us around all the messy problems described here: https://github.com/microsoft/rnx-kit/issues/1910. Mainly, we don't know which tsconfig to load. Leaving it to TS keeps the door open for us to implement a loading strategy separately, likely as a TS enhancement.

afoxman: Also, you should create a module resolution cache for better performance. I missed it when I implemented this the first time around, but I think it will help. Or, file it as a separate TODO for a different PR and link it here.

afoxman:
> @afoxman At the moment the `host.ts` tests are failing, because the tests are taking the new code path as typescript4.7 is installed. I tried a couple of things but I couldn't work out how to mock `supportsModuleSuffixes` to return false.

Mocking global functions is a pain, but doable. In this case, can you just change ts.version (and then put it back)? It's a little dangerous, but probably ok for a single test.

If not, there are examples of mocking top-level functions in host.test.ts, near the top. `resolveFileModule` and `resolvePackageModule` are both mocked.

packages/typescript-react-native-resolver/src/host.ts
johnf: Do we care if they are also adjacent? Or do they just need to be there in the same order?
My code makes sure they exist as a block; yours allows other extensions to sit in the middle.

Which do we need?

afoxman: They only need to be in the same order, not adjacent.

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -77,13 +78,18 @@ export function changeHostToUseReactNativeResolver({
     );
   }
 
+  // Depending on where we get called from platformExtensionNames may already have platform in it
```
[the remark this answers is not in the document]
johnf: `changeHostToReactNativeResolver` gets called from two places
https://github.com/microsoft/rnx-kit/blob/f28cc77ace0640bd8e9b45f27fa440ed8d26e654/packages/cli/src/typescript/project-cache.ts#L142-L156

and 

https://github.com/microsoft/rnx-kit/blob/f28cc77ace0640bd8e9b45f27fa440ed8d26e654/packages/typescript-react-native-compiler/src/program.ts#L23-L38

In the first example `platformNameExtensions` is generated as
https://github.com/microsoft/rnx-kit/blob/f28cc77ace0640bd8e9b45f27fa440ed8d26e654/packages/cli/src/typescript/project-cache.ts#L143

which comes from 

https://github.com/microsoft/rnx-kit/blob/f28cc77ace0640bd8e9b45f27fa440ed8d26e654/packages/tools-react-native/src/platform.ts#L108-L116

In the second example, it is generated as
https://github.com/microsoft/rnx-kit/blob/f28cc77ace0640bd8e9b45f27fa440ed8d26e654/packages/typescript-react-native-compiler/src/program.ts#L13-L19

Tracing  back up a number of calls we see
https://github.com/microsoft/rnx-kit/blob/f28cc77ace0640bd8e9b45f27fa440ed8d26e654/packages/typescript-react-native-compiler/src/command-line/parse.ts#L38-L41

Tracing that up further we see it comes from `process.args`  with no defaults.

So to summarise that means
```
// First instance
const platformNames = ["android", "native"];
// Second instance
const platformNames = undefined
```

So in the original code in the first instance when we unshift we were getting
```javascript
const platformNames = ["android", "android", "native"]
```

johnf:
@afoxman OK two things left

> In this PR, we should stop loading the tsconfig entirely, and just use what is coming from TypeScript itself (the resolveModuleNames() call that we hooked).

I''m not sure what you mean here and how to approach?

> Also, you should create a module resolution cache for better performance. I missed it when I implemented this the first time around, but I think it will help. Or, file it as a separate TODO for a different PR and link it here.

I've added a fairly basic cache. I've assumed that if the module starts with a dot, that it is relative and cache against the name and the containing file, otherwise I cache against just the module name.

I'm not sure if there are any hidden dragons here though.

johnf:
@afoxman small nudge to review my last comment, understand f you're busy :)

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -146,6 +199,8 @@ export function resolveModuleName(
   return module;
 }
 
+const moduleResolutionCache: Record<string, ts.ResolvedModuleFull | undefined> = {};
```
afoxman, line 202: The module resolution cache should be created using `ts.createModuleResolutionCache`. 

I probably shouldn't have brought it into this PR. It's ok to just skip it.

You have to create one per directory, so it will need a parent map of path -> cache-instance. This would have to be created and attached to the larger context object, with some helper APIs to take a module path in and get a cache instance out (possibly creating it). And then you need tests to make sure the cache is working as expected without altering the course of resolving a module. This is the harder piece.

packages/typescript-react-native-resolver/src/host.ts
```diff
@@ -162,14 +217,25 @@ export function resolveModuleNames(
   moduleNames: string[],
   containingFile: string,
   _reusedNames: string[] | undefined,
-  _redirectedReference?: ts.ResolvedProjectReference
+  redirectedReference?: ts.ResolvedProjectReference
```
afoxman, line 220: This function is the entry-point we give to TS for module resolution. It turns out that TS is passing more args than I listed here, and we've been ignoring them:

```typescript
function resolveModuleNames(
   ...
   redirectedReference: ResolvedProjectReference | undefined, 
   options: CompilerOptions
```

There are even more than that, but the one we want is `options`, and it is passed right after `redirectedReference`.

Using the passed-in options lets us hand control of option-loading back to TS, and stop doing it ourselves. See https://github.com/microsoft/rnx-kit/issues/1910#issuecomment-1290775749 for details.

I'd like to do this for both pre-4.7 and post-4.7 to keep the implementation simple and because I don't think there's a risk. That means we can have one code-path for both.

So that amounts to removing `options` from the `ResolverContext` object (types.ts, line 24) and instead use what comes in with the resolution call. We'll need to manually pass it around through our own resolver code, since it could change from call to call.

After you remove `options` from `ResolverContext`, it will need to come out of `changeHostToUseReactNativeResolver`, which will cause a **Major** version change for `@rnx-kit/typescript-react-native-resolver`. And the callers will need to stop passing it in. I had hoped they could also stop loading it, but they can't.

packages/typescript-react-native-resolver/test/host.test.ts
```diff
@@ -18,6 +18,17 @@ import {
...
 
+ let origTsVersion: string;
+ beforeEach(() => {
+    origTsVersion = ts.version;
+    (ts as any).version = "4.6.0"; // eslint-disable-line @typescript-eslint/no-explicit-any
+  });
+
+  afterEach(() => {
+    (ts as any).version = origTsVersion; // eslint-disable-line @typescript-eslint/no-explicit-any
+    jest.resetAllMocks();
```
afoxman, line 29: I don't think you need to reset any mocks, since it wasn't happening before and you didn't add any new ones.

packages/typescript-react-native-resolver/test/host.test.ts
```diff
@@ -18,6 +18,17 @@ import {
   ExtensionsJSON,
 } from "../src/extension";
 
+ let origTsVersion: string;
+ beforeEach(() => {
+    origTsVersion = ts.version;
+    (ts as any).version = "4.6.0"; // eslint-disable-line @typescript-eslint/no-explicit-any
```
afoxman, line 24: Can you add a few tests where this is set to 4.7.0 to ensure that the new code runs?

packages/typescript-react-native-resolver/src/host.ts
johnf: OK, I'll drop this, as that's definitely going to push my limits of intimacy with the code base and how typescript works.

johnf:
@afoxman I think I might have taken this as far as I can.
With the changes around options, I'm not confident I can complete them without breaking anything.
I don't have the intimate knowledge of typescript to understand what I'm doing.

afoxman:
> @afoxman I think I might have taken this as far as I can. With the changes around options, I'm not confident I can complete them without breaking anything. I don't have the intimate knowledge of typescript to understand what I'm doing.

Thanks for all the work you did on this @johnf! I can pick it up from here and get it over the finish line. :)

afoxman:
While working to finish this, I found some problems that this PR exposes re: trace logging. The PR is now too large to make sense, and is changing several things at once. I will be splitting it up into smaller PRs and submitting them separately, then closing this once everything is in and validated.

afoxman:
This PR is now covered by 4 separate PRs: #2062, #2063, #2064, and #2065.

afoxman:
Thanks, again, @johnf! Your work here stirred up some good dust, in addition to bringing new functionality. We're now on a path to remove our custom module resolver, which is a step in the right direction for projects that combine TypeScript + React Native with our tools.

PR #1959, opened 2022-10-23
fix(align-deps): cache "vigilant" mode profiles

tido64:
### Description

In repos with 1000+ packages, runtime went from 2.2s -> 0.8s.

### Test plan

n/a


kelset (approved):
🚀

PR #1960, opened 2022-10-23
ci: migrate to `align-deps`

tido64:
### Description

Replaces dep-check in CI.

### Test plan

CI should pass.


kelset (approved):
good stuff - the sooner CI migrates to this, the better we can validate that everything still works as we want 👍

PR #1972, opened 2022-10-26
fix(align-deps): fix capability aliases being checked twice

tido64:
### Description

Dependencies may be checked twice because multiple capabilities can resolve to the same package.

### Test plan

CI should pass.


packages/align-deps/src/capabilities.ts
```diff
@@ -92,11 +92,16 @@ function resolveCapability(
...
+ * Resolves specified capabilities.
+ * @param capabilities The list of capabilities to resolve
+ * @param preset The preset to use to resolve capabilities
+ * @returns A tuple of resolved dependencies and unresolved capabilities
+ */
+export function resolveCapabilitiesUnchecked(
   capabilities: Capability[],
   preset: Preset
-): Record<string, Package[]> {
+): [Record<string, Package[]>, Record<string, string[]>] {
```
afoxman, line 104: For structured data like this, a dictionary make sure consumers don't read the output incorrectly. I'd recommend switching to something like:

```
{ dependencies: Record<string, Package[]>, unresolvedCapabilities: Record<string, string[]> }
```

And maybe even clarifying the Record<> types with real indexable types that have named keys/values.

afoxman (approved):

PR #1976, opened 2022-10-26
feat(cli): migrate to `align-deps`

tido64:
### Description

Replaces `dep-check` in `cli`.

### Test plan

```
# Build
cd packages/test-app
yarn build --dependencies
yarn react-native rnx-align-deps --help

# Try a few commands, for instance:
yarn react-native rnx-align-deps
yarn react-native rnx-align-deps --set-version
```


packages/cli/src/align-deps.ts
```diff
@@ -0,0 +1,72 @@
...
+    "migrate-config": Boolean(args.migrateConfig),
+    verbose: Boolean(args.verbose),
+    write: Boolean(args.write),
+    packages: argv,
+  });
+}
+
+export const rnxAlignDepsCommand = {
+  name: "rnx-align-deps",
+  description: "Dependency checker for npm packages",
```
afoxman, line 32: Is this still "dependency checker"? Or is it something different, like "Dependency manager" or something with the word "align" in it?

afoxman (approved):

packages/cli/src/align-deps.ts
tido64: How about `Manage dependencies within a repository and across many repositories`?

afoxman: Maybe update package.json, README, and the doc site with this text too, or some context-appropriate version of it. Something more than "checker".

PR #1980, opened 2022-10-27
fix(align-deps): app requirements must resolve to a single profile

tido64:
### Description

App requirements must resolve to a single profile

### Test plan

Modify test app's config:

```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index fc4be0f6..b33fdf7f 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -154,7 +154,7 @@
         "microsoft/react-native"
       ],
       "requirements": [
-        "react-native@0.68"
+        "react-native@0.68 || 0.70"
       ],
       "capabilities": [
         "core-android",
```

Then run:

```
cd packages/align-deps
yarn build --dependencies
cd ../test-app
node ../align-deps/lib/cli.js
```


packages/align-deps/src/commands/check.ts
```diff
@@ -41,12 +41,14 @@ export function checkPackageManifest(
     path.dirname(manifestPath),
     options
   );
-  if (capabilities.length === 0) {
+  const { kitType, manifest } = config;
+
+  if (kitType === "app" && Object.keys(prodPreset).length !== 1) {
```
afoxman, line 46: Why only for apps?

afoxman (approved):

packages/align-deps/src/commands/check.ts
tido64: Libraries must be able to support multiple versions. They cannot control which version apps are running.

PR #1987, opened 2022-11-01
fix(metro-serializer-esbuild): use esbuild only for prod

tido64:
### Description

Use esbuild only for production bundles.

### Test plan

Tests were added.


kelset (changes requested):
I reckon then we should clearly state in the readme that this will be for prod buils only then.

kelset (approved):

PR #2011, opened 2022-11-03
chore: bump react-native-test-app to 2.0.0

tido64:
### Description

Bump `react-native-test-app` to 2.0.0

### Test plan

| Android | iOS | macOS | Windows |
| :-: | :-: | :-: | :-: |
| Screenshot_1667477912 | Simulator Screen Shot - iPhone 13 - 2022-11-03 at 13 17 50 | Screenshot 2022-11-03 at 14 11 46 | TODO |


packages/test-app/package.json
```diff
@@ -151,7 +151,8 @@
     },
     "alignDeps": {
       "presets": [
-        "microsoft/react-native"
+        "microsoft/react-native",
+        "@rnx-kit/scripts/rnx-align-deps.js"
```
kelset, line 155: is this intentional? does that file contain a preset?

tido64: Yes, it was always supposed to be there. Not sure why it was missing.

kelset: could we just rename it to be more clearly a preset? ex `rnx-align-deps-preset.js`

tido64: It's both a preset and a script: https://github.com/microsoft/rnx-kit/blob/945251cc9fba4ed8009727cfaa514e6385619e82/scripts/rnx-align-deps.js#L80-L92

We can probably split it, but that's outside the scope of this PR.

kelset: yeah let's split them up pls

kelset (approved):

packages/test-app/package.json
tido64: In queue: https://github.com/tido64/rnx-kit/compare/tido/bump-rnta...tido64:rnx-kit:tido/split-align-deps-preset?expand=1 😉

PR #2013, opened 2022-11-03
chore: extract preset from `rnx-align-deps.js`

tido64:
### Description

Extract preset from `rnx-align-deps.js`

### Test plan

CI should pass.


kelset (approved):
thanks 🙏

PR #2022, opened 2022-11-08
fix(dep-check): update react-native-svg package for RN 0.69+

Aure77:
### Description

react-native-svg@^13.5.0 support react native 0.69+ new architecture build (Fabric).
See: https://github.com/software-mansion/react-native-svg/releases/tag/v13.0.0

### DIFF
--- a/packages/dep-check/src/presets/microsoft/profile-0.69.ts
+++ b/packages/dep-check/src/presets/microsoft/profile-0.69.ts
@@ -109,6 +109,10 @@ const profile: Profile = {
     name: "@react-native-async-storage/async-storage",
     version: "^1.17.7",
   },
+  svg: {
+    name: "react-native-svg",
+    version: "^13.5.0",
+  },
   "test-app": {
     name: "react-native-test-app",
     version: "^1.3.10",


kelset:
thanks for submitting, I just want @tido64 to chime in because I think that at least we'll have to duplicate this into the align-deps dependency.

tido64:
I've been following the progress of react-native-svg a bit. Backwards compatibility breaks often, which is worrying to me. I intended to wait a bit until the 13.x branch is more stable before checking how viable it is. We will need to upgrade internally to make sure we can recommend it.

@Aure77 is this blocking you?

tido64:
> But maybe a confusing part of `rnx-dep-check` is why higher semver version of a dependency is considered as "non-compliant" for a version of RN ?

semver-wise, a major version introduces breaking changes. If dep-check allowed greater versions, it could inadvertently recommend versions that break compatibility. To make things simpler, the versions that dep-check recommend are just the values that are set in the profiles.

> Would an option to skip higher dep versions could be helpful ? or `customProfiles` should address it ?

`customProfiles` is the solution here if you want to use other versions than what the built-ins recommend.

kelset:
( btw @aure77 just as an headsup, be aware that we are in the process of finalizing the next version of dep-check, align-deps: https://github.com/microsoft/rnx-kit/issues/1890 )

tido64:
> ( btw @Aure77 just as an headsup, be aware that we are in the process of finalizing the next version of dep-check, align-deps: #1890 )

FYI, this has happened. Over the next weeks, we're going to telling people to start migrating to `align-deps`. We will no longer be updating `dep-check`.

`react-native-svg` 13.6.0 also recently dropped, fixing compatibility issues with older versions. I'm inclined to recommend that we consider 13.6.0 starting with 0.70. You won't be able to get Fabric + autolinking working in 0.69 anyway. I'm currently trying to get people to test internally.

tido64:
We're bumping react-native-svg to 13.7.0 in the 0.71 profile: https://github.com/microsoft/rnx-kit/pull/2032

I will close this one now. Thanks for your contribution 😄

PR #2024, opened 2022-11-10
fix(react-native-auth): add TurboModule spec

tido64:
### Description

Adds TurboModule spec. Implementing TurboModule (including backwards compatiblity) will be left for a separate PR.

### Test plan

- [x] Test that this still works with 0.64 (or whatever minimum version we support)
  - Support goes back to at least 0.62: https://github.com/facebook/react-native/blob/0.62-stable/Libraries/TurboModule/TurboModuleRegistry.js
- [x] ~~Figure out why `WithDefault` breaks codegen~~
  - `WithDefault` is only supported on components. Modules do not have defaults.

```
cd packages/test-app
yarn build --dependencies
yarn ios
```

Tap on "Acquire Token".

### DIFF
--- a/packages/react-native-auth/src/NativeAuth.ts
+++ b/packages/react-native-auth/src/NativeAuth.ts
@@ -45,6 +45,9 @@ import type { AuthResult } from "./types";
  */
 type AccountType = string;
 
+/**
+ * Specification for the native auth module.
+ */
 export interface Spec extends TurboModule {
   acquireTokenWithResource(
     resource: string,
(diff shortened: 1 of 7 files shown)


packages/react-native-auth/src/NativeAuth.ts
```diff
@@ -0,0 +1,83 @@
+import type { TurboModule } from "react-native";
+import { TurboModuleRegistry } from "react-native";
+import type { AuthResult } from "./types";
+
+/**
+ * We re-declare `AccountType` here as a `string` because codegen currently does
+ * not understand string literal unions. The workaround is to use string enums,
+ * but that has a runtime cost and bloats the bundle size. The following:
```
kelset, line 8: do we know if there's any plans for the codegen to add this? have we asked/raised it somewhere like the new arch WG?

kelset (approved):
Looks good 👍

Tested locally, and it works ✅

:shipit:

PR #2027, opened 2022-11-10
Enable rnx-bundle config to specify dev/minify

acoates-ms:
### Description

Want to be able to specify separate bundle configs for ship vs debug, with different output paths and other configs

```json
  "rnx-kit": {
    "bundle": [{
      "id": "androidDev",
      "entryFile": "src/index.tsx",
      "detectCyclicDependencies": true,
      "detectDuplicateDependencies": true,
      "typescriptValidation": true,
      "targets": [
        "android"
      ],
      "treeShake": false,
      "outputBundle": "dist/dev/android/index.bundle",
      "dev": true,
    },
    {
      "id": "androidShip",
      "entryFile": "src/index.tsx",
      "detectCyclicDependencies": true,
      "detectDuplicateDependencies": true,
      "typescriptValidation": true,
      "targets": [
        "android"
      ],
      "treeShake": true,
      "outputBundle": "dist/prod/android/index.bundle",
      "dev": false,
    }],
```


packages/cli/test/bundle/metro.test.ts
```diff
@@ -38,38 +39,37 @@ describe("CLI > Bundle > Metro > metroBundle", () => {
...
   };
 
-  const dev = true;
-  const minify = true;
-
   it("does not use a custom serializer when all plugins are disabled", async () => {
     const metroConfig = await getDefaultConfig();
     expect(metroConfig.serializer.customSerializer).toBeNil();
-    await metroBundle(metroConfig, bundleConfigNoPlugins, dev, minify);
+    // TODO input changed .. minify will no longer be true
```
tido64, line 45: Are these TODOs supposed to be fixed? Can you attach an issue if so?

tido64 (approved):
LGTM. @afoxman should sign off before you merge though.

packages/cli/test/bundle/metro.test.ts
acoates-ms: No, that was some left-over comments for myself to verify test results.  - I removed the comments.

packages/cli/src/bundle.ts
```diff
@@ -23,7 +23,9 @@ export async function rnxBundle(
 
   const bundleConfigs = getCliPlatformBundleConfigs(
     cliOptions.id,
-    cliOptions.platform
+    cliOptions.platform,
+    cliOptions.dev,
+    cliOptions.minify
```
afoxman, line 28: `getCliPlatformBundleConfigs` should just read the config, including platform-specific sections. With this PR, that would include reading `dev` and `minify`. It shouldn't take in CLI command-line overrides. That should be done below in `applyBundleConfigOverrides`.

packages/cli/src/ram-bundle.ts
```diff
@@ -41,7 +41,9 @@ export async function rnxRamBundle(
 
   const bundleConfigs = getCliPlatformBundleConfigs(
     cliOptions.id,
-    cliOptions.platform
+    cliOptions.platform,
+    cliOptions.dev,
+    cliOptions.minify
```
afoxman, line 46: Same as my comment above for "normal" bundling. `dev` and `minify` should not be sent in as overrides here. They should be used as overrides in `applyBundleConfigOverrides`.

afoxman (commented):
I understand where this is coming from, wanting to control everything through config. I deliberately left `dev` and `minify` out of config, as they don't fit with the spirit of what config is supposed to be. eg. a package's built-in opinion of what makes sense for bundling -- a suggestion to developers.

A package should not have an opinion on whether it is bundled as dev or ship. This should be a function of the larger context, such as a CI run or a script command. The developer has to decide, for each context, what makes sense, with a default toward `dev` builds *only* because that is the current standard with `react-native bundle`. 

If we further cement `dev` and `minify` as defaultable, then devs won't make that decision. And that's something I want to avoid.

The alternative (not merging this PR) is to keep these on each command line, which seems reasonable on the surface. I'm open to discussing further if you feel strongly, but as a matter of convenience, I think this is a step too far.

tido64:
Coates and I discussed this offline before the holidays and concluded that we don't really need this. I'm going to close for now.

Issue #1752, opened 2022-07-18
metro-resolver-symlinks + pnpm: unable to resolve files in module


danilobuerger:
### What happened?

As a follow up to #1744 with the same repo at https://github.com/danilobuerger/repo_rnx-kit_1744 after installing a dependency some files from that module can't be resolved. I am using `intl-pluralrules` as an example, but this happens with other dependencies too:

```
error: Error: Unable to resolve module ./factory from /.../node_modules/.pnpm/intl-pluralrules@1.3.1/node_modules/intl-pluralrules/plural-rules.js: ./factory could not be found within the project or in these directories:
  node_modules/.pnpm/intl-pluralrules@1.3.1/node_modules
  node_modules/.pnpm/node_modules
  node_modules
  1 | 'use strict';
  2 |
> 3 | var getPluralRules = require('./factory');
    |                               ^
  4 | var PseudoNumberFormat = require('./pseudo-number-format');
  5 |
  6 | function _interopDefaultLegacy (e) { return e && typeof e === 'object' && 'default' in e ? e : { 'default': e }; }
    at ModuleResolver.resolveDependency (/.../node_modules/.pnpm/metro@0.67.0/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:158:15)
    at DependencyGraph.resolveDependency (/.../node_modules/.pnpm/metro@0.67.0/node_modules/metro/src/node-haste/DependencyGraph.js:231:43)
    at Object.resolve (/.../node_modules/.pnpm/metro@0.67.0/node_modules/metro/src/lib/transformHelpers.js:129:24)
    at resolve (/.../node_modules/.pnpm/metro@0.67.0/node_modules/metro/src/DeltaBundler/traverseDependencies.js:396:33)
    at /.../node_modules/.pnpm/metro@0.67.0/node_modules/metro/src/DeltaBundler/traverseDependencies.js:412:26
    at Array.reduce (<anonymous>)
    at resolveDependencies (/.../node_modules/.pnpm/metro@0.67.0/node_modules/metro/src/DeltaBundler/traverseDependencies.js:411:33)
    at processModule (/.../node_modules/.pnpm/metro@0.67.0/node_modules/metro/src/DeltaBundler/traverseDependencies.js:140:31)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
[...]

Issue #1469, opened 2022-05-06
Package subpath './package.json' is not defined by "exports" when using @rnx-kit/metro-resolver-symlinks


nutstick:
### What happened?

 I am getting errors from the metro server after trying to integrate with `@rnx-kit/metro-resolver-symlinks`

```
error: Error: Unable to resolve module telejson from /repro-rnx-metro-symlink-exports/packages/utils/src/index.js: telejson could not be found within the project or in these directories:
  ../packages/utils/node_modules
  ../node_modules
  ../../../node_modules
```

After some debugging, I found that the package `telejson` uses `exports` fields in package.json

```
{
  "name": "telejson",
  "version": "5.3.3",
  ...
  "exports": {
    "import": "./dist/esm/index.js",
    "require": "./dist/cjs/index.js"
  },
}
```

And since @rnx-kit/metro-resolver-symlinks using `require.resolve('${pkgName}/package.json')` https://github.com/microsoft/rnx-kit/blob/master/packages/metro-resolver-symlinks/src/resolver.ts#L60. It unable to resolve to correct package

This is the error throw from `resolveFrom`
```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /Users/nuttapat.kirawittaya/Projects/repro-rnx-metro-symlink-exports/packages/utils/node_modules/telejson/package.json
    at new NodeError (internal/errors.js:322:7)
    at throwExportsNotFound (internal/modules/esm/resolve.js:332:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:565:3)
    at resolveExports (internal/modules/cjs/loader.js:450:36)
    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)
    at Function.resolve (internal/modules/cjs/helpers.js:107:19)
    at resolveFrom (/Users/nuttapat.kirawittaya/Projects/repro-rnx-metro-symlink-exports/node_modules/.pnpm/@rnx-kit+metro-resolver-symlinks@0.1.16/node_modules/@rnx-kit/metro-resolver-symlinks/lib/resolver.js:27:20)
[...]

Issue #1227, opened 2022-03-14
Package path not exported error - error when running react-native rnx-start


simarmol:
### What happened?

Trying to set up an existing project to work with rnx-kit. I am getting errors from the metro server when I run `yarn react-native rnx-start --reset-cache`. I have no problem whatsoever when I run `yarn react-native start --reset-cache`.

The errors look like this:

```
error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in /REPO//node_modules/node-html-parser/package.json
    at new NodeError (node:internal/errors:371:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:453:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:729:3)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at /REPO/node_modules/@rnx-kit/tools-react-native/lib/platform.js:77:46
    at Array.forEach (<anonymous>)
    at getAvailablePlatformsUncached (/repo/node_modules/@rnx-kit/tools-react-native/lib/platform.js:76:7)
```

I'm currently getting it for 'node-html-parser'. I was also getting the same error for a local package devDependency before, which is only configured to compile an ESM module. I removed it temporarily but immediately ran into the same error in node-html-parser. 

I'm assuming it's an esm vs. Common module thing? Which would make sense, except that the normal 'react-native start' command is working fine.

I'm also getting this warning about the blocklist, but I'm unsure if it is related.
```
● Deprecation Warning:

Warning: Metro config option `blacklistRE` is deprecated.
         Please use `blockList` instead.

● Validation Warning:

  Unknown option "blockList" with value 
  ```

I'd like some help getting it working for rnx-start if possible.  I'm not sure if I missed a config option somewhere, but I've tried various minor permutations of configs to no avail.

After looking around existing issues, I think this may be related to #805 or #1046.

### Affected Package

@rnx-kit/metro-config

### Version

 1.2.29

### Which platforms are you seeing this issue on?

- [X] Android
- [X] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
info Fetching system and libraries information...
System:
    OS: macOS 12.2.1
[...]

PR #2029, opened 2022-11-10
Allow rnx bundle config to add custom serialization plugins

acoates-ms:
### Description

Want to be able to specify custom serialization plugins in the rnx bundle config:

```json
  "rnx-kit": {
    "bundle": {
      "entryFile": "src/index.tsx",
      "customSerializationPlugins": [{"path": "<moduleNameOrPathToCustomPlugin>"}],
      "detectCyclicDependencies": true,
      "detectDuplicateDependencies": true,
      "typescriptValidation": true,
      "targets": [
        "android"
      ],
      "treeShake": false,
      "outputBundle": "dist/dev/android/index.bundle",
      "dev": true,
    },
  
```

This will add the additional plugins to the list of plugins passed into the serializer.

### DIFF
--- a/packages/cli/src/metro-config.ts
+++ b/packages/cli/src/metro-config.ts
@@ -118,6 +118,7 @@ const emptySerializerHook = (_graph: Graph, _delta: DeltaResult): void => {
 export function customizeMetroConfig(
   metroConfigReadonly: InputConfigT,
   {
+    customSerializationPlugins,
     detectCyclicDependencies,
     detectDuplicateDependencies,
     treeShake,
@@ -130,6 +131,13 @@ export function customizeMetroConfig(
   const metroConfig = metroConfigReadonly as InputConfigT;
 
   const plugins: MetroPlugin[] = [];
+  if (customSerializationPlugins) {
+    customSerializationPlugins.forEach((plugin) => {
+      const implPath = require.resolve(plugin.path, { paths: [process.cwd()] });
+      const impl = require(implPath);
+      plugins.push(impl(plugin.options));
+    });
+  }
   if (typeof detectDuplicateDependencies === "object") {
     plugins.push(DuplicateDependencies(detectDuplicateDependencies));
   } else if (detectDuplicateDependencies !== false) {
--- a/packages/cli/src/metro-config.ts
+++ b/packages/cli/src/metro-config.ts
@@ -131,16 +131,14 @@ export function customizeMetroConfig(
   const metroConfig = metroConfigReadonly as InputConfigT;
 
   const plugins: MetroPlugin[] = [];
-  if (extraSerializationPlugins) {
-    if (Array.isArray(extraSerializationPlugins))
-      extraSerializationPlugins.forEach((plugin) => {
-        const implPath = require.resolve(plugin.path, {
-          paths: [metroConfig.projectRoot],
-        });
-        const impl = require(implPath);
-        plugins.push(impl(plugin.options));
+  if (Array.isArray(extraSerializationPlugins))
+    extraSerializationPlugins.forEach((plugin) => {
+      const implPath = require.resolve(plugin.path, {
+        paths: [metroConfig.projectRoot],
       });
-  }
+      const impl = require(implPath);
+      plugins.push(impl(plugin.options));
+    });
   if (typeof detectDuplicateDependencies === "object") {
     plugins.push(DuplicateDependencies(detectDuplicateDependencies));
   } else if (detectDuplicateDependencies !== false) {
--- a/packages/cli/src/metro-config.ts
+++ b/packages/cli/src/metro-config.ts
@@ -118,7 +118,7 @@ const emptySerializerHook = (_graph: Graph, _delta: DeltaResult): void => {
 export function customizeMetroConfig(
   metroConfigReadonly: InputConfigT,
   {
-    customSerializationPlugins,
+    extraSerializationPlugins,
     detectCyclicDependencies,
     detectDuplicateDependencies,
     treeShake,
@@ -131,12 +131,15 @@ export function customizeMetroConfig(
   const metroConfig = metroConfigReadonly as InputConfigT;
 
   const plugins: MetroPlugin[] = [];
-  if (customSerializationPlugins) {
-    customSerializationPlugins.forEach((plugin) => {
-      const implPath = require.resolve(plugin.path, { paths: [process.cwd()] });
-      const impl = require(implPath);
-      plugins.push(impl(plugin.options));
-    });
+  if (extraSerializationPlugins) {
+    if (Array.isArray(extraSerializationPlugins))
+      extraSerializationPlugins.forEach((plugin) => {
+        const implPath = require.resolve(plugin.path, {
+          paths: [metroConfig.projectRoot],
+        });
+        const impl = require(implPath);
+        plugins.push(impl(plugin.options));
+      });
   }
   if (typeof detectDuplicateDependencies === "object") {
     plugins.push(DuplicateDependencies(detectDuplicateDependencies));
--- a/packages/config/src/bundleConfig.ts
+++ b/packages/config/src/bundleConfig.ts
@@ -10,7 +10,7 @@ export type TypeScriptValidationOptions = {
   throwOnError?: boolean;
 };
 
-export type CustomSerializationPluginOption = {
+export type ExtraSerializationPluginOption = {
   path: string;
   options?: object;
 };
@@ -22,7 +22,7 @@ export type BundlerPlugins = {
   /**
    * Specify additional custom metro serializer plugins
    */
-  customSerializationPlugins?: CustomSerializationPluginOption[];
+  extraSerializationPlugins?: ExtraSerializationPluginOption[];
 
   /**
    * Choose whether to detect cycles in the dependency graph. `true` uses defaults,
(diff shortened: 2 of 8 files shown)


packages/cli/src/metro-config.ts
```diff
@@ -130,6 +131,13 @@ export function customizeMetroConfig(
   const metroConfig = metroConfigReadonly as InputConfigT;
 
   const plugins: MetroPlugin[] = [];
+  if (customSerializationPlugins) {
```
tido64, line 134: We should ensure we have an array here:
```suggestion
  if (Array.isArray(customSerializationPlugins)) {
```

packages/config/src/bundleConfig.ts
```diff
@@ -10,10 +10,20 @@ export type TypeScriptValidationOptions = {
   throwOnError?: boolean;
 };
 
+export type CustomSerializationPluginOption = {
+  path: string;
+  options?: object;
+};
```
tido64, line 16: Should we match how Babel, ESLint, Jest and others specify plugins?
```suggestion
export type CustomSerializationPluginOption =
  | string
  | [string, Record<string, unknown>];
```

packages/config/src/bundleConfig.ts
```diff
@@ -10,10 +10,20 @@ export type TypeScriptValidationOptions = {
...
+
 /**
  * Parameters controlling bundler plugins.
  */
 export type BundlerPlugins = {
+  /**
+   * Specify additional custom metro serializer plugins
+   */
+  customSerializationPlugins?: CustomSerializationPluginOption[];
```
tido64, line 25: Can we rename this to make it clear that it won't replace the built-in plugins? E.g.:

```suggestion
  extraSerializationPlugins?: ExtraSerializationPluginOption[];
```

packages/cli/src/metro-config.ts
```diff
@@ -130,6 +131,13 @@ export function customizeMetroConfig(
...
+    customSerializationPlugins.forEach((plugin) => {
+      const implPath = require.resolve(plugin.path, { paths: [process.cwd()] });
```
tido64, line 136: Should this be relative to project root?
```suggestion
      const implPath = require.resolve(plugin.path, { paths: [metroConfig.projectRoot] });
```

packages/cli/src/metro-config.ts
tido64: Btw, you don't need the extra if check, `Array.isArray` accepts `null` and `undefined`.

packages/cli/src/metro-config.ts
```diff
@@ -130,6 +131,14 @@ export function customizeMetroConfig(
   const metroConfig = metroConfigReadonly as InputConfigT;
 
   const plugins: MetroPlugin[] = [];
+  if (Array.isArray(extraSerializationPlugins))
+    extraSerializationPlugins.forEach((plugin) => {
+      const implPath = require.resolve(plugin.path, {
+        paths: [metroConfig.projectRoot],
+      });
+      const impl = require(implPath);
```
afoxman, line 139: Dynamically resolving and requiring packages doesn't work with ES modules. We've faced this bug several times, across different packages. At the minimum, this needs to be wrapped in a try/catch with graceful error handling. Preferably, with a specific case for trapping and reporting on ES modules, as users will inadvertently try to use them and get confused when they don't work with cryptic errors from node.

packages/config/src/bundleConfig.ts
```diff
@@ -10,10 +10,20 @@ export type TypeScriptValidationOptions = {
...
+
 /**
  * Parameters controlling bundler plugins.
  */
 export type BundlerPlugins = {
+  /**
+   * Specify additional custom metro serializer plugins
+   */
+  extraSerializationPlugins?: ExtraSerializationPluginOption[];
+
```
afoxman, line 26: The name includes "serialization" which users won't understand if they're unfamiliar with our serializer plugin package. Same with "extra", as the dup/cyclic plugins aren't touted as plugins in config -- they're just options top the end-user.

Maybe call this `plugins` and update the comment to be more explicit about how they are applied, including a reference to our Metro Serializer Plugin package. This gives the user more context.

Ergonomically, plugins tend to have options. Since this is an open-ended way to have any plugin run, you'll need room for options here. Maybe with a unique type for each array entry that is the union of either a plain string (no options) or a dictionary with keys for `path` and `options`.

afoxman (commented):
This PR should include a path to rationalize the two existing plugins - dupes and cyclic dependencies - with this new generic mechanism.

Maybe deprecate both of those config options, with a message that users should name the individual @rnx-kit packages instead, and that the flags will go away in a future version. Include an example in the output - something they can copy/paste.

There might be more we can do here. @tido64 any opinions/ideas after just going through the dep-check to align-deps upgrade?

packages/cli/src/metro-config.ts
acoates-ms: Do we have an example where we do this already?

acoates-ms:
> This PR should include a path to rationalize the two existing plugins - dupes and cyclic dependencies - with this new generic mechanism.
> 
> Maybe deprecate both of those config options, with a message that users should name the individual @rnx-kit packages instead, and that the flags will go away in a future version. Include an example in the output - something they can copy/paste.
> 
> There might be more we can do here. @tido64 any opinions/ideas after just going through the dep-check to align-deps upgrade?

The current settings allow an easy way to have default plugins, and yet be able to remove them.  The duplication detection and cycle detection plugins are both in the config by default.  If we moved those settings into a generic plugin array, it is not clear how you would easily be able to specify in the config that a plugin should be removed from the list.

packages/cli/src/metro-config.ts
tido64: Both `@react-native-community/cli` and this package are transpiled to CJS. Changing this package to be ESM compatible requires using `import()`, which is async, which in turn means that this function and its callees (and its callees and so on) need to become async. Adding support for ESM now would require a bigger refactoring. Is that something that we need to support now? It seems out of scope for this PR.

afoxman: Yeah, that stack trace is pretty good, as is the error message. 

When devs hit this, the ones we faced in the past were all require(package.json). The stacks and messages that showed up were more ... confusing, from the devs point of view.

Seeing the example trace from tido64, I'm ok with letting that just happen. I think even new-ish users will figure it out and not file bugs against rnx-kit. That was what I was trying to avoid.

For completeness, we fixed the issues by reading the file contents and running them through JSON.parse(). This is why you won't find an example try/catch with ESM-error-detection. Here are two of them:

https://github.com/microsoft/rnx-kit/issues/1227
https://github.com/microsoft/rnx-kit/issues/1469

And this other issues is a related example of how an ESM include can be hard to decipher. This one is coming from Metro, so its console output has extra stuff in it. The stack trace is what to look at:

https://github.com/microsoft/rnx-kit/issues/1752

afoxman:
> > This PR should include a path to rationalize the two existing plugins - dupes and cyclic dependencies - with this new generic mechanism.
> > Maybe deprecate both of those config options, with a message that users should name the individual @rnx-kit packages instead, and that the flags will go away in a future version. Include an example in the output - something they can copy/paste.
> > There might be more we can do here. @tido64 any opinions/ideas after just going through the dep-check to align-deps upgrade?
> 
> The current settings allow an easy way to have default plugins, and yet be able to remove them. The duplication detection and cycle detection plugins are both in the config by default. If we moved those settings into a generic plugin array, it is not clear how you would easily be able to specify in the config that a plugin should be removed from the list.

I'm saying taking away the special status of dupes/cycles, and start with no plugins. If you want them, add them in. They're on by default now, through the config options, so that would be changing, but it would be done in a way that migrates users via a deprecation message followed later by removing support for the dupes/cycles config options (and probably failing if we see them).

packages/cli/src/metro-config.ts
tido64: These are related to export maps, not so much requiring ESM from CJS. We will have an issue with them regardless if we try to import anything that wasn't explicitly exported.

tido64:
> > > This PR should include a path to rationalize the two existing plugins - dupes and cyclic dependencies - with this new generic mechanism.
> > > Maybe deprecate both of those config options, with a message that users should name the individual @rnx-kit packages instead, and that the flags will go away in a future version. Include an example in the output - something they can copy/paste.
> > > There might be more we can do here. @tido64 any opinions/ideas after just going through the dep-check to align-deps upgrade?
> > 
> > 
> > The current settings allow an easy way to have default plugins, and yet be able to remove them. The duplication detection and cycle detection plugins are both in the config by default. If we moved those settings into a generic plugin array, it is not clear how you would easily be able to specify in the config that a plugin should be removed from the list.
> 
> I'm saying take away the special status of dupes/cycles, and start with no plugins. If you want them, add them in. They're on by default now, through the config options, so that would be changing, but it would be done in a way that migrates users via a deprecation message followed later by removing support for the dupes/cycles config options (and probably failing if we see them).
> 
> This can be a TODO for later. As long as there's a plan for collapsing away the special plugins into the general bucket.

I'd be fine with taking away their special status, but if the `plugins` field is empty, they should be enabled by default as usual. But if we're doing this, what about type validation? It doesn't get installed via the serializer, but via a hook. Should that also be a plugin?

tido64:
@acoates-ms, @afoxman: I posted an alternative implementation of this. Let me know what you think.

tido64:
Superseded by #2084

PR #2045, opened 2022-11-22
chore(website): remove local blog in favour of the DevBlog

kelset:
### Description

hey there - as mentioned a few times over the last few weeks, now that we have the [official devblog up and running](https://devblogs.microsoft.com/react-native/) we will move all the blogpost content over there - the blogposts from here have already been migrated over.

This PR takes care of redirecting the links to it; for the single posts, it's using the only alternative we could find: using the [HTTP redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#html_redirections) via [Head Metadata](https://docusaurus.io/docs/markdown-features/head-metadata).

### Test plan

I'm not sure if we deploy website previews, but if not, just clone this branch locally, go in the `docsite` folder and do yarn then yarn start and check out that the Blog entry in the main top navbar sends you to the DevBlog. You can also try navigating to the blogpost url and see that it redirects to the new one.


tido64:
Can we use this plugin? https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects/

afoxman:
> Can we use this plugin? https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects/

We can but it does client-side redirects, which I was hoping to avoid.

kelset:
> Can we use this plugin? [docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects/)

I already tried using this, and it doesn't accept external urls; it only allows "internal" redirects (ex. from "/blog/" to "/new-blog" and not full urls)

afoxman:
> > Can we use this plugin? [docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects/)
> 
> I already tried using this, and it doesn't accept external urls; it only allows "internal" redirects (ex. from "/blog/" to "/new-blog" and not full urls)

We can control tags that show in <head> using this: https://docusaurus.io/docs/markdown-features/head-metadata

Not sure if it overwrites or augments. The latter is what we need.

kelset:
thanks for the link @afoxman! That helped, I had to re-add the files in but at least now there's a redirect working:

screengrab redirect

it still flashes for a second but it works as far as I can tell.

tido64 (approved):

PR #2046, opened 2022-11-23
refactor(cli): extract TypeScript plugin to separate package

### LINKED ISSUE #1360: CLI: Extract TypeScript plugin so it can be used separately from CLI
When we introduced the type checking feature in #392, we embedded it directly into the CLI. It should be extracted to a separate package so it can be used without pulling in CLI.

tido64:
### Description

Extract TypeScript plugin to separate package

Resolves #1360.

### Test plan

CI should pass.

To verify that bundling catches errors, first build everything:

```
cd packages/test-app
yarn build --dependencies
```

Make the following changes:

```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index 9a744fe5..1fb35e1f 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -84,7 +84,9 @@
             "react-is"
           ]
         },
-        "typescriptValidation": true,
+        "typescriptValidation": {
+          "throwOnError": true
+        },
         "treeShake": false,
         "targets": [
           "android",
diff --git a/packages/test-app/src/App.native.tsx b/packages/test-app/src/App.native.tsx
index d4533e7d..291d97b0 100644
--- a/packages/test-app/src/App.native.tsx
+++ b/packages/test-app/src/App.native.tsx
@@ -15,7 +15,6 @@ import {
 // @ts-expect-error no types for "react-native/Libraries/Core/ReactNativeVersion"
 import { version as coreVersion } from "react-native/Libraries/Core/ReactNativeVersion";
 import { Colors, Header } from "react-native/Libraries/NewAppScreen";
-// @ts-expect-error no types for "react-native/Libraries/Utilities/DebugEnvironment"
 import { isAsyncDebugging } from "react-native/Libraries/Utilities/DebugEnvironment";
 import manifest from "../app.json";

```

Then bundle:

```
yarn bundle:ios
```


packages/metro-plugin-typescript/README.md
```diff
@@ -0,0 +1,29 @@
+<!--remove-block start-->
```
afoxman, line 1: Can you also add an entry for this file under docsite? Or is that all automatic now?

packages/metro-plugin-typescript/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { createProjectCache } from "./projectCache";
```
afoxman, line 1: Is this needed in the public API? Looks like the CLI only uses the plugin, and the project cache is managed internally.

afoxman (approved):
Looks good to me!

tido64:
@afoxman, before I merge this, do we want to land #1956 first? I want to avoid causing more friction for community contributions.

afoxman:
> @afoxman, before I merge this, do we want to land #1956 first? I want to avoid causing more friction for community contributions.

I was wondering about that... I have almost no experience with rebasing, but I think that is what we'd need to do for that other PR if this went first. Do you think it would be easy/painless? Not sure how much file overlap there is.

Either way, someone has to face the pain. If you're ok waiting and dealing with it, I'll try to get closure on the other PR soon so we can get it merged.

afoxman:
> Please do. I'm fine with rebasing this PR.

Ok. I'm working in finalizing the other PR today/tomorrow. Keep an eye out for it.

PR #2053, opened 2022-11-29
chore: don't install `@vue/compiler-sfc`

tido64:
### Description

`depcheck` doesn't use it because we don't use Vue. It's currently triggering a security alert because it pulls in an older version of `loader-utils`.

### Test plan

CI runs `depcheck` and should pass.


kelset (approved):
IIIINTERESTING APPROACH!

I don't know how you figured out this trick but damn 👏

:shipit:

tido64:
> IIIINTERESTING APPROACH!
> 
> I don't know how you figured out this trick but damn 👏
> 
> :shipit:

I totally didn't find this on a site called StackOverflow 😛

PR #2057, opened 2022-12-01
Update README.md for npx

pvinis:
### Description

adding some instructions to run the command with just npx

### DIFF
--- a/packages/align-deps/README.md
+++ b/packages/align-deps/README.md
@@ -45,6 +45,12 @@ npm add --save-dev @rnx-kit/align-deps
 yarn rnx-align-deps [options] [packages...]
 ```
 
+or just
+
+```sh
+npx @rnx-kit/align-deps [options] [packages...]
+```
+
 Listing paths to packages that should be checked is optional. If omitted,
 `align-deps` will look for the closest `package.json` using Node module
 resolution. If the target package is a root package defining workspaces, they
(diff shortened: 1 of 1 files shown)


kelset:
LGTM

kelset (changes requested):
Thanks for adding this Pavlos!

I was talking with Tommy and we think it might be better to add a bit more text around the npx variant, something like "if you want to take it for a test run, you can even use it via npx, but please for the love of Thor don't use npx commands on CI"

pvinis:
interesting! I can add some more. mostly I wanted to do the PR and see if it makes sense, which it did to me.

also, I thought this is kind of a locally ran command 🤔. I wouldn't use it in ci.

also also, did I miss the memo? what's wrong with npx on ci? 😬😱

tido64:
> also also, did I miss the memo? what's wrong with npx on ci? 😬😱

There's nothing inherently wrong with `npx`. With `align-deps`, the issue you can run into is that we update some dependencies, and your CI randomly breaks because now you're out of sync. That's why we recommend you lock to a specific version, and bump only when you're ready.

kelset:
> also also, did I miss the memo? what's wrong with npx on ci? 😬😱

To make an extreme example: let's say you set up your CI with an npx command for library X, `npx X-command`. You then go focus on something else for 6 months. Author of library X in the meantime releases X 2.0, in which there's a breaking change in how `X-command` works: your CI will go instantly kaboom the moment this new version is released.

tido64:
@pvinis: I've updated the README with instructions for npx. Thanks for pointing it out! Let us know if there's anything else we're missing.

pvinis:
awesome! thank you tido

PR #2061, opened 2022-12-02
Add support for react-native-windows to jest-preset

acoates-ms:
### Description
jest-preset should apply transforms to the react-native-windows packages to support react-native-windows out of the box.

### DIFF
--- a/packages/jest-preset/test/index.test.ts
+++ b/packages/jest-preset/test/index.test.ts
@@ -239,9 +239,10 @@ describe("jest-preset", () => {
       "node_modules/@office-iss/rex-win32/rex-win32.js": true,
       "node_modules/@react-native-community/cli/index.js": false,
       "node_modules/@react-native-windows/virtualized-list/src/VirtualizedList.js":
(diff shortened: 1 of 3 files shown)


packages/jest-preset/test/index.test.ts
```diff
@@ -239,9 +239,10 @@ describe("jest-preset", () => {
       "node_modules/@office-iss/rex-win32/rex-win32.js": true,
       "node_modules/@react-native-community/cli/index.js": false,
       "node_modules/@react-native-windows/virtualized-list/src/VirtualizedList.js":
-        true,
```
acoates-ms, line 242: I'm curious if anyone knows why this was explicitly called out as not being transformed. -- I confirmed locally that the @react-native-windows projects within react-native-windows do indeed require transformation to run.

tido64: I think we just missed this. It should be transformed like you said.

tido64 (approved):

PR #2062, opened 2022-12-03
Revert specialized trace logging in our React Native TypeScript resolver

afoxman:
### Description

This PR removes the specialized trace logging that exists in our custom TypeScript module resolver. It reverts our resolver to using TypeScript's trace() function for logging, which writes to stdout.

This aligns the logging behavior of our resolver and TypeScript, which is necessary for future work in this area (e.g. moduleSuffixes, TS 4.7).

**This is a BREAKING change to our TypeScript resolver and our "rn-tsc" TypeScript compiler wrapper**. 

Our CLI is affected, but not in any visible way to other developers. The CLI uses our TypeScript resolver, but never made use of the trace logging facilities, so the CLI update is code-cleanup only.

### Test plan

All changes have tests which verify them.


afoxman:
This PR was spun out of work that started in #1956.

tido64 (approved):

PR #2063, opened 2022-12-03
Move behavior from TS resolver to TS compiler wrapper

afoxman:
### Description

This PR moves a behavior out of `@rnx-kit/typescript-react-native-resolver` and into `@rnx-kit/typescript-react-native-compiler`.

The specific behavior being moved is the prepending of the target platform to the list of platform file extensions. This is something callers are responsible for doing.

In our repo, there are two callers that use the resolver -- the CLI and the `rn-tsc` compiler wrapper. The CLI was already doing this, but `rn-tsc` was not. This change updates `rn-tsc`, effectively "moving" the behavior into it.

### Test plan

All tests and CI loops pass.


afoxman:

PR #2064, opened 2022-12-03
Use compiler options from TS resolver callbacks, rather than top-level TS project

afoxman:
### Description

This PR changes our React Native TypeScript module resolver. The resolver used to accept TS compiler options from the top-level package being bundled, and use them when resolving all modules -- even those outside of the top-level package.

The compiler options can change for each module. TypeScript already handles this, and passes in an options object for each module being resolved. Our resolver will now use this options object.

**This is a breaking change for `@rnx-kit/typescript-react-native-resolver`. The public API has changed.**

The bulk of the code changes are plumbing. Passing around an options object, rather than getting it from a global context.

### Test plan

All tests and CI loops pass.


afoxman:

PR #2065, opened 2022-12-04
Delegate RN module resolution directly to TypeScript (4.7+ only), rather than our own resolver

### LINKED ISSUE #1944: rnx-kit CLI should fail when it encounters tsconfig with path-remapping (TS < 4.7)

### LINKED ISSUE #1945: rnx-kit should use `moduleSuffixes` when working against TS 4.7 or newer

afoxman:
### Description

This PR updates the way we resolve modules (import/require) to files (.js/.ts) for React Native projects.

For developers using TypeScript 4.7 or later, we delegate React Native module resolution to TypeScript, rather than running out own custom resolver. This is possible using the new `moduleSuffixes` compiler option. Using the TypeScript resolver is best because it supports many more features and is kept up-to-date with the latest changes in the Node ecosystem.

Our custom resolver will still run for developers on older versions of TypeScript.

Resolves #1944.
Resolves #1945.

### Test plan

This PR updates existing tests, and adds new ones, to cover both pre-4.7 and post-4.7 scenarios. This includes unit tests and end-to-end (real-world) tests.


afoxman:

PR #2071, opened 2022-12-07
Delegate RN module resolution directly to TypeScript (4.7+ only), rather than our own resolver

### LINKED ISSUE #1944: rnx-kit CLI should fail when it encounters tsconfig with path-remapping (TS < 4.7)

### LINKED ISSUE #1945: rnx-kit should use `moduleSuffixes` when working against TS 4.7 or newer

afoxman:
### Description

This PR updates the way we resolve modules (import/require) to files (.js/.ts) for React Native projects. 

For developers using TypeScript 4.7 or later, we delegate React Native module resolution to TypeScript, rather than running out own custom resolver. This is possible using the new `moduleSuffixes` compiler option. Using the TypeScript resolver is best because it supports many more features and is kept up-to-date with the latest changes in the Node ecosystem.

Our custom resolver will still run for developers on older versions of TypeScript.

This PR deliberately avoids making the same changes to our custom resolver `@rnx-kit/typescript-react-native-resolver` and the related `rn-tsc` package `@rnx-kit/typescript-react-native-compiler`. Instead, it marks those packages as deprecated in favor of using `@rnx-kit/metro-plugin-typescript` and `tsc` >= 4.7 with `moduleSuffixes`. This sets the stage for removing these packages from the repo. Customers will still be able to use them from NPM, but we won't develop or maintain them.

**NOTE**: This PR adds a safety check in our custom resolver. It detects when developers are using TypeScript path remapping features (which is doesn't support), and fails with an explicit message. This has caused problems which manifest in odd ways, and the explicit failure should help developers detect and move past any issues more quickly.

Resolves #1944.
Resolves #1945.

### Test plan

This PR adds tests to cover both pre-4.7 and post-4.7 scenarios in the Metro plugin.


kelset:
Great stuff, I like the tests and the deprecation warnings 👍

Just as a data point to keep in mind, TS 4.7 was released back in May https://github.com/microsoft/TypeScript/releases/tag/v4.7.2

packages/metro-plugin-typescript/src/resolver.ts
```diff
@@ -0,0 +1,263 @@
...
+  moduleName: string,
+  containingFile: string,
+  options: ts.CompilerOptions
+): ts.CompilerOptions {
+  if (!options.moduleSuffixes) {
+    //  `moduleSuffixes` is not defined. Return a copy of the input object with
+    //  `moduleSuffixes` set to the list of React Native platform extensions.
+    return {
+      ...options,
+      moduleSuffixes: [...context.platformFileExtensions],
```
tido64, line 36: Can we reference `context.platformFileExtensions` directly instead of making a copy?

packages/metro-plugin-typescript/src/resolver.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  //  `moduleSuffixes` is already defined in the module's compiler options.
+  //  We cannot safely modify it because we don't know how to order the entries
+  //  when adding those needed for React Native. Ordering matters because it controls
+  //  file selection. e.g. If "ios" comes before "native", then "foo.ios.ts" will be
+  //  resolved ahead of "foo.native.ts".
+  //
+  //  The best we can do is check `moduleSuffixes` to see if the React Native entries
+  //  we need are already there, in order. If not, we have to fail.
```
tido64, line 47: I'm not sure I understand why we need to do this. Shouldn't the app's `moduleSuffixes` always be used? Does it make sense to surface Android errors if we are targeting iOS? What happens if we have a platform-agnostic dependency with `moduleSuffixes: ["native", ""]`?

packages/metro-plugin-typescript/src/host.ts
```diff
@@ -0,0 +1,82 @@
...
+          platform,
+          disableReactNativePackageSubstitution
+        ),
+      };
+
+      host.resolveModuleNames = resolveModuleNames.bind(undefined, context);
+      host.resolveTypeReferenceDirectives = resolveTypeReferenceDirectives.bind(
+        undefined,
+        context
+      );
```
tido64, line 67: Why do we need to override `resolveTypeReferenceDirectives`? Where do we differ from vanilla TypeScript?

packages/metro-plugin-typescript/src/resolver.ts
afoxman: This is trying to ensure that any package we see -- app or dependency -- has moduleSuffixes that match the target react native platform you gave on the command-line.

Example: If you're targeting iOS, but you've put moduleSuffixes = [".android", ".native", ""] in your tsconfig.json, then we can't proceed because the ios file extensions are missing.

Example 2: you pull in a dependency that specifies moduleSuffixes -- maybe for a whole different purpose unrelated to RN. We need to detect that and fail gracefully.

If a package has moduleSuffixes already, like example you gave -- ["native", ""] -- we can't proceed. If we did, file resolution for that package wouldn't be consistent with the rest of the build. It might kind of work, as in this case, if the package doesn't use .ios or .android files, but we don't know that. This is why we have the detailed error message with a link to the issue. I doubt we'll get many reports there,but its an opportunity to see if there are cases which are sprouting up that we didn't think of.

tido64: > Example: If you're targeting iOS, but you've put moduleSuffixes = [".android", ".native", ""] in your tsconfig.json, then we can't proceed because the ios file extensions are missing.

I think this is a wrong assumption to make. What if the intention of the lib author is to use `.native` for all platforms except for Android?

> Example 2: you pull in a dependency that specifies moduleSuffixes -- maybe for a whole different purpose unrelated to RN. We need to detect that and fail gracefully.

I think this is exactly why only the app's `moduleSuffixes` should be applied.

> If a package has moduleSuffixes already, like example you gave -- ["native", ""] -- we can't proceed.

This will fail on packages such as AsyncStorage: `.native` is for _all_ native platforms. Files without a suffix is for web.

packages/metro-plugin-typescript/src/resolver.ts
afoxman: > Great stuff, I like the tests and the deprecation warnings 👍
> 
> Just as a data point to keep in mind, TS 4.7 was released back in May https://github.com/microsoft/TypeScript/releases/tag/v4.7.2 and as far as I can tell it's still downloaded a few M times per week https://www.npmjs.com/package/typescript?activeTab=versions
> 
> so deprecation will probably have to stay in place for maybe another 6 months or so

Yes, TS 4.7 is definitely still "new". And what we have still works. The last step is deleting the code from our repo. The NPM packages will still live on, though, for anyone who still needs them. The biggest "loss" for those people will be the READMEs on our site, though they'll still be in the packages themselves.

afoxman:
> Great stuff, I like the tests and the deprecation warnings 👍
> 
> Just as a data point to keep in mind, TS 4.7 was released back in May https://github.com/microsoft/TypeScript/releases/tag/v4.7.2 and as far as I can tell it's still downloaded a few M times per week https://www.npmjs.com/package/typescript?activeTab=versions
> 
> so deprecation will probably have to stay in place for maybe another 6 months or so

Yes, TS 4.7 is definitely still "new". And what we have still works. The last step is deleting the code from our repo. The NPM packages will still live on, though, for anyone who still needs them. The biggest "loss" for those people will be the READMEs on our site, though they'll still be in the packages themselves.

packages/metro-plugin-typescript/src/resolver.ts
afoxman: > I think this is a wrong assumption to make. What if the intention of the lib author is to use `.native` for all platforms except for Android?

Are you saying I should allow the special case for moduleSuffixes to be allowed when it has everything except the platform name? I could do that...

> > Example 2: you pull in a dependency that specifies moduleSuffixes -- maybe for a whole different purpose unrelated to RN. We need to detect that and fail gracefully.
> 
> I think this is exactly why only the app's `moduleSuffixes` should be applied.

I don't think we can do that. If the library author explicitly defined moduleSuffixes, they are expressing intent about how files in their package should be resolved. If we change that, then we could end up with a different/broken version of the library because moduleSuffixes controls which files are imported/required.

Remember moduleSuffixes isn't an RN feature in TS, it's a general purpose feature that can be used for any file suffix.

> > If a package has moduleSuffixes already, like example you gave -- ["native", ""] -- we can't proceed.
> 
> This will fail on packages such as AsyncStorage: `.native` is for _all_ native platforms. Files without a suffix is for web.

If files without a suffix are for web, then they must have the convention that EVERY file has a .native variant for RN builds. Otherwise, bundling would break - regardless of moduleSuffixes. Putting web code into "extensionless" files breaks the RN concept of platform file extensions altogether. Metro, for example, would get tripped up by that if one of the source files didn't have a .native variant.

afoxman: On additional note on using tsconfig from app vs library -- this is entirely controlled by TypeScript itself, now that we're using a delegating resolver. When they call into our hooked function, they pass us the CompilerOptions to use for resolving that particular module to a file. 

I'm not sure if TypeScript reads the tsconfig.json for that parent package of each module, of if it uses the top-level (app) tsconfig.json. But whatever TypeScript does is the defacto standard, and it is what developers will expect.

afoxman: I added this case.

tido64 (approved):
I'm still uncertain about whether it makes sense to verify `moduleSuffixes`, but I think this is good enough to merge. We can deal with `moduleSuffixes` related issues later.

PR #2084, opened 2023-01-06
feat(cli): add new `plugins` option

tido64:
### Description

Generalized CLI's plugins configuration.

This is an alternative implementation of #2029.

### Remaining work

- [x] Update website

### Test plan

Build:

```
cd packages/test-app
yarn build --dependencies
```

Run `yarn build:ios` with random legacy options:

```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index c651e9ae..69779731 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -79,6 +79,8 @@
         "entryFile": "src/index.ts",
         "assetsDest": "dist",
         "treeShake": false,
+        "detectCyclicDependencies": true,
+        "typescriptValidation": true,
         "plugins": [
           "@rnx-kit/metro-plugin-cyclic-dependencies-detector",
           [
```

or:

```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index c651e9ae..f2534eca 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -79,6 +79,13 @@
         "entryFile": "src/index.ts",
         "assetsDest": "dist",
         "treeShake": false,
+        "detectCyclicDependencies": true,
+        "detectDuplicateDependencies": {
+          "ignoredModules": [
+            "react-is"
+          ]
+        },
+        "typescriptValidation": true,
         "plugins": [
           "@rnx-kit/metro-plugin-cyclic-dependencies-detector",
           [
```

Example output:

```
test-app % yarn bundle:ios
yarn run v1.22.19
$ react-native rnx-bundle --platform ios
info Bundling ios...
warn 'detectCyclicDependencies', 'detectDuplicateDependencies', 'typescriptValidation' have been replaced by a new 'plugins' option. Below is an equivalent config:

    "plugins": [
      "@rnx-kit/metro-plugin-cyclic-dependencies-detector",
      ["@rnx-kit/metro-plugin-duplicates-checker", {
        "ignoredModules": [
          "react-is"
        ]
      }],
      "@rnx-kit/metro-plugin-typescript"
    ]

For more details, see https://github.com/microsoft/rnx-kit/tree/main/packages/cli#readme
                    Welcome to Metro!
              Fast - Scalable - Integrated


info Writing bundle output to:, dist/main.ios.jsbundle
info Writing sourcemap output to:, dist/main.ios.jsbundle.map
info Done writing bundle output
info Done writing sourcemap output
info Copying 6 asset files
info Done copying assets
✨  Done in 3.61s.
```

### DIFF
--- a/packages/cli/src/metro-config.ts
+++ b/packages/cli/src/metro-config.ts
@@ -25,6 +25,59 @@ function importPlugin(name: string) {
   return plugin.default || plugin;
 }
 
+function readLegacyOptions({
+  detectCyclicDependencies,
+  detectDuplicateDependencies,
+  typescriptValidation,
+}: BundlerPlugins): string | undefined {
+  const oldOptions = [];
+  const deprecatedOptions: Record<string, Record<string, unknown> | true> = {};
+
+  if (detectCyclicDependencies) {
+    oldOptions.push("detectCyclicDependencies");
+    deprecatedOptions["@rnx-kit/metro-plugin-cyclic-dependencies-detector"] =
+      detectCyclicDependencies;
+  }
+  if (detectDuplicateDependencies) {
+    oldOptions.push("detectDuplicateDependencies");
+    deprecatedOptions["@rnx-kit/metro-plugin-duplicates-checker"] =
+      detectDuplicateDependencies;
+  }
+  if (typescriptValidation) {
+    oldOptions.push("typescriptValidation");
+    deprecatedOptions["@rnx-kit/metro-plugin-typescript"] =
+      typescriptValidation;
+  }
+
+  if (oldOptions.length === 0) {
+    return undefined;
+  }
+
+  return (
+    `'${oldOptions.join("', '")}' have been replaced by a new 'plugins' ` +
+    "option. Below is an equivalent config:\n" +
+    "\n" +
+    '    "plugins": [\n' +
+    Object.entries(deprecatedOptions)
+      .map(([name, options]) => {
+        if (typeof options === "object") {
+          const json = JSON.stringify(options, null, 2)
+            .split("\n")
+            .join("\n      ");
+          return `      ["${name}", ${json}]`;
+        } else {
+          return `      "${name}"`;
+        }
+      })
+      .join(",\n") +
+    "\n" +
+    "    ]\n" +
+    "\n" +
+    "For more details, see " +
+    "https://github.com/microsoft/rnx-kit/tree/main/packages/cli#readme"
+  );
+}
+
 /**
  * Customize the Metro configuration.
  *
@@ -47,18 +100,9 @@ export function customizeMetroConfig(
     SerializerConfigT["experimentalSerializerHook"]
   > = {};
 
-  const oldOptions = [
-    "detectCyclicDependencies",
-    "detectDuplicateDependencies",
-    "typescriptValidation",
-  ];
-  if (oldOptions.some((option) => option in bundlerPlugins)) {
-    const deprecatedOptions = oldOptions.map((key) => `'${key}'`).join(", ");
-    warn(
-      `${deprecatedOptions} have been replaced by a new 'plugins' option. ` +
-        "Please use this new option instead. For more details, see" +
-        "https://github.com/microsoft/rnx-kit/tree/main/packages/cli#readme"
-    );
+  const legacyWarning = readLegacyOptions(bundlerPlugins);
+  if (legacyWarning) {
+    warn(legacyWarning);
 
     const {
       detectCyclicDependencies,
--- a/packages/cli/src/metro-config.ts
+++ b/packages/cli/src/metro-config.ts
@@ -1,4 +1,5 @@
 import type { BundlerPlugins } from "@rnx-kit/config";
+import { warn } from "@rnx-kit/console";
 import { CyclicDependencies } from "@rnx-kit/metro-plugin-cyclic-dependencies-detector";
 import { DuplicateDependencies } from "@rnx-kit/metro-plugin-duplicates-checker";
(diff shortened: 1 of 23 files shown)


tido64:
cc @acoates-ms

acoates-ms (approved):

packages/cli/src/metro-config.ts
```diff
@@ -19,34 +34,78 @@ import type { InputConfigT, SerializerConfigT } from "metro-config";
...
+        : [entry, undefined];
+      const plugin = importPlugin(module);
+      switch (plugin.type) {
+        case "analyzer":
+          metroPlugins.push(plugin(options));
+          break;
+
+        case "serializerHook": {
+          const hook = plugin(options, print);
+          metroConfig.serializer.experimentalSerializerHook = hook;
```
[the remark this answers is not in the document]
tido64: Yes, I think that makes sense. Let me come up with something.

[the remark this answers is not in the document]
tido64: I checked Metro and it looks like we should be able to chain the hooks. The return value is ignored.

packages/cli/src/metro-config.ts
```diff
@@ -19,34 +34,78 @@ import type { InputConfigT, SerializerConfigT } from "metro-config";
  */
 export function customizeMetroConfig(
   metroConfigReadonly: InputConfigT,
-  {
-    detectCyclicDependencies,
-    detectDuplicateDependencies,
-    treeShake,
-    typescriptValidation,
-  }: BundlerPlugins,
+  bundlerPlugins: BundlerPlugins,
```
afoxman, line 37: nit: I don't know what name to use, but `BundlerPlugins` probably isn't right anymore since it contains a `plugins` member alongside `treeShake`. Up to you if you want to rename, just doesn't seem to fit anymore.

packages/cli/src/metro-config.ts
```diff
@@ -19,34 +34,78 @@ import type { InputConfigT, SerializerConfigT } from "metro-config";
...
+  const serializerHooks: SerializerConfigT["experimentalSerializerHook"][] = [];
+
+  const oldOptions = [
+    "detectCyclicDependencies",
+    "detectDuplicateDependencies",
+    "typescriptValidation",
+  ];
+  if (oldOptions.some((option) => option in bundlerPlugins)) {
+    const deprecatedOptions = oldOptions.map((key) => `'${key}'`).join(", ");
+    warn(
```
afoxman, line 54: Consider replacing individual warnings with a single warning about using deprecated props, and include a paste-able equivalent they can drop into config.

packages/cli/src/metro-config.ts
```diff
@@ -19,34 +34,78 @@ import type { InputConfigT, SerializerConfigT } from "metro-config";
...
+        ? entry
+        : [entry, undefined];
+      const plugin = importPlugin(module);
+      switch (plugin.type) {
+        case "serializer":
+          metroPlugins.push(plugin(options));
+          break;
+
+        case "serializerHook":
+          serializerHooks.push(plugin(options, print));
```
afoxman, line 95: If this evolves over time, a more durable approach would be to use a dictionary with named entries, rather than positional args:
```
{
  options,
  print
}
```

packages/metro-plugin-cyclic-dependencies-detector/src/index.ts
```diff
@@ -16,3 +16,7 @@ export function CyclicDependencies(
     detectCycles(entryPoint, graph, pluginOptions);
   };
 }
+
+CyclicDependencies.type = "serializer";
+
+export default CyclicDependencies;
```
afoxman, line 22: The changeset file says 'minor' but this seems like a breaking change. Same with the other plugins.

afoxman (commented):
Looks good to me. There's one question I had about versioning -- does the change to `export default` constitute a breaking change? If so, the changeset files should switch from "minor" to "major". Or go back to named exports.

packages/metro-plugin-cyclic-dependencies-detector/src/index.ts
tido64: We still kept `export function CyclicDependencies` further up, so this shouldn't be a breaking change.

packages/cli/src/metro-config.ts
tido64: Changed this to output the equivalent:
```
warn 'detectCyclicDependencies', 'detectDuplicateDependencies', 'typescriptValidation' have been replaced by a new 'plugins' option. Below is an equivalent config:

    "plugins": [
      "@rnx-kit/metro-plugin-cyclic-dependencies-detector",
      ["@rnx-kit/metro-plugin-duplicates-checker", {
        "ignoredModules": [
          "react-is"
        ]
      }],
      "@rnx-kit/metro-plugin-typescript"
    ]

For more details, see https://github.com/microsoft/rnx-kit/tree/main/packages/cli#readme
```

packages/cli/src/metro-config.ts
tido64: I moved `plugins` and `treeShake` out of there. `treeShake` is only available during bundling, and `plugins` is just a single field now.

packages/cli/src/metro-config.ts
tido64: I think I may have misunderstood this initially as we want to ensure unique serializer hooks (and made changes to accommodate). I don't think I will change the signature of the hooks in this PR. We can do that whenever it makes sense.

afoxman (approved):

PR #2100, opened 2023-01-13
chore: so long, @afoxman, and thanks for all the fish

tido64:
### Description

Remove @afoxman from `CODEOWNERS`.

### Test plan

n/a


kelset (approved):
😢

PR #2108, opened 2023-01-17
fix(metro-resolver-symlinks): also load `require` if `default` and `node` are missing

tido64:
### Description

Handle packages that are missing the `default` exports conditional

### Test plan

Tested internally.


tido64:
cc @crazyfraggle

kelset (approved):

PR #2115, opened 2023-01-17
chore: remove `jest-extended` to reduce maintenance overhead

tido64:
### Description

Remove `jest-extended` to reduce maintenance overhead

### Test plan

Existing tests should pass.


packages/metro-plugin-typescript/package.json
```diff
@@ -39,8 +39,7 @@
   },
   "devDependencies": {
     "@rnx-kit/scripts": "*",
-    "@types/metro": "^0.66.0",
-    "jest-extended": "^2.0.0"
+    "@types/metro": "^0.66.0"
```
kelset, line 42: I'm scared by the answer but since I'm noticing this... who's maintaining the Metro types? DT repo?

kelset (approved):

packages/metro-plugin-typescript/package.json
tido64: ~~Adam and~~ I are.

PR #2127, opened 2023-01-18
Add metro plugin support for @rnx-kit/third-party-notices

AAHAbbas:
### Description

Adding metro plugin support for @rnx-kit/third-party-notices as it makes integration simpler and more streamlined.
### Test plan

Build and install pods:
```
cd packages/test-app
yarn build --dependencies
```

Ran `yarn bundle` with these changes:
```diff
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index c651e9ae..3c25f2df 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -38,6 +38,7 @@
     "@rnx-kit/cli": "*",
     "@rnx-kit/metro-config": "*",
     "@rnx-kit/metro-plugin-cyclic-dependencies-detector": "*",
+    "@rnx-kit/third-party-notices": "*",
     "@rnx-kit/metro-plugin-duplicates-checker": "*",
     "@rnx-kit/metro-resolver-symlinks": "*",
     "@rnx-kit/metro-serializer": "*",
@@ -89,7 +90,14 @@
               ]
             }
           ],
-          "@rnx-kit/metro-plugin-typescript"
+          "@rnx-kit/metro-plugin-typescript",
+          [
+            "@rnx-kit/third-party-notices",
+            {
+              "rootPath": ".",
+              "sourceMapFile": "./dist/main.android.bundle.map"
+            }
+          ]
         ],
         "targets": [
           "android",
```

Example *.LICENSE.txt file generated:
[*.LICENSE.txt](https://github.com/microsoft/rnx-kit/files/10455109/main.android.bundle.LICENSE.txt)

### DIFF
--- a/.changeset/clever-dancers-itch.md
+++ b/.changeset/clever-dancers-itch.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/third-party-notices": minor
+---
+
+Add metro plugin support for @rnx-kit/third-party-notices
--- a/.changeset/clever-dancers-itch.md
+++ b/.changeset/clever-dancers-itch.md
@@ -2,4 +2,4 @@
 "@rnx-kit/third-party-notices": minor
 ---
 
-Add metro plugin support for @rnx-kit/third-party-notices
+Add Metro plugin support
--- a/packages/third-party-notices/README.md
+++ b/packages/third-party-notices/README.md
@@ -61,3 +61,22 @@ writeThirdPartyNotices({
   sourceMapFile: "./dist/myPackage.js.map",
 });
 ```
+### As a plugin
+Import and add the plugin to `ThirdPartyNotices` in your `metro.config.js`, and
+optionally configure it to your liking:
+
+```diff
+ const { makeMetroConfig } = require("@rnx-kit/metro-config");
++const { ThirdPartyNotices } = require("@rnx-kit/third-party-notices");
++const { MetroSerializer } = require("@rnx-kit/metro-serializer");
+
+ module.exports = makeMetroConfig({
+   serializer: {
++    customSerializer: MetroSerializer([
++      ThirdPartyNotices({
++        rootPath: ".",
++        sourceMapFile: "./dist/myPackage.js.map",
++      }),
++    ]),
+   },
+ });
--- a/packages/third-party-notices/README.md
+++ b/packages/third-party-notices/README.md
@@ -61,25 +61,3 @@ writeThirdPartyNotices({
   sourceMapFile: "./dist/myPackage.js.map",
 });
 ```
-
-### As a plugin
-
-Import and add the plugin to `ThirdPartyNotices` in your `metro.config.js`, and
-optionally configure it to your liking:
-
-```diff
- const { makeMetroConfig } = require("@rnx-kit/metro-config");
-+const { ThirdPartyNotices } = require("@rnx-kit/third-party-notices");
-+const { MetroSerializer } = require("@rnx-kit/metro-serializer");
-
- module.exports = makeMetroConfig({
-   serializer: {
-+    customSerializer: MetroSerializer([
-+      ThirdPartyNotices({
-+        rootPath: ".",
-+        sourceMapFile: "./dist/myPackage.js.map",
-+      }),
-+    ]),
-   },
- });
-```
--- a/packages/third-party-notices/README.md
+++ b/packages/third-party-notices/README.md
@@ -61,7 +61,9 @@ writeThirdPartyNotices({
   sourceMapFile: "./dist/myPackage.js.map",
 });
 ```
+
 ### As a plugin
+
 Import and add the plugin to `ThirdPartyNotices` in your `metro.config.js`, and
 optionally configure it to your liking:
 
@@ -80,3 +82,4 @@ optionally configure it to your liking:
 +    ]),
    },
  });
+```
--- a/packages/third-party-notices/README.md
+++ b/packages/third-party-notices/README.md
@@ -61,3 +61,22 @@ writeThirdPartyNotices({
   sourceMapFile: "./dist/myPackage.js.map",
 });
 ```
+
+### As a plugin
+
+Import and add the plugin to `ThirdPartyNotices` in your `metro.config.js`, and
+optionally configure it to your liking:
+
+```diff
+ const { makeMetroConfig } = require("@rnx-kit/metro-config");
++const { ThirdPartyNotices } = require("@rnx-kit/third-party-notices");
++const { MetroSerializer } = require("@rnx-kit/metro-serializer");
+
+ module.exports = makeMetroConfig({
+   serializer: {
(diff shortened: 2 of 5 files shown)


packages/third-party-notices/src/index.ts
```diff
@@ -1,5 +1,22 @@
...
+export function ThirdPartyNotices(
+  options: WriteThirdPartyNoticesOptions
+): MetroPlugin {
+  return (
+    _entryPoint: string,
+    _preModules: ReadonlyArray<Module>,
+    _graph: Graph,
+    _options: SerializerOptions
+  ) => {
+    writeThirdPartyNotices(options);
```
tido64, line 16: I don't think we can call `writeThirdPartyNotices` here. At this point, the source map file has not been written to disk yet. However, you can extract modules from `graph` like we do in `metro-plugin-duplicates-checker`:

https://github.com/microsoft/rnx-kit/blob/f719c3dd52be7c2e154a43ea2dc8df6bcf425e5e/packages/metro-plugin-duplicates-checker/src/gatherModules.ts#L75-L81

You can probably copy `gatherModulesFromSources` and modify it to store the module path instead. This module map can then be fed directly to `extractLicenses`:

https://github.com/microsoft/rnx-kit/blob/f719c3dd52be7c2e154a43ea2dc8df6bcf425e5e/packages/third-party-notices/src/write-third-party-notices.ts#L54

Once you've done that, the `sourceMapFile` option should be redundant. I think we can also get `rootPath` from `options.projectRoot`.

tido64 (changes requested):
Thanks for working on this @AAHAbbas :heart:

I left a few comments. You know where to find me if you have any questions. 😉

packages/third-party-notices/src/index.ts
tido64: You should probably also use the source map location to determine the output path. For example:
```js
    const prefix = options.sourceMapUrl?.replace(/\.map$/, "") ?? "main";
    const outfile = prefix + ".LICENSE.txt";
```

tido64: Another thing you should consider is whether you want to generate the TPN only when `options.dev` is false.

packages/third-party-notices/README.md
```diff
@@ -61,3 +61,25 @@ writeThirdPartyNotices({
...
++const { MetroSerializer } = require("@rnx-kit/metro-serializer");
+
+ module.exports = makeMetroConfig({
+   serializer: {
++    customSerializer: MetroSerializer([
++      ThirdPartyNotices({
++        rootPath: ".",
++        sourceMapFile: "./dist/myPackage.js.map",
++      }),
++    ]),
```
tido64, line 82: Do we need to update this now?

packages/third-party-notices/src/index.ts
```diff
@@ -1,5 +1,31 @@
...
+import { Graph, Module, SerializerOptions } from "metro";
+import { WriteThirdPartyNoticesOptions } from "./types";
+import type { MetroPlugin } from "@rnx-kit/metro-serializer";
+import {
+  gatherModulesFromSources,
+  writeThirdPartyNoticesFromMap,
+} from "./write-third-party-notices";
+
 export { writeThirdPartyNotices } from "./write-third-party-notices";
+export function ThirdPartyNotices(
```
tido64, line 10: Nit:
```suggestion
export { writeThirdPartyNotices } from "./write-third-party-notices";

export function ThirdPartyNotices(
```

packages/third-party-notices/src/index.ts
```diff
@@ -1,5 +1,31 @@
...
+  options: WriteThirdPartyNoticesOptions
+): MetroPlugin {
+  return (
+    _entryPoint: string,
+    _preModules: ReadonlyArray<Module>,
+    graph: Graph,
+    serializerOptions: SerializerOptions
+  ) => {
+    if (!serializerOptions.dev) {
```
tido64, line 19: Shouldn't this be the other way around?
```suggestion
    if (serializerOptions.dev) {
```

packages/third-party-notices/src/write-third-party-notices.ts
```diff
@@ -2,6 +2,7 @@ import fs from "fs";
 import { findPackage, readPackage } from "@rnx-kit/tools-node/package";
 import { resolve } from "path";
 import { promisify } from "util";
+import type { BasicSourceMap } from "metro-source-map";
```
tido64, line 5: Do we need to add `@types/metro-source-map` to `package.json` for this type?

packages/third-party-notices/src/write-third-party-notices.ts
```diff
@@ -64,6 +65,27 @@ export async function writeThirdPartyNotices(
...
+    ? createLicenseJSON(licenses)
+    : createLicenseFileContents(licenses, preambleText, additionalText);
+
+  if (!outputFile) {
+    const prefix = sourceMapFile?.replace(/\.map$/, "") ?? "main";
+    outputFile = prefix + ".LICENSE.txt";
+  }
+
+  const writeFileAsync = promisify(fs.writeFile);
+  await writeFileAsync(outputFile, outputText, "utf8");
```
tido64, line 86: Can we use `fs.writeFileSync` here? It's going to be faster not to wait for a tick in Node.
```suggestion
  fs.writeFileSync(outputFile, outputText);
```

.changeset/clever-dancers-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rnx-kit/third-party-notices": minor
+---
+
+Add metro plugin support for @rnx-kit/third-party-notices
```
tido64, line 5: This entry will be added to `@rnx-kit/third-party-notices`'s change log file. It's redundant to mention it again.

```suggestion
Add Metro plugin support
```

packages/third-party-notices/README.md
AAHAbbas: Removed the changes.

packages/third-party-notices/src/write-third-party-notices.ts
AAHAbbas: Right, added `@types/metro-source-map` and `@types/metro` to `devDependencies`.

packages/third-party-notices/src/write-third-party-notices.ts
AAHAbbas: Fixed.

packages/third-party-notices/README.md
tido64: I think we want to keep this. Just update the options.

AAHAbbas: Ohh, I see. I'll fix it. Will remove the `rootPath` option.

tido64: Yeah, just reflect the latest changes 😄

PR #2135, opened 2023-01-20
chore: enable `consistent-type-imports` ESLint rule

tido64:
### Description

Ensure we use `import type` consistently.

### Test plan

n/a


kelset (approved):
very cool much smooth me like :shipit:

PR #2145, opened 2023-01-25
fix(metro-plugin-typescript): lower Node version requirement

tido64:
### Description

There's no reason why we need to be on 14.18+. The first 14 LTS should also work.

### Test plan

n/a


tido64:
cc @acoates-ms

kelset (approved):

PR #2148, opened 2023-01-25
fix(cli): resolve `cli-server-api` starting from `react-native`

tido64:
### Description

Resolve `@react-native-community/cli-server-api` starting from `react-native` to avoid relying hoisting.

### Test plan

Tested internally.


tido64:
cc @crazyfraggle

kelset (approved):

PR #2157, opened 2023-01-31
fix(metro-serializer-esbuild): don't add namespace to all source files

tido64:
### Description

Don't add namespace to all source files. esbuild currently adds it to all file paths in the source map. This prevents tools from resolving files correctly.

### Test plan

```
cd packages/test-app
yarn build --dependencies
yarn bundle+esbuild --platform ios --dev false
node ../third-party-notices/lib/cli.js --rootPath . --sourceMapFile dist/main+esbuild.ios.jsbundle.map --outputFile LICENSE.txt
```

Verify that:
- `LICENSE.txt` is non-empty
- `dist/main+esbuild.ios.jsbundle.map` only contains a single instance of `virtual:metro`

### DIFF
--- a/.changeset/fresh-pillows-matter.md
+++ b/.changeset/fresh-pillows-matter.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-serializer-esbuild": patch
+---
+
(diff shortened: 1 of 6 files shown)


acoates-ms (approved):

kelset (approved):
tested locally, everything working as expected 👍

Screenshot 2023-01-31 at 16 17 32

PR #2160, opened 2023-01-31
fix(metro-serializer-esbuild): virtual pre-modules should inherit namespace

tido64:
### Description

Virtual pre-modules should inherit the namespace of the top-level prelude.

### Test plan

Tested internally.


tido64:
cc @acoates

acoates-ms (approved):

PR #2162, opened 2023-02-01
fix(metro-resolver-symlinks): disable exports map

tido64:
### Description

When `experimental_retryResolvingFromDisk` is enabled, don't parse exports maps as they currently take precedence over the `react-native` field.

In the scenario I was investigating, it's increasing the bundle size because CJS was picked over ESM.

### Test plan

Tested internally.


tido64:
cc @acoates-ms

kelset (approved):

PR #2166, opened 2023-02-01
feat(eslint-plugin): add `no-const-enum` rule

tido64:
### Description

Added `no-const-enum` rule.

### Test plan

Tests were added.

### DIFF
--- a/packages/eslint-plugin/src/index.js
+++ b/packages/eslint-plugin/src/index.js
@@ -6,6 +6,7 @@ module.exports = {
     recommended: require("./configs/recommended"),
   },
   rules: {
+    "no-const-enum": require("./rules/no-const-enum"),
     "no-export-all": require("./rules/no-export-all"),
   },
 };
--- a/packages/eslint-plugin/src/rules/no-const-enum.js
+++ b/packages/eslint-plugin/src/rules/no-const-enum.js
@@ -1,6 +1,10 @@
 // @ts-check
 "use strict";
 
+/**
+ * @typedef {import("estree").Node & { const?: boolean; }} Node
+ */
+
 /** @type {import("eslint").Rule.RuleModule} */
 module.exports = {
   meta: {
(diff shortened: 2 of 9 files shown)


acoates-ms (approved):

packages/eslint-plugin/src/rules/no-const-enum.js
```diff
@@ -0,0 +1,32 @@
...
+    },
+  },
+  create: (context) => {
+    return {
+      TSEnumDeclaration: (/** @type {Node} */ node) => {
+        if (node.const) {
+          context.report({
+            node,
+            message:
+              "Prefer string literal unions over `const enum` to avoid bundle bloat, and package boundary issues.",
```
dzearing, line 26:
```suggestion
              "Prefer string literal unions over `const enum` to avoid bundle bloat and package boundary issues.",
```

packages/eslint-plugin/src/index.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
     recommended: require("./configs/recommended"),
   },
   rules: {
+    "no-const-enum": require("./rules/no-const-enum"),
```
dzearing, line 9: update the readme.md

dzearing: context: i point teams to this package, and they have asked in the past for docs explaining why export stars or const enums are bad. Consider pulling some of these notes into the readme:

const enums bad: https://hackmd.io/bBcd6R-1TB6Zq95PSquooQ?both
export * bad: https://hackmd.io/Z021hgSGStKlYLwsqNMOcg?both

dzearing (approved):
Thank you for making this! Consider my comments before checking in, but feel free to make a separate checkin to address.

packages/eslint-plugin/src/index.js
tido64: Thanks for these! I've included them in the readme.

PR #2167, opened 2023-02-01
@rnx-kit/metro-plugin-typescript fails to typecheck on windows

acoates-ms:
There is a mismatch of path separators when running on windows that causes @rnx-kit/metro-plugin-typescript to not type check anything when running on windows machines.


Dewsk (approved):
Great find and fix.

Saadnajmi (approved):

tido64 (approved):

PR #2174, opened 2023-02-02
feat(react-native-host): introduce @rnx-kit/react-native-host

tido64:
### Description

The aim of this package is to provide a backwards (and forwards) compatible way of initializing React Native, regardless of whether you're on New Architecture or have gone fully bridgeless.`@rnx-kit/react-native-host` will also a provide a
simple way to enable split bundles and service delivery.

Note: This currently only contains iOS/macOS code.

### Test plan

Internal testing.


packages/react-native-host/ios/ReactNativeHost.h
```diff
@@ -0,0 +1,53 @@
...
+
+#import <React/RCTBridgeDelegate.h>
+
+#import "RNXHostConfig.h"
+
+#if TARGET_OS_OSX
+#define RNXView NSView
+@class NSView;
+#else
+#define RNXView UIView
```
Saadnajmi, line 11: Q: In RN-macOS, I've been looking at making things like these typedefs to nice logic from preprocessor to compiler time. I'm curious what you think about such a change and whether it's worth doing here as well? I _think_ its beneficial, but I feel my main justification is a bias against `#define` statements 😅.

tido64: Do you have any concrete examples? I'm all for replacing preprocessor macros with something nicer.

Saadnajmi: See https://github.com/microsoft/react-native-macos/blob/main/React/Base/RCTUIKit.h#L486-L564

Elsewhere in this file/ the .m file we have `RCTPlatformView` which is essentially just this macro.

tido64: Are you saying we should import this file and use `RCTPlatformView` instead? Is this available in core?

Saadnajmi: I'm suggesting you change your existing macros with typedefs. I haven't tried upstreaming the iOS bits of `RCTUIKit` because then RN Core devs would have to remember to use either `RCTPlatformView` or `RCTUIView` in place of `UIView` for all new code. I didn't think that was sustainable. I'm open to the idea, just haven't figured out the best way to go about it. While the macros/typedefs RN-macOS has only live in RN-macOS, I think it's fine to make essentially identical duplicates. They're simple enough.

Saadnajmi: If you want to be fancy, obj-c also has `@compatibility_alias` (which we also use in 1 or 2 spots) but it seems... not well documented https://gcc.gnu.org/onlinedocs/gcc-4.5.4/gcc/compatibility_005falias.html

harrieshin: also, isn't the folderpath misleading that is iOS?

tido64: > I'm suggesting you change your existing macros with typedefs.

Yes, I completely forgot about those.

> also, isn't the folderpath misleading that is iOS?

Consumers typically don't see this path but I've renamed it to `mac`. Hopefully that covers iOS, macOS and more:

image

packages/react-native-host/cocoa/ReactNativeHost.h
```diff
@@ -0,0 +1,53 @@
...
+    NS_SWIFT_NAME(using(module:block:));
+
+/// Creates a React root view with the specified module and initial properties.
+///
+/// - Parameters:
+///     - moduleName: Name of the module to create root view of
+///     - initialProperties: Properties passed to the module
+- (RNXView *)viewWithModuleName:(NSString *)moduleName
+              initialProperties:(nullable NSDictionary *)initialProperties
+    NS_SWIFT_NAME(view(moduleName:initialProperties:));
```
mganandraj, line 49: Will View keep a strong reference to the host ?

mganandraj: I think having another class ReactNativeView which can be attached/detached with  a ReactNativeHost is an alternative and slightly better abstraction ..

packages/react-native-host/cocoa/ReactNativeHost.h
```diff
@@ -0,0 +1,53 @@
...
+typedef NSView RNXView;
+#else
+@class UIView;
+typedef UIView RNXView;
+#endif
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Hosts a React Native instance.
+@interface ReactNativeHost : NSObject <RCTBridgeDelegate>
```
mganandraj, line 18: I think this is a minimal but good foundation for what we want to achieve .. 
On split bundles, Teams iOS is using splitbundle load apis available in RCTBridge .. Android also has similar methods in the CatalystInstace.. So for non-dev mode, we could standardize the implementation to use splitbundle apis .. 
But, for dev server scenarios, we don't have a good solution to support split bundles !

mganandraj (approved):

packages/react-native-host/cocoa/ReactNativeHost.h
tido64: The host can be reached via the bridge. This is still true in New Arch as well. I added a convenience method for retrieving it:

image

kelset (approved):

PR #2180, opened 2023-02-06
feat(react-native-test-app-msal): add flag to support optional auth module registration

sveinar:
### Description

The issue I am trying to resolve here is concerning situations where the hostapp is providing its own auth module with its own registration. 

I am addressing the issue of 'multiple module with the same name' by preventing the sources for the MSAL provided auth module in this package to load if a special flag has been set.

The name of this flag is `com.microsoft.reacttestapp.msal.disableRnxAuthModule`which is a boolean.

This is a sample setting for disabling the auto-linking module registration if the customer wants to provide a custom auth module in their host-app themselves:

```gradle.properties
...
...
com.microsoft.reacttestapp.msal.disableRnxAuthModule=true
```

Please note that the modified build.gradle file here still must include the dependency to the `rnx-kit_react-native-auth` unconditionally. This is due to the fact that the host-provided auth module would be using the class definitions for the auth module it provides.

### Test plan

Two scenarios in mind when testing, with and without the hostapp providing its own auth module.

### DIFF
--- a/.changeset/short-fishes-speak.md
+++ b/.changeset/short-fishes-speak.md
@@ -1,5 +1,5 @@
 ---
-"@rnx-kit/react-native-test-app-msal": patch
+"@rnx-kit/react-native-test-app-msal": minor
 ---
 
 add flag to support optional auth module registration
--- a/.changeset/short-fishes-speak.md
+++ b/.changeset/short-fishes-speak.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/react-native-test-app-msal": patch
+---
+
+add flag to support optional auth module registration
--- a/packages/react-native-test-app-msal/android/build.gradle
+++ b/packages/react-native-test-app-msal/android/build.gradle
@@ -73,12 +73,12 @@ repositories {
     }
 }
 
-def rnxDisableAuthProperty = "com.microsoft.reacttestapp.msal.disableRnxAuthModule"
 def rnxAuth = findNodeModulesPath("@rnx-kit/react-native-auth")
+def rnxDisableAuthProperty = "com.microsoft.reacttestapp.msal.disableRnxAuthModule"
 def skipAuthModuleRegistration =
     rnxAuth != null &&
-        rootProject.hasProperty(rnxDisableAuthProperty) &&
-        rootProject.property(rnxDisableAuthProperty).toString().equalsIgnoreCase("true")
+    rootProject.hasProperty(rnxDisableAuthProperty) &&
+    rootProject.property(rnxDisableAuthProperty).toString().equalsIgnoreCase("true")
 
 android {
     def manifest = new JsonSlurper().parseText(findFile("app.json").text)
@@ -153,7 +153,7 @@ android {
 
         main.res.srcDirs += generatedResDir
 
-        if (skipAuthModuleRegistration || rnxAuth == null) {
+        if (skipAuthModuleRegistration) {
             main.java.srcDirs += "src/nornx/java"
         } else {
             main.java.srcDirs += "src/rnx/java"
--- a/packages/react-native-test-app-msal/android/build.gradle
+++ b/packages/react-native-test-app-msal/android/build.gradle
@@ -73,12 +73,11 @@ repositories {
     }
 }
 
-def rnxAuth = findNodeModulesPath("@rnx-kit/react-native-auth")
+def rnxAuthPackage = findNodeModulesPath("@rnx-kit/react-native-auth")
 def rnxDisableAuthProperty = "com.microsoft.reacttestapp.msal.disableRnxAuthModule"
-def skipAuthModuleRegistration =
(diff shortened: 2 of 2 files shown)


.changeset/short-fishes-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rnx-kit/react-native-test-app-msal": patch
```
tido64, line 2: This should be `minor`:
```suggestion
"@rnx-kit/react-native-test-app-msal": minor
```

packages/react-native-test-app-msal/android/build.gradle
```diff
@@ -73,7 +73,11 @@ repositories {
     }
 }
 
+def rnxDisableAuthProperty = "com.microsoft.reacttestapp.msal.disableRnxAuthModule"
 def rnxAuth = findNodeModulesPath("@rnx-kit/react-native-auth")
+def skipAuthModuleRegistration =
+    rootProject.hasProperty(rnxDisableAuthProperty) &&
+    rootProject.property(rnxDisableAuthProperty)
```
tido64, line 80: Should this property be tested against something? What happens when we set `com.microsoft.reacttestapp.msal.disableRnxAuthModule=false` in `gradle.properties`?

packages/react-native-test-app-msal/android/build.gradle
```diff
@@ -73,7 +73,11 @@ repositories {
     }
 }
 
+def rnxDisableAuthProperty = "com.microsoft.reacttestapp.msal.disableRnxAuthModule"
 def rnxAuth = findNodeModulesPath("@rnx-kit/react-native-auth")
+def skipAuthModuleRegistration =
+    rootProject.hasProperty(rnxDisableAuthProperty) &&
```
tido64, line 79: I think we should add the `rnxAuth` check here to avoid accidentally enabling auth module registration when it's not available:
```suggestion
def skipAuthModuleRegistration =
    rnxAuth != null &&
    rootProject.hasProperty(rnxDisableAuthProperty) &&
```

tido64 (commented):
Thanks for submitting this, @sveinar. I left a few comments.

.changeset/short-fishes-speak.md
sveinar: Makes sense! Thanks.

packages/react-native-test-app-msal/android/build.gradle
sveinar: You are right. Thinking that the 'object' would naturally translate to 'boolean' was a bit naive hehe.

packages/react-native-test-app-msal/android/build.gradle
sveinar: Agreed!

tido64 (approved):
Thanks again for adding this! Is an equivalent change for iOS needed?

packages/react-native-test-app-msal/android/build.gradle
```diff
@@ -74,6 +74,11 @@ repositories {
 }
 
 def rnxAuth = findNodeModulesPath("@rnx-kit/react-native-auth")
+def rnxDisableAuthProperty = "com.microsoft.reacttestapp.msal.disableRnxAuthModule"
+def skipAuthModuleRegistration =
+    rnxAuth != null &&
```
tido64, line 79: Actually, on closer inspection, this looks inverted. Shouldn't this be something like:
```suggestion
    rnxAuth == null ||
```

packages/react-native-test-app-msal/android/build.gradle
```diff
@@ -74,6 +74,11 @@ repositories {
...
+    rootProject.hasProperty(rnxDisableAuthProperty) &&
+    rootProject.property(rnxDisableAuthProperty).toString().equalsIgnoreCase("true")
```
tido64, line 81: Actually, on closer inspection, this looks inverted. Shouldn't this be something like:
```suggestion
    rnxAuth == null ||
    (rootProject.hasProperty(rnxDisableAuthProperty) &&
        rootProject.property(rnxDisableAuthProperty).toString().equalsIgnoreCase("true"))
```

sveinar:
> Thanks for submitting this, @sveinar. I left a few comments.

sveinar:
hihi, accidentally hit 'close' on the PR - apologizes.

I pushed and tested this again. Hope the formatting is still good (Android Studio insisted on a different formatting).

packages/react-native-test-app-msal/android/build.gradle
sveinar: Yes, effectively doing this makes the logic further down a bit simpler! Thanks.

tido64:
Looks good! Thanks again ❤️

sveinar:
> 



> adding

This change is not needed at all for iOS due to the way iOS handles modules. All good there. Tested ok!

PR #2187, opened 2023-02-08
Update new package creation command in CONTRIBUTING.md

SparshaSaha:
### Description

It is mentioned in the contribution guide that to create a new package we must run the command `yarn new-package --name <package_name>`. This does not seem to work. The command that works instead is `yarn new-package <package_name>`.

Updated the contribution guide

### Test plan

Since this is a documentation change, manual reviews should suffice


tido64 (approved):
Thanks for updating the docs! We haven't had to use `--name` in quite some time. 😛

Issue #2191, opened 2023-02-09
Update metro-resolver-symlinks for new Metro


poniraq:
### What happened?

There is an experimental functionality for a "metro-resolver-symlinks", that patches metro DependencyGraph to use `fs.existsSync` as fallback resolution method.

packages/metro-resolver-symlinks/src/utils/patchMetro.ts
```typescript
this._doesFileExist = (filePath: string): boolean => {
  return this._hasteFS.exists(filePath) || fileExists(filePath);
};
```

Right now it brakes latest version of Metro, because facebook team changed `_hasteFS` property name first to `_snapshotFS` and shortly after to `_fileSystem`.

### Affected Package

@rnx-kit/metro-resolver-symlinks

### Version

0.1.25

### Which platforms are you seeing this issue on?

- [X] Android
- [X] iOS
- [X] macOS
- [X] Windows

### System Information

```true
System:
    OS: macOS 13.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.77 GB / 16.00 GB
    Shell: 3.5.0 - /usr/local/bin/fish
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: Not Found
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
    Watchman: 2022.12.12.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.3 - /usr/local/lib/ruby/gems/2.7.0/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 22.2, iOS 16.2, macOS 13.1, tvOS 16.1, watchOS 9.1
    Android SDK: Not Found
  IDEs:
    Android Studio: 2021.3 AI-213.7172.25.2113.9123335
    Xcode: 14.2/14C18 - /usr/bin/xcodebuild
  Languages:
    Java: 11.0.15 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: Not Found
    react-native: Not Found
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

### Steps to Reproduce

just setup a project with latest react-native version & apply "@rnx-kit/metro-resolver-symlinks" with `experimental_retryResolvingFromDisk: true`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

[...]

PR #2192, opened 2023-02-10
docs(changeset): adjust `metro-resolver-symlinks` metro patch

poniraq:
adjust `metro-resolver-symlinks` metro patch to use any available hasteFS alias

### Description

There is an experimental functionality for a "metro-resolver-symlinks", that patches metro DependencyGraph to use `fs.existsSync` as fallback resolution method.

packages/metro-resolver-symlinks/src/utils/patchMetro.ts
```typescript
this._doesFileExist = (filePath: string): boolean => {
  return this._hasteFS.exists(filePath) || fileExists(filePath);
};
```

Right now it brakes latest version of Metro, because facebook team changed `_hasteFS` property name first to `_snapshotFS` and shortly after to `_fileSystem`.

One potential fix would be to just update the hasteFS property name to what it currently is.
```typescript
this._doesFileExist = (filePath: string): boolean => {
  return this._fileSystem.exists(filePath) || fileExists(filePath);
};
```

But it would break backward compatibility with all the previous metro-resolver versions, so a better approach would be to try all known hasteFS aliases.
```typescript
this._doesFileExist = (filePath: string): boolean => {
  const hasteFS = this._hasteFS || this._snapshotFS || this._fileSystem;
  return hasteFS.exists(filePath) || fileExists(filePath);
};
```

Resolves [#2191](https://github.com/microsoft/rnx-kit/issues/2191)

### Test plan

Issue can be reproduced by setting up hello world with latest react-native (that will include the latest metro version).
You then just apply the `metro-resolver-symlinks` as per README with `experimental_retryResolvingFromDisk: true` option.

```javascript
// metro.config.js
const MetroSymlinksResolver = require('@rnx-kit/metro-resolver-symlinks');

module.exports = {
  resolver: {
    resolveRequest: MetroSymlinksResolver({
      experimental_retryResolvingFromDisk: true,
    })
  }
};
```

The only possible edge-case or improvement i can think about is to either make hasteFS alias list configurable, so users can patch the problem themselves in the future. Because it looks like facebook team took a fancy to changing hasteFS property name.

### DIFF
--- a/packages/metro-resolver-symlinks/src/utils/patchMetro.ts
+++ b/packages/metro-resolver-symlinks/src/utils/patchMetro.ts
@@ -92,8 +92,8 @@ export function patchMetro(options: Options): void {
   DependencyGraph.prototype.orig__createModuleResolver =
     DependencyGraph.prototype._createModuleResolver;
   DependencyGraph.prototype._createModuleResolver = function (): void {
+    const hasteFS = this._hasteFS || this._snapshotFS || this._fileSystem;
     this._doesFileExist = (filePath: string): boolean => {
-      const hasteFS = this._hasteFS || this._snapshotFS || this._fileSystem;
       return hasteFS.exists(filePath) || fileExists(filePath);
     };
 
--- a/packages/metro-resolver-symlinks/src/utils/patchMetro.ts
+++ b/packages/metro-resolver-symlinks/src/utils/patchMetro.ts
@@ -93,7 +93,8 @@ export function patchMetro(options: Options): void {
     DependencyGraph.prototype._createModuleResolver;
   DependencyGraph.prototype._createModuleResolver = function (): void {
     this._doesFileExist = (filePath: string): boolean => {
-      return this._hasteFS.exists(filePath) || fileExists(filePath);
+      const hasteFS = this._hasteFS || this._snapshotFS || this._fileSystem;
+      return hasteFS.exists(filePath) || fileExists(filePath);
     };
 
     this.orig__createModuleResolver();
--- a/.changeset/shaggy-trains-press.md
+++ b/.changeset/shaggy-trains-press.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-resolver-symlinks": patch
+---
+
+adjust `metro-resolver-symlinks` metro patch to use any available hasteFS alias


poniraq:

packages/metro-resolver-symlinks/src/utils/patchMetro.ts
```diff
@@ -93,7 +93,8 @@ export function patchMetro(options: Options): void {
     DependencyGraph.prototype._createModuleResolver;
   DependencyGraph.prototype._createModuleResolver = function (): void {
     this._doesFileExist = (filePath: string): boolean => {
-      return this._hasteFS.exists(filePath) || fileExists(filePath);
+      const hasteFS = this._hasteFS || this._snapshotFS || this._fileSystem;
```
tido64, line 96: Can we move this line outside `this._doesFileExist` so we don't have to check every time it gets called?

tido64 (commented):
Thanks for fixing this! Love that super detailed description ❤️

packages/metro-resolver-symlinks/src/utils/patchMetro.ts
poniraq: I'm not sure. it would be ideal to intercept hasteFS instance in constructor, but this property is not immediately available. see `DependencyGraph` [source](https://github.dev/facebook/metro/blob/2c8d5ce35015d11aa828c36205807f176f474a88/packages/metro/src/node-haste/DependencyGraph.js#L118)

```javascript
constructor(
    config: ConfigT,
    options?: {
      +hasReducedPerformance?: boolean,
      +watch?: boolean,
    },
  ) {
    super();
    ...
    this._readyPromise = haste.build().then(({fileSystem, hasteModuleMap}) => {
      log(createActionEndEntry(initializingMetroLogEntry));
      config.reporter.update({type: 'dep_graph_loaded'});

      this._fileSystem = fileSystem; // <-- here
      this._hasteModuleMap = hasteModuleMap;
      ...
      this._createModuleResolver();
    });
  }
```

its populated only after `haste.build()` resolves. i suppose we could cache the result of this expression like this:
```typescript
let hasteFS = null;
this._doesFileExist = (filePath: string): boolean => {
  if (!hasteFS) {
    hasteFS = this._hasteFS || this._snapshotFS || this._fileSystem;
  }
  return hasteFS.exists(filePath) || fileExists(filePath);
};
```

but if hasteFS prop on `DependencyGraph` gets reassigned for some reason in the future, we would end up with a wrong HasteFS instance in cache. plus if its performance you're concerned about, then i'd say symlink resolution & `fs.existsSync` introduce enough of an overhead, so there should be no notable difference between calling this expression every time & caching it.

do you have any better ideas?

tido64: This code is inside `_createModuleResolver` and it doesn't get called till _after_ `_filesystem` is set:

```diff
   constructor(
     config: ConfigT,
     options?: {
       +hasReducedPerformance?: boolean,
       +watch?: boolean,
     },
   ) {
     super();
     ...
     this._readyPromise = haste.build().then(({fileSystem, hasteModuleMap}) => {
       log(createActionEndEntry(initializingMetroLogEntry));
       config.reporter.update({type: 'dep_graph_loaded'});
 
+      this._fileSystem = fileSystem; // <-- here
       this._hasteModuleMap = hasteModuleMap;
       ...
+      this._createModuleResolver(); // <-- patched method called here
     });
   }
```

I think it should be safe to capture `_filesystem` here. Or am I missing something?

poniraq: d'oh! you're right, thanks. i completely forgot what's exactly getting patched here. i'll make the change. what's the usual strategy here? do you want me to ammend the commit or just add a new one on top so the conversation is preserved?

tido64: We prefer adding new commits so we can get a sense of the history of changes.

poniraq: done. please let me know if i need to adjust commit message. you clearly have some sort of system here, but i can't find it in your contribution guide. or do i need to make every commit via Changelog?

tido64: Oh, we squash everything before merging so you only need the one change log. The separate commits are for when we need to go back to the PR and look at the history.

I'll add some more info in the contribution guide.

tido64 (approved):
Looks good! Thanks again for fixing the patch 😊

PR #2207, opened 2023-02-16
Add Help and new lines to rnx-start

johnf:
### Description

I like to be able to hit enter in rnx-start to clear space to more easily differentiate between previous and new log messages when debugging.

For good measure I also implemented a help function that reprints the initial help message.

### Test plan

* Run react-native rnx-start
* Press "h" and see help
* Hit enter a few times and see new lines

### DIFF
--- a/packages/cli/src/start.ts
+++ b/packages/cli/src/start.ts
@@ -260,6 +260,10 @@ export async function rnxStart(
             terminal.log(chalk.green("Reloading app..."));
             messageSocketEndpoint.broadcast("reload", undefined);
             break;
+
+          case "return":
+            terminal.log("");
+            break;
         }
       }
     });
--- a/packages/cli/src/start.ts
+++ b/packages/cli/src/start.ts
@@ -92,6 +92,20 @@ export async function rnxStart(
     cliOptions.customLogReporterPath
   );
 
+  const printHelp = () => {
+    const dim = chalk.dim;
+    const press = dim(" › Press ");
+    [
+      ["r", "reload the app"],
+      ["d", "open developer menu"],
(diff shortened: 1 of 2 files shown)


packages/cli/src/start.ts
```diff
@@ -92,6 +92,19 @@ export async function rnxStart(
...
   );
 
+  const printHelp = () => {
+    const dim = chalk.dim;
+    const press = dim(" › Press ");
+    [
+      ["r", "reload the app"],
+      ["d", "open developer menu"],
+      ["a", "show bundler address QR code"],
+      ["ctrl-c", "quit"],
```
tido64, line 102: I think we should also add `h` to the help message:

```suggestion
      ["a", "show bundler address QR code"],
      ["h", "show this help message"],
      ["ctrl-c", "quit"],
```

packages/cli/src/start.ts
```diff
@@ -251,6 +255,14 @@ export async function rnxStart(
             terminal.log(chalk.green("Reloading app..."));
             messageSocketEndpoint.broadcast("reload", undefined);
             break;
+
+          case "h":
+            printHelp();
+            break;
```
tido64, line 261: Nit: Can you sort the cases?

tido64 (commented):
Hi, thanks for adding this! I left a couple of comments, but it looks good otherwise.

Can you also run `yarn change` at the repo root to add a change log? The CI will fail if you don't have one.

johnf:
@tido64 thanks, all done

tido64 (approved):
Thanks again, @johnf!

PR #2233, opened 2023-02-21
chore: update `yarn.lock` to unblock publishing

tido64:
### Description

Update `yarn.lock` to unblock publishing

### Test plan

n/a


tido64:
Bypassing branch protections since this is blocking publishing.

PR #2234, opened 2023-02-21
Introduce react-native-error-trace-decorator which helps in efficiently symbolicating error stack traces

SparshaSaha:
### Description

Symbolicating error stack traces for react-native apps which have multiple bundles can be a complicated process.
This Pull Request adds a tool, `react-native-error-trace-decorator` which aims at simplifying this process by automating most of the steps required to symbolicate a stack trace which spans across multiple bundles.

This package will be experimental for now.

### Test plan

UTs have been written for all the files in the package.
We have tested this internally across different stack traces + sourcemap combinations and it works nicely.


incubator/react-native-error-trace-decorator/src/check_arguments_validity.ts
```diff
@@ -0,0 +1,81 @@
...
+      bundleIdentifier: "Identifier for Bundle 2",
+      sourcemap: "path/to/sourcemap/for/bundle2",
+    },
+  ],
+};
+const configFilePattern = `Please create a JSON config file in the following format:\n${JSON.stringify(
+  exampleConfigFile
+)}`;
+
+export const checkArgumentValidity = (args: minimist.ParsedArgs): boolean => {
```
tido64, line 28: `yargs` already does a lot of this and allows you to specify additional sanitization checks. Would it make sense to switch to that instead?

incubator/react-native-error-trace-decorator/src/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+import { error } from "@rnx-kit/console";
+import type { IConfigFile } from "./types";
+
+/**
+ * Takes a buffer, joins it with a newline and symbolicates it with the provided sourcemap
+ *
+ * @param buffer    a string array representing the buffer
+ * @param sourcemap sourcemap that must be used to symbolicate the passed buffer
+ */
+export const symbolicateBuffer = (buffer: string[], sourcemap: string) => {
```
tido64, line 12: For legibility reasons, we should use `function` instead of arrow-functions:

```suggestion
export function symbolicateBuffer(buffer: string[], sourcemap: string): void {
```

This applies to all files.

incubator/react-native-error-trace-decorator/src/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (buffer.length > 0) {
+    // Write buffer to a temp file
+    fse.writeFileSync("./temp.rnbuffertrace", buffer.join("\n"));
+
+    // Symbolicate buffer
+    ChildProcess.execSync(
+      `npx metro-symbolicate ${sourcemap} < ./temp.rnbuffertrace`,
```
tido64, line 19: Where does `metro-symbolicate` come from? Do we need to add it to dependencies?

incubator/react-native-error-trace-decorator/src/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+      error("Atleast one config is required");
+      return false;
+    }
+    // Parse through all configs and check if sourcemap files exist
+    for (
+      let bundleIndex = 0;
+      bundleIndex < configFile.configs.length;
+      bundleIndex++
+    ) {
+      const currentConfig = configFile.configs[bundleIndex];
```
tido64, line 50: Use for-of instead:
```suggestion
    for (const currentConfig of configFile.configs) {
```

incubator/react-native-error-trace-decorator/src/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+      if (
+        !currentConfig ||
+        !currentConfig.sourcemap ||
```
tido64, line 53: You can use optional chaining here:
```suggestion
        !currentConfig?.sourcemap ||
```

incubator/react-native-error-trace-decorator/src/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+        !fse.existsSync(currentConfig.sourcemap) ||
+        !currentConfig.bundleIdentifier
+      ) {
+        error(
+          `Config: { bundleIdentifier: ${currentConfig.bundleIdentifier}, sourcemap: ${currentConfig.sourcemap} } is not proper`
+        );
```
tido64, line 59: Should we split this up to provide better error messages?

incubator/react-native-error-trace-decorator/README.md
```diff
@@ -0,0 +1,121 @@
...
+If the tool is added as a dependency in your package, run the following:
+
+```sh
+yarn react-native-error-trace-decorator --errorFile <path_to_the_error_file> --configFile <path_to_the_config_file>
+```
+
+Alternatively, this tool can be directly run using npx,
+
+```sh
+npx react-native-error-trace-decorator --errorFile <path_to_the_error_file> --configFile <path_to_the_config_file>
```
tido64, line 120: We probably need to add a prefix here:
```suggestion
npx @rnx-kit/react-native-error-trace-decorator --errorFile <path_to_the_error_file> --configFile <path_to_the_config_file>
```

incubator/react-native-error-trace-decorator/src/index.ts
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env node
+
+import { symbolicateStackTraces } from "./symbolicate_stack_traces";
+
+symbolicateStackTraces();
```
tido64, line 5: Can we rename this to a `cli.ts`, and make `index.ts` re-export the function instead:

```suggestion
export { symbolicateStackTraces } from "./symbolicate_stack_traces";
```

incubator/react-native-error-trace-decorator/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@types/fs-extra": "^11.0.0",
+    "@types/minimist": "^1.2.0",
+    "eslint": "^8.0.0",
+    "jest": "^27.0.0",
+    "prettier": "^2.8.0",
+    "typescript": "^4.0.0"
+  },
+  "dependencies": {
+    "@rnx-kit/console": "*",
+    "fs-extra": "^11.1.0",
```
tido64, line 43: I've looked at all the instances where this is used, and it doesn't look like we use anything that doesn't already exist in Node. Can we replace it?

incubator/react-native-error-trace-decorator/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@rnx-kit/scripts": "*",
+    "@types/fs-extra": "^11.0.0",
+    "@types/minimist": "^1.2.0",
+    "eslint": "^8.0.0",
+    "jest": "^27.0.0",
+    "prettier": "^2.8.0",
+    "typescript": "^4.0.0"
+  },
+  "dependencies": {
+    "@rnx-kit/console": "*",
```
tido64, line 42: This should be set to a version:
```suggestion
    "@rnx-kit/console": "^1.0.11",
```

incubator/react-native-error-trace-decorator/src/extract_and_process_error_stack.ts
```diff
@@ -0,0 +1,89 @@
...
+     * In our experience, this works much faster than symbolicating line by line
+     *
+     * 1. Prints lines as is when there are no matches.
+     * 2. Keeps a running buffer of lines with same identifiers.
+     * 3. Whenever there is a identifier mismatch, flushes buffer and creates a new buffer
+     */
+    let buffer: string[] = [];
+    let bufferConfig: IBundleInterface | undefined;
+
+    errorFile.forEach((errorLine: string) => {
```
tido64, line 36: Nit: Prefer for-of:
```suggestion
    for (const errorLine of errorFile) {
```

incubator/react-native-error-trace-decorator/src/extract_and_process_error_stack.ts
```diff
@@ -0,0 +1,89 @@
...
+const getIdentifierForLine = (
+  errorLine: string,
+  configs: IBundleInterface[]
+) => {
+  for (let configIndex = 0; configIndex < configs.length; configIndex++) {
+    if (errorLine.includes(configs[configIndex].bundleIdentifier)) {
+      return configs[configIndex];
+    }
+  }
+  return;
```
tido64, line 88: Use `find`:
```suggestion
  return configs.find((config) => errorLine.includes(config.bundleIdentifier));
```

incubator/react-native-error-trace-decorator/src/check_arguments_validity.ts
```diff
@@ -0,0 +1,81 @@
...
+  configs: [
+    {
+      bundleIdentifier: "Identifier for Bundle 1",
+      sourcemap: "path/to/sourcemap/for/bundle1",
+    },
+    {
+      bundleIdentifier: "Identifier for Bundle 2",
+      sourcemap: "path/to/sourcemap/for/bundle2",
+    },
+  ],
```
tido64, line 22: - What do we do when a bundle identifier is not listed here?
- If this has to match what is used in the error, can't we just extract the identifiers from there?

tido64 (commented):
Thanks, @SparshaSaha. I've had a very quick look at this and left some comments.

incubator/react-native-error-trace-decorator/src/utils.ts
motiz88: FWIW `metro-symbolicate` has a programmatic API (undocumented at the moment) that you may want to use instead of spawning a process. https://github.com/facebook/metro/blob/main/packages/metro-symbolicate/src/Symbolication.js

Also, if there's any chance that you're calling `metro-symbolicate` on the same source map multiple times in a single run, you should almost certainly be creating and caching a reusable [symbolication contexts](https://github.com/facebook/metro/blob/main/packages/metro-symbolicate/src/Symbolication.js#L830-L841) for each source map, so as to avoid the overhead of parsing the source map multiple times. (The standard source map encoding is optimised for size, not performance.)

incubator/react-native-error-trace-decorator/src/check_arguments_validity.ts
SparshaSaha: @tido64  When a bundle identifier is not listed in the configs, we print that error line as it is.
Actually a bundle Identifier is very specific to how bundles are named.
For that reason, I do not think we will be able to figure out bundles from the error file.

Example errorLine:

at ol (at data/common/rnbundle/900p75yhh4/hermes.android.bundle:1:20)

another error line:
at ol (at data/common/allbundles/mybundle.android.bundle:1:20)

Since error stacks can greatly vary, it might be hard to extract bundleNames from error lines automatically

tido64: In the example you provided, are both pointing to the same bundle or different bundles?

Sorry if you've already thought about this. I don't know anything about how you use this tool. I'm just curious if we could get the bundle name by splitting the lines, for instance `data/common/rnbundle/900p75yhh4/hermes.android.bundle:1:20` becomes `hermes.android.bundle`, then if the user provides a location where all the sourcemaps live, we can just try and look them up. That would simplify the configuration, or even remove the need for a config as it's small enough to be a flag on the command line.

SparshaSaha: Hi @tido64 , in the example I provided, both of these are valid error stacks.

What I mean is, in the first case, `900p75yhh4` is the unique identifier. (we can have multiple bundles named hermes.android.bundle)

In the second case `mybundle.android.bundle` is the unique identifier. 

We do not have a very well-defined structure on how an error line would look like. It would really depend on how the bundles were created and where they are being used.

tido64: Ah, I see. So in the first case, the config might be like below?

```js
    {
      bundleIdentifier: "900p75yhh4",
      sourcemap: "path/to/sourcemap/for/900p75yhh4",
    },
```

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: @tido64 since we are running npx metro-symbolicate, metro-symbolicate will be automatically fetched before running the command if it is not present in the user's system

SparshaSaha: Thanks @motiz88 , I will explore the api. It will save us loads of time, if we don't have to create child processes. :))

tido64: Does `metro-symbolicate` version need to match the Metro version that was used to create the bundles in the first place? Or is this a package that can be used independently?

incubator/react-native-error-trace-decorator/package.json
SparshaSaha: Yeah. I think we can replace it with `console`.
Thanks @tido64

yarn.lock
```diff
@@ -409,12 +409,12 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/parser@npm:^7.0.0, @babel/parser@npm:^7.1.0, @babel/parser@npm:^7.13.16, @babel/parser@npm:^7.14.0, @babel/parser@npm:^7.14.7, @babel/parser@npm:^7.20.13, @babel/parser@npm:^7.20.7":
-  version: 7.20.15
-  resolution: "@babel/parser@npm:7.20.15"
+"@babel/parser@npm:^7.0.0, @babel/parser@npm:^7.1.0, @babel/parser@npm:^7.13.16, @babel/parser@npm:^7.14.0, @babel/parser@npm:^7.14.7, @babel/parser@npm:^7.20.7":
```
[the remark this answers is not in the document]
SparshaSaha: Yeah. I must have done some errors while rebasing! 😅 
Let me fix that

incubator/react-native-error-trace-decorator/package.json
tido64: Sorry, I meant `fs-extra`. All the functions that you use from `fs-extra` also exist in regular `fs`.

tido64: FWIW, `@rnx-kit/console` gives you colors and makes all the tools within rnx-kit consistent.

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: @tido64 sure we can. 😄  
Would you like to suggest something on this regard?

incubator/react-native-error-trace-decorator/src/index.ts
SparshaSaha: That makes sense. Sure I will do accordingly

incubator/react-native-error-trace-decorator/src/utils.ts
tido64: Something like this:

```js
let errors = 0;

if (!currentConfig?.sourcemap || !fs.existsSync(currentConfig.sourcemap)) {
  error(`Sourcemap does not exist: ${currentConfig.sourcemap}`);
  ++errors;
}

if (!currentConfig.bundleIdentifier) {
  error("Bundle identifier must be set");
  ++errors;
}
```

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: As far as I have seen, it can be used independently because it mostly deals with line numbers and sourcemaps but I am not a 100% sure.

incubator/react-native-error-trace-decorator/src/check_arguments_validity.ts
SparshaSaha: Yes. That is correct.

incubator/react-native-error-trace-decorator/package.json
SparshaSaha: I see. Got it.

incubator/react-native-error-trace-decorator/src/extract_and_process_error_stack.ts
SparshaSaha: Using find

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: Replaced with "function"

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: Using for - of everywhere

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: Updated to optional chaining

incubator/react-native-error-trace-decorator/src/extract_and_process_error_stack.ts
SparshaSaha: Changed

incubator/react-native-error-trace-decorator/package.json
SparshaSaha: Updated

incubator/react-native-error-trace-decorator/src/check_arguments_validity.ts
SparshaSaha: Sure. I will update the code to use yargs

SparshaSaha: Using yargs

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: Hi @motiz88 , 
I tried including metro-symbolicate as a dependency but it seems like this package does not really export anything out of it's index.js. 
So I am not being able to import the internal methods. This is the index.js file: https://github.com/facebook/metro/blob/main/packages/metro-symbolicate/src/index.js

We can make some changes in metro-symbolicate to export those packages while preserving its ability to run with npx.
Once that is done, we can use that directly here

incubator/react-native-error-trace-decorator/package.json
SparshaSaha: Replaced fs-extra with fs

yarn.lock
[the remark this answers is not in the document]
SparshaSaha: fixed yarn lock

incubator/react-native-error-trace-decorator/src/index.ts
SparshaSaha: Done

incubator/react-native-error-trace-decorator/src/utils.ts
SparshaSaha: Changed

incubator/react-native-error-trace-decorator/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "eslint": "^8.0.0",
+    "jest": "^27.0.0",
+    "prettier": "^2.8.0",
+    "typescript": "^4.0.0"
+  },
+  "dependencies": {
+    "@rnx-kit/console": "^1.0.11",
+    "metro-symbolicate": "^0.76.0",
+    "source-map": "^0.5.6",
+    "yargs": "^17.7.0"
```
tido64, line 42: We probably need to downgrade this to `^16.0.0` since that's what the rest of the repository is using.

https://github.com/microsoft/rnx-kit/blob/887574d07ab0c59f6d22ff2f93327bea71d125d5/scripts/align-deps-preset.js#L73-L76

tido64 (commented):
This looks good to me. I think there's still potential to improve the UX. For instance, since we control the output, can we bake the bundle identifiers into the source map files? That way we only need the paths to source maps. For now, this is a good first version. There are only a couple of minor things that we need to address to pass CI.

incubator/react-native-error-trace-decorator/package.json
SparshaSaha: Sure. Will do that

SparshaSaha: Changed

tido64:
Can you please add a change log entry? You can do this by running `yarn change`.

tido64 (approved):
Let's get this in 🚀

PR #2260, opened 2023-02-28
[rn-changelog-generator] Add validation command and export

lunaleaps:
### Description

__What issue are you trying to solve?__
 
Currently there are two places for validating a changelog message in a React Native pull request. There is logic that lives here: https://github.com/facebook/react-native/blob/main/packages/react-native-bots/dangerfile.js#L61 and there is also Meta-internal logic. These two validations have gotten slightly out-of-sync.

__How does this change address the issue?__
This change moves the validation logic to this tool. This tool currently handles parsing out the changelog so its logical for it to also provide validation. It is however a bit unfortunately named "generator".  Publishing this new version, I will update the OSS and internal validation to use the `validation` library export.

### Test plan
Added tests for validation util and run `yarn test:local` under `rn-changelog-generator`


incubator/rn-changelog-generator/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rnx-kit/rn-changelog-generator",
-  "version": "0.3.2",
+  "version": "0.4.0",
```
kelset, line 3: you don't need to change this, our tooling will take care of it; revert this and run `yarn change` instead :) see [more here](https://github.com/microsoft/rnx-kit/blob/main/CONTRIBUTING.md#change-logs)

kelset: ps: while you are doing this, can you also change the test script to be `"test:local": "RN_REPO='../../../react-native' rnx-kit-scripts test "`

kelset (changes requested):
overall looks ok, thanks for doing this work! I like the idea of migrating the logic here so that it's a single source of truth.

One minor thing to fix then it's ready to go 👍

incubator/rn-changelog-generator/package.json
lunaleaps: updated!

kelset (approved):
Thanks!

PR #2264, opened 2023-03-06
fix(cli): forward build options to `metro-serializer-esbuild`

tido64:
### Description

Forward build options to `metro-serializer-esbuild`. This allows users to configure what standard to target.

### Test plan

n/a

### DIFF
--- a/packages/cli/src/metro-config.ts
+++ b/packages/cli/src/metro-config.ts
@@ -158,8 +158,12 @@ export function customizeMetroConfig(
   }
 
   if (extraParams.treeShake) {
-    metroConfig.serializer.customSerializer =
-      MetroSerializerEsbuild(metroPlugins);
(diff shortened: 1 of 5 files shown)


mganandraj (approved):

packages/cli/src/metro-config.ts
```diff
@@ -158,8 +158,12 @@ export function customizeMetroConfig(
   }
 
   if (extraParams.treeShake) {
-    metroConfig.serializer.customSerializer =
-      MetroSerializerEsbuild(metroPlugins);
+    metroConfig.serializer.customSerializer = MetroSerializerEsbuild(
```
rasaha91, line 161: Should we check that a customSerializer isn't already set? If so we should probably alert the user that we are overriding theirs?

tido64: Yes, this is a good idea.

rasaha91 (approved):

PR #2325, opened 2023-03-30
docs: bump typescript to 5.0.2

tido64:
### Description

Bump typescript to 5.0.2 and fix build issues.

### Test plan

```
cd docsite
yarn
yarn typecheck
```

### DIFF
--- a/docsite/src/pages/index.tsx
+++ b/docsite/src/pages/index.tsx
@@ -1,22 +1,23 @@
-import React from "react";
-import clsx from "clsx";
-import Layout from "@theme/Layout";
 import Link from "@docusaurus/Link";
 import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
-import styles from "./index.module.css";
+import Layout from "@theme/Layout";
+import clsx from "clsx";
+import React from "react";
 import HomepageFeatures from "../components/HomepageFeatures";
+import styles from "./index.module.css";
 
 function HomepageHeader() {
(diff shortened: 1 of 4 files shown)


docsite/src/pages/index.tsx
```diff
@@ -1,22 +1,23 @@
...
 function HomepageHeader() {
   const { siteConfig } = useDocusaurusContext();
+  const { title1, title2 } = siteConfig.customFields;
   return (
     <header className={styles.headerBanner}>
       <div className="container">
         <h1 className={clsx("hero__title", styles.headerTitle)}>
           <div className={clsx(styles.headerTitleText)}>
-            {siteConfig.customFields.title1}
+            {typeof title1 === "string" ? title1 : "React Native"}
```
SparshaSaha, line 17: Will it make sense to memoize this part using React.useMemo?

SparshaSaha (approved):

docsite/src/pages/index.tsx
tido64: Strings are immutable and are always ref comparable. This will never trigger a re-render unless the value changes. Even so, `title1` and `title2` are always defined:

https://github.com/microsoft/rnx-kit/blob/72456638f0ce5b76feed8ab0b3a9c1e399e1d549/docsite/docusaurus.config.js#L27-L28

The reason we do this check here is because `tsc` doesn't know that.

PR #2327, opened 2023-04-01
fix(cli): add --id flag to rnx-start

AAHAbbas:
### Description
The default bundle configuration (first configuration) is used by `rnx-start` when the server configuration is absent. 
Adding `--id` flag to `rnx-start` which enables specifying which bundle configuration to use. 

### Test plan
Remove the server configuration and modify `start.ts` to verify which configuration is loaded:

```diff
diff --git a/packages/cli/src/start.ts b/packages/cli/src/start.ts
index 61b73e29..4221bff3 100644
--- a/packages/cli/src/start.ts
+++ b/packages/cli/src/start.ts
@@ -74,6 +74,7 @@ export async function rnxStart(
   cliOptions: CLIStartOptions
 ): Promise<void> {
   const serverConfig = getKitServerConfig(cliOptions);
+  console.log(serverConfig);

   const { createDevServerMiddleware, indexPageMiddleware } = friendlyRequire<
     typeof CliServerApi
diff --git a/packages/test-app/package.json b/packages/test-app/package.json
index 196b9672..0b2ea748 100644
--- a/packages/test-app/package.json
+++ b/packages/test-app/package.json
@@ -155,21 +155,6 @@
         }
       }
     ],
-    "server": {
-      "projectRoot": "src",
-      "plugins": [
-        "@rnx-kit/metro-plugin-cyclic-dependencies-detector",
-        [
-          "@rnx-kit/metro-plugin-duplicates-checker",
-          {
-            "ignoredModules": [
-              "react-is"
-            ]
-          }
-        ],
-        "@rnx-kit/metro-plugin-typescript"
-      ]
-    },
     "alignDeps": {
       "presets": [
         "microsoft/react-native",
```

Build and run Metro server:

```
cd packages/test-app
yarn build --dependencies
yarn start --id esbuild
```

### DIFF
--- a/packages/cli/react-native.config.js
+++ b/packages/cli/react-native.config.js
@@ -98,6 +98,10 @@ module.exports = {
           name: "--no-interactive",
           description: "Disables interactive mode.",
         },
+        {
+          name: "--id [string]",
+          description: "Specify which bundle configuration to use.",
+        },
       ],
     },
     rnxCopyAssetsCommand,
--- a/.changeset/silly-penguins-applaud.md
+++ b/.changeset/silly-penguins-applaud.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/cli": patch
+---
+
+Add `--id` flag to `rnx-start`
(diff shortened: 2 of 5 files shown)


packages/cli/react-native.config.js
```diff
@@ -98,6 +98,10 @@ module.exports = {
           name: "--no-interactive",
           description: "Disables interactive mode.",
         },
+        {
+          name: "--id [string]",
+          description: "Specify which bundle configuration to use.",
```
tido64, line 103: I think we should be clearer on the intent of this flag:
```suggestion
          description: "Specify which bundle configuration to use if server configuration is missing.",
```

The entry in README should also be updated.

tido64 (approved):
Thanks for adding this. I left a small comment.

tido64 (approved):

PR #2334, opened 2023-04-10
fix(metro-service): avoid spinning up the file watcher

### LINKED ISSUE #2332: rnx-bundle runs the metro Server without "watch:false"
The command ends up either in @react-native-community/cli-plugin-metro buildBundleWithConfig or in @rnx-kit/metro-service calling new Server(config).

When doing a single bundle command it should use new Server(config, {watch:false}) to prevent the extra work of spinning up the jest-haste-map file watcher.

tido64:
### Description

Avoid spinning up the file watcher when bundling.

Resolves #2332.

### Test plan

n/a

### DIFF
--- a/.changeset/polite-needles-tell.md
+++ b/.changeset/polite-needles-tell.md
@@ -1,0 +1,6 @@
+---
+"@rnx-kit/metro-service": patch
+"@rnx-kit/cli": patch
+---
+
+Avoid spinning up the file watcher when bundling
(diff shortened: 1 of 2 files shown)


acoates-ms:
In OMR at least we were hitting this case at the top of the function, so this change would not be sufficient.
```js
  const buildBundleWithConfig = getBundlerFromCliPlugin();
  if (buildBundleWithConfig) {
    return buildBundleWithConfig(args, config, output);
  }
```

tido64:
Closing as this fix doesn't apply to newer versions of the CLI, and I don't have time to do a proper testing.

PR #2335, opened 2023-04-10
fix(metro-config): apply `@react-native/metro-config` if available

tido64:
### Description

Get default config from `@react-native/metro-config` if available.

This will be required react-native-test-app to support 0.72: https://github.com/microsoft/react-native-test-app/issues/1359

### Test plan

Make sure configs are correctly merged:

```
cd packages/test-app
yarn build --dependencies
node -e "console.log(require('./metro.config.js'))"
```


kelset (approved):
LGTM

PR #2363, opened 2023-04-14
fix: fix multiple Kotlin Gradle plugin loads warning

tido64:
### Description

Fix the "Kotlin Gradle plugin was loaded multiple times" warning.

### Test plan

```
cd packages/test-app/android
./gradlew assembleDebug
```


kelset (approved):
yup, tested locally, all works 👍

PR #2374, opened 2023-04-18
fix(eslint-plugin): ignore namespace exports

### LINKED ISSUE #2373: eslint-plugin: incorrect expansion of namespace exports in fixer
### What happened?

Currently the `no-export-all` fixer has incorrect expansion of namespace exports: it will export all the individual identifiers instead of the namespace.

Consider this scenario:

```ts
// utils/foo.ts
export const bar = 'bar';
export type Baz = {};

// utils/index.ts
export * as foo from './foo'; // ❌ lint error!

// bar.ts
import { foo } from './utils';
const bar: foo.Bar = {};
console.log(foo.bar);
```

The `no-export-all` fixer will fix the lint error in `my-pkg/index.ts` as follows, which breaks the consuming code:
```ts
// utils/index.ts
export { bar } from './foo';
export type { Baz } from './foo';

// bar.ts
import { foo } from 'my-pkg'; // ❌ error, foo does not exist!
const bar: foo.Bar = {};
console.log(foo.bar);
```

Correct output (I think):
```ts
// utils/index.ts
import * as foo from './foo';
export { foo };
```

### Affected Package

@rnx-kit/eslint-plugin

### Version

0.4.1

n/a

### System Information

n/a

### Steps to Reproduce

see above

### Code of Conduct

ecraig12345:
### Description

Currently the `no-export-all` fixer has incorrect expansion of namespace exports (`export * as foo from './foo'`): it will export all the individual identifiers instead of the namespace, which breaks consuming code. More details in #2373.

~~This PR adds a different fix specific to this case. New example code:~~
```ts
// import * as foo from './foo';
// export { foo };
```
After discussion with @tido64, we decided to update the rule to ignore namespace exports.

Fixes #2373

### Test plan

- Added a test

### DIFF
--- a/.changeset/bright-clocks-switch.md
+++ b/.changeset/bright-clocks-switch.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/eslint-plugin": patch
+---
+
+no-export-all: Don't flag or autofix namespace exports (`export * as ns`)
--- a/packages/eslint-plugin/src/rules/no-export-all.js
+++ b/packages/eslint-plugin/src/rules/no-export-all.js
@@ -4,6 +4,7 @@
 /**
  * @typedef {import("@typescript-eslint/types").TSESTree.Node} Node
  * @typedef {import("eslint").Rule.RuleContext} ESLintRuleContext
+ * @typedef {import("eslint").Rule.ReportFixer} ESLintReportFixer
  * @typedef {{ exports: string[], types: string[] }} NamedExports
  *
  * @typedef {{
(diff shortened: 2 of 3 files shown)


.changeset/bright-clocks-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rnx-kit/eslint-plugin": patch
+---
+
+no-export-all: Don't autofix namespace exports (`export * as ns`)
```
ecraig12345, line 5:
```suggestion
no-export-all: Correctly fix namespace exports (`export * as ns`)
```

tido64 (approved):
Thanks for the fix ❤️

PR #2376, opened 2023-04-18
fix(eslint-plugin): add support for TS "export import ="

### LINKED ISSUE #2375: eslint-plugin: no-export-all misses exports defined with TS "export import ="
### What happened?

In TypeScript, it's possible to declare exports with the syntax `export import Foo = /*name or require*/`. The fixer for `no-export-all` currently misses these exports, causing erroneous code to be generated.

```ts
// common.ts
export type Foo = {};
export const bar = 'bar';

// stuff.ts (demonstrating various valid syntax...not whether it makes sense)
import * as common from './common';
export import utils = common;
export import Foo = common.Foo;
export import bar = common.bar;
export import common = require('./common');
namespace Baz {
  export const baz = 'baz';
}
export import baz = Baz.baz;
export const other = 'other';

// index.ts
export * from './stuff';
```

Current output for `index.ts`:
```ts
export { other } from './stuff';
```

Expected output (roughly):
```ts
export { utils, Foo, bar, common, baz, other }'
```

This is based on real code, and there are enough instances of it that I can't just remove the usage (at least not without another lint rule banning this usually-unnecessary syntax). With this syntax, it's not possible to determine whether the referenced identifier is actually a type or a value without additional complexity.

### Affected Package

@rnx-kit/eslint-plugin

### Version

0.4.1

n/a

### System Information

n/a

### Steps to Reproduce

see above

### Code of Conduct

ecraig12345:
### Description

In TypeScript, it's possible to declare exports with the syntax `export import Foo = /*name or require()*/`. The fixer for `no-export-all` currently misses these exports, causing erroneous code to be generated. See #2375 for more details.

This PR adds support for detecting these kinds of exports. Currently it assumes that these are always value exports, due to the additional complexity required to figure out the actual type.

Fixes #2375 

### Test plan

- Added a test
- Locally link the rule to a repo with this type of export and verify the correct fix is applied.

### DIFF
--- a/.changeset/fifty-numbers-complain.md
+++ b/.changeset/fifty-numbers-complain.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/eslint-plugin": patch
+---
+
+no-export-all: add support in fixer for TS `export import =`
--- a/packages/eslint-plugin/src/rules/no-export-all.js
+++ b/packages/eslint-plugin/src/rules/no-export-all.js
@@ -330,6 +330,14 @@ function extractExports(context, moduleId, depth) {
             }
             break;
           }
+
+          case "TSImportEqualsDeclaration":
+            if (node.isExport) {
+              // export import foo = require('./foo');
+              // export import Bar = Foo.Bar;
+              exports.add(node.id.name);
+            }
+            break;
         }
       },
     });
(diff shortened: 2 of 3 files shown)


tido64 (approved):

tido64:
Thank you ❤️

PR #2381, opened 2023-04-18
chore: remove `metro-swc-worker`

tido64:
### Description

Metro is experimenting with [SWC](https://swc.rs/) here: https://github.com/facebook/metro/pull/948

### Test plan

n/a


kelset (approved):
o7

PR #2402, opened 2023-05-08
fix(cli): add Kotlin plugin to `copy-assets` command

tido64:
### Description

Add Kotlin plugin to `copy-assets` command.

### Test plan

Tested internally.

### DIFF
--- a/.changeset/moody-seahorses-agree.md
+++ b/.changeset/moody-seahorses-agree.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/cli": patch
+---
+
+Add Kotlin plugin to `copy-assets` command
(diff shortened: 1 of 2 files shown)


tido64:
@sveinar: Can you check whether this works for you?

sveinar:
> @sveinar: Can you check whether this works for you?

Thanks, I tried but I ran into a problem of the form:
```
e: /Users/sveinar/.gradle/caches/transforms-3/e034d85dc48167c9d68b935151fa7783/transformed/fragment-1.5.4-api.jar!/META-INF/fragment_release.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.6.0, expected version is 1.4.0.
```

kelset (approved):

PR #2461, opened 2023-06-14
RuntimeScheduler is only used by TurboModule Manager in >RN0.72

shwanton:
### Description

- When I was testing loading AsyncStorage as a [TurboModule for macOS](https://github.com/j-piasecki/react-native-async-storage/commit/6d2faaaab9770ef54bb4d1175d24f30a21bc03d8), I ran into this build error:
`No matching function for call to 'RCTAppSetupDefaultJsExecutorFactory'`
CleanShot 2023-06-14 at 09 33 44

- `RuntimeScheduler` was added as a param to `RCTAppSetupDefaultJsExecutorFactory` [in RN Core here](https://github.com/facebook/react-native/pull/36209)
https://github.com/facebook/react-native/pull/36209/files#diff-1e16648bc876095f7583276815d1964d165986e1f5174ff23123c8c027a2ef0bR44

- We could pick that change, or just remove the param till it's merged.

### Test plan

- Not sure how to test this inside RNXKit, however here is the AsyncStorage TM working on macOS with this fix:

https://github.com/microsoft/rnx-kit/assets/96719/2dfe7e90-2047-4a92-8a8c-a4feebb748bc

### DIFF
--- a/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
+++ b/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
@@ -31,12 +31,7 @@
 #define USE_RUNTIME_SCHEDULER 0
 #endif  // __has_include(<React-RCTAppDelegate/RCTLegacyInteropComponents.h>)
 
-// RCTAppSetupDefaultJsExecutorFactory is in different locations for iOS (0.71)/macOS(0.71)
-#if !TARGET_OS_OSX
-#import <React/RCTAppSetupUtils.h>
-#else
-#import <React-RCTAppDelegate/RCTAppSetupUtils.h>
-#endif // TARGET_OS_OSX
+#endif  // __has_include(<React/RCTAppSetupUtils.h>)
 
 #endif  // USE_TURBOMODULE
 
--- a/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
+++ b/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
@@ -45,7 +45,7 @@
     _turboModuleManager = [[RCTTurboModuleManager alloc] initWithBridge:bridge
                                                                delegate:self
                                                               jsInvoker:callInvoker];
-    return RCTAppSetupDefaultJsExecutorFactory(bridge, _turboModuleManager, _runtimeScheduler);
+    return RCTAppSetupDefaultJsExecutorFactory(bridge, _turboModuleManager);
 #else
     _turboModuleManager = [[RCTTurboModuleManager alloc] initWithBridge:bridge
                                                                delegate:self
--- a/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
+++ b/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
@@ -16,10 +16,27 @@
 #else
 #import <React-RCTAppDelegate/RCTAppSetupUtils.h>
 #import <React/RCTSurfacePresenterBridgeAdapter.h>
+
+// We still get into this path because react-native-macos 0.71 picked up some
+// 0.72 bits. AFAICT, `RCTLegacyInteropComponents.h` is a new addition in 0.72
+// in both react-native and react-native-macos.
+#if __has_include(<React-RCTAppDelegate/RCTLegacyInteropComponents.h>)
 #import <react/renderer/runtimescheduler/RuntimeScheduler.h>
 #import <react/renderer/runtimescheduler/RuntimeSchedulerCallInvoker.h>
+#if __has_include(<React/RCTRuntimeExecutorFromBridge.h>)
+#import <React/RCTRuntimeExecutorFromBridge.h>
+#endif  // __has_include(<React/RCTRuntimeExecutorFromBridge.h>)
 #define USE_RUNTIME_SCHEDULER 1
-#endif  // __has_include(<React/RCTAppSetupUtils.h>)
+#else
+#define USE_RUNTIME_SCHEDULER 0
+#endif  // __has_include(<React-RCTAppDelegate/RCTLegacyInteropComponents.h>)
+
+// RCTAppSetupDefaultJsExecutorFactory is in different locations for iOS (0.71)/macOS(0.71)
+#if !TARGET_OS_OSX
+#import <React/RCTAppSetupUtils.h>
+#else
+#import <React-RCTAppDelegate/RCTAppSetupUtils.h>
+#endif // TARGET_OS_OSX
 
 #endif  // USE_TURBOMODULE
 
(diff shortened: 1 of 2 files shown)


packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
```diff
@@ -45,7 +45,7 @@ @implementation RNXTurboModuleAdapter {
     _turboModuleManager = [[RCTTurboModuleManager alloc] initWithBridge:bridge
                                                                delegate:self
                                                               jsInvoker:callInvoker];
-    return RCTAppSetupDefaultJsExecutorFactory(bridge, _turboModuleManager, _runtimeScheduler);
+    return RCTAppSetupDefaultJsExecutorFactory(bridge, _turboModuleManager);
```
tido64, line 48: Would something like this work?
```suggestion
#if __has_include(<React/RCTAppSetupUtils.h>)  // <0.72
    return RCTAppSetupDefaultJsExecutorFactory(bridge, _turboModuleManager);
#else
    return RCTAppSetupDefaultJsExecutorFactory(bridge, _turboModuleManager, _runtimeScheduler);
#endif  // __has_include(<React/RCTAppSetupUtils.h>)
```

tido64 (commented):
Thanks for fixing New Arch on 0.71! I thought we had tested this but apparently not 😅

packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
shwanton: This still has the build issue since the `RCTAppSetupDefaultJsExecutorFactory` function signature in `RCTAppSetupUtils` doesn't have `RuntimeScheduler` param.

https://github.com/cipolleschi/react-native/blob/e03455a5d376ebdb8947e18fb7ef7687f88c819b/Libraries/AppDelegate/RCTAppSetupUtils.h#L44

tido64: This still fails? I had a look at the files and from what I can see, `RuntimeScheduler` got added starting 0.72:

- 0.71: https://github.com/facebook/react-native/blob/0.71-stable/React/AppSetup/RCTAppSetupUtils.h#L38
- 0.72: https://github.com/facebook/react-native/blob/0.72-stable/packages/react-native/Libraries/AppDelegate/RCTAppSetupUtils.h#L45

So this `#ifdef` should work? In 0.72, `React/RCTAppSetupUtils.h` no longer exists. It's `React-RCTAppDelegate/RCTAppSetupUtils.h` instead.

shwanton: `USE_RUNTIME_SCHEDULER` already encapsulates that logic in order to enter this block and use `RuntimeScheduler`
https://github.com/microsoft/rnx-kit/blob/058e10b1c16f0e5c4bf4e149a42b5b882848ec78/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm#L13-L22

iOS: 
On 0.71,`USE_RUNTIME_SCHEDULER` is false because it still has `<React/RCTAppSetupUtils>`.
On 0.72, this will flip to `true` and use the Runtime Scheduler.

macOS:
On 0.71 for macOS, `RCTAppSetupUtils` is already in `RCTAppDelegate/RCTAppSetupUtils.h`
I believe this got picked back from 0.72 😬
https://github.com/microsoft/react-native-macos/blob/9b098f5caef423742d2cf39f480c281d17960572/Libraries/AppDelegate/RCTAppSetupUtils.h#L36-L39

CleanShot 2023-06-14 at 14 22 04

I'm not sure of a clean way to support 0.71 & 0.72 at the same time for each platform.
Can we just fix for macOS now (since iOS doesn't see this path) & then fix both when we upgrade to 0.72?

tido64: Can we use this instead?

```diff
diff --git a/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm b/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
index 1fcb2d7b..de2166ed 100644
--- a/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
+++ b/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
@@ -16,9 +16,18 @@
 #else
 #import <React-RCTAppDelegate/RCTAppSetupUtils.h>
 #import <React/RCTSurfacePresenterBridgeAdapter.h>
+
+// We still get into this path because react-native-macos 0.71 picked up some
+// 0.72 bits. AFAICT, `RCTLegacyInteropComponents.h` is a new addition in 0.72
+// in both react-native and react-native-macos.
+#if __has_include(<React-RCTAppDelegate/RCTLegacyInteropComponents.h>)
 #import <react/renderer/runtimescheduler/RuntimeScheduler.h>
 #import <react/renderer/runtimescheduler/RuntimeSchedulerCallInvoker.h>
 #define USE_RUNTIME_SCHEDULER 1
+#else
+#define USE_RUNTIME_SCHEDULER 0
+#endif  // __has_include(<React-RCTAppDelegate/RCTLegacyInteropComponents.h>)
+
 #endif  // __has_include(<React/RCTAppSetupUtils.h>)

 #endif  // USE_TURBOMODULE
```

shwanton: This solves the issue with passing `_runtimeScheduler` to the `RCTAppSetupDefaultJsExecutorFactory`, however the import is different on both platforms.

I had to add this additional line, which will break on 0.72 😅
```
RCTAppSetupDefaultJsExecutorFactory

File: packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm [hunk 2]
```diff
@@ -16,10 +16,27 @@
...
+#else
+#define USE_RUNTIME_SCHEDULER 0
+#endif  // __has_include(<React-RCTAppDelegate/RCTLegacyInteropComponents.h>)
+
+// RCTAppSetupDefaultJsExecutorFactory is in different locations for iOS (0.71)/macOS(0.71)
+#if !TARGET_OS_OSX
+#import <React/RCTAppSetupUtils.h>
+#else
+#import <React-RCTAppDelegate/RCTAppSetupUtils.h>
+#endif // TARGET_OS_OSX
```
tido64, line 39: I'm confused about this addition. Isn't this already gated by the `#ifdef` on line 13?

https://github.com/microsoft/rnx-kit/blob/2c08b5c156883f03e5d8e768fcbff025dbae83d7/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm#L13

shwanton: Oops, my bad, my patch was faulty. Your change works on both iOS/macOS.

tido64 (approved):
Thanks for taking the time to fix this ❤️

tido64:
image

Can you run `yarn change` at project root and add an appropriate change log entry?

PR #2467, opened 2023-06-16
fix(metro-config): fix out-of-tree platforms not being included on react-native 0.72

### LINKED ISSUE #2466: Update needed for RN 0.72
### What happened?

TypeError: getDefaultConfig is not a function
    at loadMetroConfig (D:\2Office\Build\JsSrc\sdx_privacysettingsdialog_x64_debug\node_modules\@rnx-kit\metro-service\lib\config.js:133:27)
    at Object.rnxBundle [as func] (D:\2Office\Build\JsSrc\sdx_privacysettingsdialog_x64_debug\node_modules\@rnx-kit\cli\lib\bundle.js:11:67)

### Affected Package

@rnx-kit/metro-service

### Version

3.0.4

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### System Information

```true
<blank>
```

### Steps to Reproduce

`getDefaultConfig` is no longer exported from `@react-native-community/cli-plugin-metro`.  See https://github.com/react-native-community/cli/pull/1875

### Code of Conduct

tido64:
### Description

Fix compatibility with react-native 0.72

Resolves #2466.

### Test plan

n/a

### DIFF
--- a/packages/metro-service/src/config.ts
+++ b/packages/metro-service/src/config.ts
@@ -160,10 +160,15 @@ function getDefaultConfigProvider(): typeof getDefaultConfigInternal {
     const {
       getDefaultConfig,
     } = require("@react-native-community/cli-plugin-metro");
-    return getDefaultConfig;
+
+    // Starting with `react-native` 0.72, we need to build a complete Metro
+    // config upfront and will no longer need to get a default config here.
+    return getDefaultConfig ? getDefaultConfig : () => ({});
   } catch (_) {
-    return getDefaultConfigInternal;
+    // Ignore
   }
+
+  return getDefaultConfigInternal;
 }
 
 /**
--- a/.changeset/real-rules-speak.md
+++ b/.changeset/real-rules-speak.md
@@ -1,0 +1,6 @@
+---
+"@rnx-kit/metro-config": patch
+"@rnx-kit/metro-service": patch
+---
+
+Fix out-of-tree platforms not being included on react-native 0.72
--- a/packages/metro-config/package.json
+++ b/packages/metro-config/package.json
@@ -23,17 +23,22 @@
     "@rnx-kit/babel-preset-metro-react-native": "^1.1.4",
     "@rnx-kit/console": "^1.0.0",
     "@rnx-kit/tools-node": "^2.0.0",
+    "@rnx-kit/tools-react-native": "^1.3.1",
     "@rnx-kit/tools-workspaces": "^0.1.3"
   },
   "peerDependencies": {
     "@react-native/metro-config": "*",
     "metro-config": ">=0.58.0",
     "metro-react-native-babel-preset": "*",
+    "metro-resolver": ">=0.58.0",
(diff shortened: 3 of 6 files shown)


acoates-ms (approved):

packages/metro-service/src/config.ts
```diff
@@ -160,10 +160,23 @@ function getDefaultConfigProvider(): typeof getDefaultConfigInternal {
...
+    if (getDefaultConfig) {
+      return getDefaultConfig;
+    }
   } catch (_) {
-    return getDefaultConfigInternal;
+    // Ignore
   }
+
+  try {
+    const { getDefaultConfig } = require("@react-native/metro-config");
```
acoates-ms, line 171: Should @react-native/metro-config be added as a required package to one of the align-deps profile capabilities?

acoates-ms:
```
error Invalid platform "windows" selected.
info Available platforms are: "android", "ios". If you are trying to bundle for an out-of-tree platform, it may not be installed.
```

I get the following error due to the new default config not picking up the remaining part of the metro-config which was previously in the getDefaultConfig from the CLI. https://github.com/react-native-community/cli/blob/1796cdd794979dcf37ba5c7ec4303154bb71d56c/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts#LL44C5-L44C5 -- I'm sure we need the other parts of this config too.

kelset (approved):
LGTM; let's also do what Andrew says in the profile (which I think you might have already done by now)

tido64:
> LGTM; let's also do what Andrew says in the profile (which I think you might have already done by now)

I've added a new capability:
image

Still looking into the invalid platform `windows` error. I get no complaints in react-native-test-app. But it fails in rnx-kit.

tido64:
@acoates-ms: My testing was flawed, so I didn't repro the issue correctly. The issue was that `@rnx-kit/metro-config` overrides `resolver.platforms` regardless of what we set earlier. Anyway, with the new changes, we should no longer call `getDefaultConfig` from `metro-service`. We should instead expect a full config from the user. This means that we have to fix `@rnx-kit/metro-config` so that it correctly handles out-of-tree platforms.

@kelset: This also means that `@react-native/metro-config` as is will break react-native-macos/-windows. Any suggestions for how we should approach here? Fixing upstream seems like the right thing to do, but it adds complexity as you can see in this PR.

tido64:
@acoates-ms: In any case, I think this is as correct as I can make it. Can you verify on your side that it works?

acoates-ms:
> @acoates-ms: In any case, I think this is as correct as I can make it. Can you verify on your side that it works?

Yup, I was able to bundle with these changes.

packages/metro-service/src/config.ts
acoates-ms: Same question, but now should it be "metro-resolver" added to the metro-config capability for 0.72?

PR #2480, opened 2023-06-25
Fix metro 0.72 sourcemap options

renchap:
### Description

When using the esbuild serializer with RN 0.72 (Metro 0.76.5), the build crashes with this error:
```
✘ [ERROR] options.shouldAddToIgnoreList is not a function [plugin @rnx-kit/metro-serializer-esbuild]

    ../../node_modules/metro/src/DeltaBundler/Serializers/helpers/getSourceMapInfo.js:18:23:
      18 │     isIgnored: options.shouldAddToIgnoreList(module),
```

`shouldAddToIgnoreList` has been added in https://github.com/facebook/metro/commit/82bd64a9720174a9e2a02fb73bbef292153e20f1#diff-6a12274bc59498289ea08e6f204df8fe5e262338a7c77344463497df264440c1R175 but does not have a default or a fallback, so Metro crashes as we dont provide it.

### Test plan

I tested that it fixes the production build on my app after upgrading to RN 0.72

### DIFF
--- a/.changeset/ten-keys-hope.md
+++ b/.changeset/ten-keys-hope.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-serializer-esbuild": patch
+---
+
+Add `shouldAddToIgnoreList` to `sourceMappingOptions`, this became mandatory in Metro 0.76.3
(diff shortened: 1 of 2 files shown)


renchap:

renchap:

tido64 (approved):
Weird that I haven't seen this, even when I was testing this last night. In any case, thanks for fixing it ❤️

PR #2504, opened 2023-07-04
RFC: React Native WebAPIs

tido64:
### Abstract

React Native currently lacks a well-defined, stable and complete API surface. Compared to what both [Android](https://developer.android.com/reference) and [iOS](https://developer.apple.com/documentation/technologies) provide out of the box, React Native core is missing quite a lot today. To fill that gap, we have a "wild west" of community modules, each with its own set of interfaces and behaviors. While this is in line with the broader experience in the web/npm space, at the end of the day this means that it is the developers' responsibility to find modules that fit their needs and that are seemingly actively maintained.

Additionally, the APIs are not compatible with Web APIs, thus closing the door to a wealth of open source libraries that do not have explicit React Native support. This often means that developers cannot reuse existing web code, and must search for or even create one for their needs.

In this RFC, we are proposing to close this gap by providing our own implementation of the [Web APIs](https://developer.mozilla.org/en-US/docs/Web/API) for React Native. The ultimate goal is to open the possibility to run non-UI code directly in a React Native app; to provide a familiar environment for existing web developers as well as a well-documented API for developers of any experience level.

[Click here to read the rendered version](https://github.com/microsoft/rnx-kit/blob/tido/react-native-standard-api/text/0002-react-native-webapis.md)

---

This RFC is meant to be complementary to [React DOM for Native](https://github.com/react-native-community/discussions-and-proposals/pull/496)


text/0002-react-native-standard-api.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Infrastructure and repository setup
+
+On top of what has been mentioned above, we are still investigating the right
+approach for how the code for this effort should be created and organised. The
+most likely approach will involve a monorepo (similar to
+[`rnx-kit`](https://github.com/microsoft/rnx-kit)) where each module will be its
+own dedicated package. For now, we will suggest that implementations for all
+supported platforms be present in the one package. We won't have
+`/packages/battery-manager-ios` and `/packages/battery-manager-android`, only
```
acoates-ms, line 125: I think we're going to have to support multiple options here.  It is unlikely that all packages would support all platforms to start with, and similar to native modules, different module owners are likely to have varying bandwidth and skills to support all platforms.  Also what does "all platforms" mean?  Does it include windows, macos?  tvos?  skia?  etc?

text/0002-react-native-webapi.md
```diff
@@ -1,4 +1,4 @@
-- Title: react-native-standard-api
+- Title: react-native-webapi
```
tido64, line 1: We should mirror the web name:
```suggestion
- Title: react-native-webapis
```

text/0002-react-native-webapi.md
```diff
@@ -53,8 +53,8 @@ Our goal is not to reimplement a browser, e.g.
 a full browser with everything that entails, including modules that they may
 never get used. Native apps, and mobile apps especially, cannot be shipped with
 unused bits; nor does it make any sense to include MBs of dependencies that are
-never used. Ideally, migrating from community modules to the standard API should
-not increase the final app size (at least not significantly).
+never used. Ideally, migrating from community modules to the WebAPI should not
```
tido64, line 56:
```suggestion
never used. Ideally, migrating from community modules to the WebAPIs should not
```

text/0002-react-native-webapi.md
```diff
@@ -89,9 +89,9 @@ graph TD;
 ### Modularity
 
 We want to avoid introducing unused modules and adding unnecessary bloat to the
-app bundle. The standard API must therefore be broken down into smaller modules
-that can be installed separately. These modules are installed by autolinking,
-and must therefore be explicitly added to an app's `package.json`.
+app bundle. The WebAPIs must therefore be broken down into smaller modules that
+can be installed separately. These modules are installed by autolinking, and
+must therefore be explicitly added to an app's `package.json`.
```
tido64, line 94:
```suggestion
app bundle. WebAPIs must therefore be broken down into smaller modules that can
be installed separately. These modules are installed by autolinking, and must
therefore be explicitly added to an app's `package.json`.
```

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,221 @@
...
+module.exports = {
+  dependency: {
+    api: {
+      polyfill: "./polyfill.js",
+    },
+  },
+};
+```
+
+Polyfills are gathered and passed to Metro via [`serializer.getPolyfills`][].
```
tido64, line 115: We should make it clear that any package with a valid declaration will be autolinked, not just packages under the `@react-native-webapis` scope.

NickGerleman (commented):
This overall proposal makes a lot of sense.

Inside of React native, a similar sort of polyfill organization (registered differently) was used for `Performance` APIs. 
1. https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/WebPerformance/Performance.js
1. https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Core/setUpPerformance.js
1. https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Core/InitializeCore.js#L31

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Infrastructure and repository setup
+
+On top of what has been mentioned above, we are still investigating the right
+approach for how the code for this effort should be created and organised. The
+most likely approach will involve a monorepo (similar to
+[`rnx-kit`](https://github.com/microsoft/rnx-kit)) where each module will be its
```
necolas, line 122: I think it's important for this to live in the `react-native` repo and not be another 3rd party project that people have to go looking around for to find modules. My position is that the DOM/Web APIs is the long-term future of the maintained RN JS APIs, and it's critical that they be hosted and maintained out of the main monorepo infra.

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,221 @@
...
+dependencies they need to add. At minimum we should:
+
+1. Implement a tool for detecting usage of web APIs
+   - The tool should be able to list used web APIs and flag uses that have not
+     been polyfilled.
+   - If possible, the tool should also recommend which dependencies to add
+     and/or automatically add it to `package.json`.
+   - **Note:** While we say "tool" here, it doesn't necessarily have to be a
+     standalone thing. A Babel plugin or similar would also fit. The less users
+     have to worry about it, the better.
```
necolas, line 147: Agree that having auto-magic setup would be a great DX, and an essential part of opening up RN to developers coming from Web in the future.

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,221 @@
...
+  implies that we have worked out what modules to prioritize by this stage.
+
+- **Run stage:** By this stage, we aim to have a good level of confidence on the
+  viability of this proposal. Documentation on how to use this new set of
+  modules will be prepared to help developers leverage them to bring their web
+  code to native platforms.
+
+## Unresolved questions
+
+- Which parts of the Web API do we prioritize first?
```
necolas, line 203: We've been implementing MutationObserver and IntersectionObserver. Some Web APIs like those will come up against differences in the execution model of React Native vs Web, and we're working on an RFC to address that as well.

necolas:
Exciting to see this proposal and how complementary it is to other API convergence efforts. I appreciate the attention given to ideas for native app tooling as well. What are some of your next steps? Are you already prototyping the battery API using this approach?

text/0002-react-native-webapis.md
tido64: After having tried `getModulesRunBeforeMainModule`, I've realized that we actually need this: https://github.com/facebook/metro/issues/850

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,221 @@
...
+## Adoption strategy
+
+We will be following the crawl-walk-run methodology:
+
+- **Crawl stage:** Only one module will be implemented so that the tooling and
+  infrastructure can be built and flows can be set up. The infrastructure should
+  allow for multiple teams and developers to work discretely on the specific
+  modules that they are interested in, removing most of the friction to enable
+  contributors to just focus on producing code. We want to evaluate whether
+  things up to this point still make sense and adapt if otherwise.
```
kelset, line 190: During this phase we should also figure out how to create a stable testing infra.

About this, an interesting angle is that there's already a fully suite of JS [testing from W3C](https://www.w3.org/QA/Test/) that we could leverage potentially. Here's an example, [Geolocation](https://dev.w3.org/geo/api/test-suite/), and the single test: https://dev.w3.org/geo/api/test-suite/t00001.js

it'd be great if we could figure out how to directly reuse them

text/0002-react-native-webapis.md
necolas: Depends how long "experimental" goes on for. We're currently greatly slowed by a lot of past forking and out-of-repo initiatives, and have long and complicated projects to defork those things. That needs to be avoided with new work.

We also have existing Web APIs that we've implemented and they're not separate repos or projects either. They'll be highly dependent on new internals like an event loop model and native extension points, so being out of core will make it harder for Web APIs with those needs to be tested and inform development.

Curious to hear @rubennorte's thoughts

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,226 @@
...
+
+- **Web APIs:** On the surface, there are little to no differences from typical
+  web code.
+
+- **JS shim:** This is a thin layer for marshalling web API calls to native
+  modules. It needs to be installed before it can be used. The current thinking
+  is to install this layer as polyfills. Installing them on-demand is preferred,
+  but we need to start somewhere and we should be able to improve this layer
+  later without affecting users. In any case, we will need tools to include
+  polyfills similarly to [autolinking][].
```
kelset, line 76: quicknote, draft implementation here: https://github.com/microsoft/rnx-kit/pull/2580/files

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,226 @@
...
+  - As proof-of-concept, we've implemented the
+    [`Battery Status API`](https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API)
+    as it is small and self-contained.
+    - https://github.com/microsoft/rnx-kit/pull/2590
+    - Note that we're currently using
+      [`serializer.getModulesRunBeforeMainModule`][] until something better
+      exists. It currently requires that listed modules are explicitly imported
+      in the bundle itself, which we do not want. We need something akin to
+      Webpack's [entry points](https://webpack.js.org/concepts/entry-points/).
+      More details here: https://github.com/facebook/metro/issues/850.
```
kelset, line 215: another note that might be worth having in this section for this first initial phase: we'll keep all the packages marked as private so that we won't be publishing them to npm quite yet. This mean that for now to make them work you'll need to be in the rnx-kit monorepo.

text/0002-react-native-webapis.md
kelset: usually the related test suite seems to be linked in this page, at the top -> https://w3c.github.io/battery/#the-batterymanager-interface

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,227 @@
...
+  for all the platforms will take a massive amount of time and funding —
+  realistically, we will select a subset of APIs to focus on, based on needs and
+  real usage data.
+
+## Rationale, alternatives, and prior art
+
+- A variation of the current proposal without polyfills was considered, but it
+  would require users to change web code to accommodate native. For instance,
+  `navigator.getBattery()` would have to be rewritten as
+  `require("@react-native-webapis/battery-status").getBattery()`.
```
cdsanchez, line 169: It may be tangential to this - but it seems that another benefit of the require/import-based pattern is that you could directly reference only the relevant types rather than bringing in the global ambient DOM typings provided by TypeScript (which may bring in symbols that are not actually available in your RN runtime). I'm not sure how this works for Flow.

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,227 @@
...
+  modules. It needs to be installed before it can be used. The current thinking
+  is to install this layer as polyfills. Installing them on-demand is preferred,
+  but we need to start somewhere and we should be able to improve this layer
+  later without affecting users. In any case, we will need tools to include
+  polyfills similarly to [autolinking][].
+
+- **Turbo/native module:** This is the layer that accepts JS calls and forwards
+  them platform implementations. Modules are installed via the standard
+  autolinking mechanism.
+
```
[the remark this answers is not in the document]
tido64: Thanks for pointing this out. I've added it as a note.

text/0002-react-native-webapis.md
tido64: I haven't really thought of this. I was so focussed on being able to consume web code as is. I don't know if we can configure TypeScript to be more selective here, but I've added your comment here so we don't forget.

text/0002-react-native-webapis.md
kelset: another note: to help us prioritize against our internal needs, we've been creating a tool to help us scrape a codebase to find usages of every WebAPI so that we can be data-driven https://github.com/microsoft/rnx-kit/pull/2621

text/0002-react-native-webapis.md
rubennorte: @necolas I think colocation is more critical for APIs that interact with internals from RN (like all the DOM traversal & layout APIs that interact with Fabric), but it might not be as important for standalone APIs like Battery, etc. that could be easily implemented as a userland module. I agree with the general direction of eventually having this provided by the framework out of the box, but I don't think it's that important at this point (and it'll help with iteration speed).

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,292 @@
...
+- Alternatively, native code may not be necessary at all if a complete
+  implementation can be done in pure JS.
+
+### Modularity
+
+We want to avoid introducing unused modules and adding unnecessary bloat to the
+app bundle. WebAPIs must therefore be broken down into smaller modules that can
+be installed separately. These modules are installed by autolinking, and must
+therefore be explicitly added to an app's `package.json`.
```
rubennorte, line 96: I've been thinking about this too but the direction I was considering was a bit different. If we use the dependencies listed in `package.json` and the current autolinking approach, we might end up including JS and native code in the binary that's not actually used. The approach I was considering instead was doing 2 things:
1. Use metadata generated by Metro during bundling to know what native modules are actually used in the code. Use that information for autolinking (so a module that's installed but not actually imported in the bundle would **not** be included).
2. Use a Babel transform to automatically replace the browser API with imports to the native modules. E.g.:
```javascript
// From
navigator.getBattery();

// To
require("@react-native-webapis/battery-status").getBattery();
````

Optionally, we could also make it so the resolved module is configurable by the user, so we can actually swap implementations whenever necessary.

File: text/0002-react-native-webapis.md [hunk 15]
```diff
@@ -0,0 +1,292 @@
...
+
+This is where the need for a monorepo with a quick devloop comes in handy: so
+that partners/interested companies can volunteer to take ownership of some
+specific webapis and work on them in parallel, in a coordinate manner.
+
+### Discovery
+
+The number of modules is high. Finding which modules provide which part of Web
+APIs can be overwhelming for consumers. We need tools that can tell users which
+dependencies they need to add. At minimum we should:
```
rubennorte, line 178: See my previous comment about autolinking and how it could completely solve the discovery problem :)

text/0002-react-native-webapis.md
tido64: Hi @rubennorte! Thanks for your feedback. I had some thoughts around these approaches as well (that I probably should put somewhere in the document),

1. You're right that current native autolinking is too naive. Your proposal makes sense — the only thing I would add is that we probably still need to keep the current approach as fallback for when there is no bundle e.g., because a dev server is used. Somewhere in this document, I mentioned a tool for telling authors that they're using a Web API that has not been polyfilled and either suggest or add a reasonable library. This same tool should be able to produce the metadata you mentioned.
2. Our current approach is to inject the polyfills at the beginning of the bundle. This approach simple to implement, but doesn't allow for tree shaking like yours does. Do you think replacing calls is something that we can perform with 100% accuracy? How do we know `navigator` doesn't refer to something else? I think the same problem applies to `Range` and other APIs? I'm not so well-versed in Babel or AST manipulation in general, so please forgive the dumb questions.

I'm thinking that we want to align with the behaviour of native autolinking i.e., we should allow autopolyfilling to be disabled via the very [same mechanisms](https://github.com/react-native-community/cli/blob/main/docs/autolinking.md#how-can-i-disable-autolinking-for-unsupported-library). If we can tell people that it works the exact same way native autolinking does, that will reduce cognitive load. We should aim to fix native autolinking behaviour as necessary. I'm a bit wary of introducing yet another concept and making it hard to debug/wrap your head around.

rubennorte: > the only thing I would add is that we probably still need to keep the current approach as fallback for when there is no bundle e.g., because a dev server is used.

Yeah, I think there are cases where forcing this is reasonable (including when using OTA), so we definitely need a way to force certain native modules to always be included regardless of whether they're used or not. We probably shouldn't be doing that by default though.

In the future, if we shipped the browser APIs with RN, we could have an option in the RN config like:

```typescript
browserGlobals: {
  select?: 'detected' | 'all' | 'none',
  include?: Array<string>,  // doesn't work with 'all'
  exclude?: Array<string>, // doesn't work with 'none'
}
``` 

> Somewhere in this document, I mentioned a tool for telling authors that they're using a Web API that has not been polyfilled and either suggest or add a reasonable library. This same tool should be able to produce the metadata you mentioned.

I think that tool would be useful but only in the case we discussed before (using OTA or SSR). For the metadata, my approach was about native modules, not about Web APIs specifically (which has the added benefit of reducing bloat in a broader set of cases).

> Our current approach is to inject the polyfills at the beginning of the bundle. This approach simple to implement, but doesn't allow for tree shaking like yours does. Do you think replacing calls is something that we can perform with 100% accuracy? How do we know navigator doesn't refer to something else? I think the same problem applies to Range and other APIs? I'm not so well-versed in Babel or AST manipulation in general, so please forgive the dumb questions.

We can tell if the reference to `navigator` is global or if it's redefined/imported from another module. The only requirement from this would be that you use it directly (as in `navigator.getBattery()`) instead of using indirection (e.g.: `const {getBattery} = navigator; getBattery();`).

> I'm thinking that we want to align with the behaviour of native autolinking i.e., we should allow autopolyfilling to be disabled via the very [same mechanisms](https://github.com/react-native-community/cli/blob/main/docs/autolinking.md#how-can-i-disable-autolinking-for-unsupported-library). If we can tell people that it works the exact same way native autolinking does, that will reduce cognitive load. We should aim to fix native autolinking behaviour as necessary. I'm a bit wary of introducing yet another concept and making it hard to debug/wrap your head around.

I think it's fair to focus on the current capabilities in RN to implement this, but we need to make sure it wouldn't conflict with the direction that we want to take in the long term.

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,292 @@
...
+module.exports = {
+  dependency: {
+    api: {
+      polyfill: "./polyfill.js",
+    },
+  },
+};
+```
+
+Polyfills are gathered and passed to Metro. Any dependency with a correctly
```
rubennorte, line 117: cc @motiz88

text/0002-react-native-webapis.md
```diff
@@ -0,0 +1,292 @@
...
+the whole `react-native-webapis` module, but only the dedicated
+`@react-native-webapis/battery-status` submodule.
+
+Additionally, we want to avoid requiring that users manually add polyfills for
+the modules they need. Instead, we propose that modules that implement a part of
+the API to declare a polyfill in their `react-native.config.js`:
+
+```js
+// react-native.config.js
+module.exports = {
```
rubennorte, line 108: @tido64  could you use a specific example here? I'm not sure I understand the shape of this config 100%.

tido64: At the moment, there really isn't much else to it. We're building on top of the existing autolinking schema as defined in the CLI: https://github.com/react-native-community/cli/blob/main/docs/autolinking.md#what-do-i-need-to-have-in-my-package-to-make-it-work

Just to be clear, this `react-native.config.js` lives in the module package. Not in the app project.

kelset (approved):
thanks everyone for your inputs! this has made the RFC so much better - you can keep up with the actual coding work over in the issues section and the webapis folder of the repo: https://github.com/microsoft/rnx-kit/tree/main/incubator/%40react-native-webapis

PR #2541, opened 2023-07-17
fix(react-native-test-app-msal): implement `initWithHost:`

tido64:
### Description

Implement `initWithHost:` introduced in `react-native-test-app` 2.5.11

### Test plan

```
cd packages/test-app
yarn build --dependencies
open ios/SampleCrossApp.xcworkspace
```

In Xcode, set breakpoints as in the screenshot below, and build/run the app:

image

Once the app is open, dismiss the red box and tap on "MicrosoftAccounts (iOS/macOS)". The breakpoint in `init(host:)` should hit. The other one should not.

### DIFF
--- a/.changeset/shiny-bugs-lick.md
+++ b/.changeset/shiny-bugs-lick.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/react-native-test-app-msal": patch
+---
+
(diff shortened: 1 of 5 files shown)


kelset (approved):
Tested, works as expected by following the steps:

Screenshot 2023-07-18 at 15 20 54

hits the first breakpoint, skips the second, as expected 👍

PR #2542, opened 2023-07-17
feat(react-native-lazy-index): Extend experiences section

AAHAbbas:
### Description
Extending `parseExperiencesFromObject` to support new format to enable/disable flighted experiences. An experience won't be loaded when setting `flighted: false`. 
 
### Test plan
Added unit tests

### DIFF
--- a/.changeset/dirty-crabs-hang.md
+++ b/.changeset/dirty-crabs-hang.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/react-native-lazy-index": patch
+---
+
+Extend `parseExperiencesFromObject` to support enabling/disabling flighted experiences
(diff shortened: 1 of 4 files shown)


tido64 (commented):
Thanks for adding this. Left a few comments that should be addressed before we can merge it.

.changeset/dirty-crabs-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rnx-kit/react-native-lazy-index": patch
```
tido64, line 2: This is a new feature so it should be set to minor:
```suggestion
"@rnx-kit/react-native-lazy-index": minor
```

tido64 (approved):
Just one minor thing and we can get this merged. 😄

PR #2543, opened 2023-07-19
fix(build): bump `@octokit/core` to 5.x

tido64:
### Description

BREAKING CHANGE: `@octokit/core` 5.0.0 drops support for Node 14 and 16.

Testing shows that at least Node 16 still works if you import `node-fetch`, so technically, if your setup does not enforce Node version, you can still use this package.

### Test plan

```sh
cd incubator/build
yarn build --dependencies
yarn rnx-build -p ios --device-type simulator ../../packages/test-app
```

### DIFF
--- a/.changeset/hungry-rabbits-learn.md
+++ b/.changeset/hungry-rabbits-learn.md
@@ -1,0 +1,7 @@
+---
+"@rnx-kit/build": minor
+---
+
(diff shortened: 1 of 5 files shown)


kelset (approved):
it's working correctly in my test 👍

PR #2549, opened 2023-07-21
chore: enable `verbatimModuleSyntax` on ESM modules

tido64:
### Description

Enables additional guard rails for ESM modules.

See https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax

### Test plan

CI should pass.


kelset (approved):
LGTM

Issue #2457, opened 2023-06-09
Have workspace Resolution functionality along with Node Module Resolution for dependency resolution


knarwani:
**What I need?**
Can we please have the workspace resolution functionality as well along with node_module resolution of the dependencies? 

**Why I need?**
I am using rnx-kit for aligning dependencies in my project. I am working on a monorepo and I wanted to use align-deps before installing any dependencies in my project (To run rnx-kit I have created a bundle of it and calling it). For that I need workspace resolution of few of my internal packages which are yet not available in node_modules. So, can we please have the workspace resolution functionality as well along with node_module resolution of the dependencies?

**Where to change?**
I think [this](https://github.com/microsoft/rnx-kit/blob/main/packages/align-deps/src/preset.ts#L24) is the place where my functionality can be added.

krishna8770:
Hi there, 

I wanted to change [this](https://github.com/microsoft/rnx-kit/blob/main/packages/align-deps/src/preset.ts) file where loadPreset function will change to 
```
function loadPreset(
  preset: string,
  projectRoot: string,
  resolve = require.resolve
): Preset {
  switch (preset) {
    case "microsoft/react-native":
      return reactNativePreset;
    default: {
      const info = loadInfo(preset);
      if (info !== undefined) {
        const returnResolvedPath = path.join(path.dirname(info.packageJsonPath), parsedInfo.path);
        return require(returnResolvedPath);
      }
      else {
        return require(resolve(preset, { paths: [projectRoot] }));
      }
    }
  }
}
```

and will add this const in [config.ts](https://github.com/microsoft/rnx-kit/blob/main/packages/align-deps/src/config.ts)

```
export const allPackages = getPackageInfos(process.cwd());
```

I tried this solution in my local and is working fine. Can you have a look at this code and let me know if I can go ahead and create a PR for this change.

Please let me know what your views on this are.

Thanks.

krishna8770:
Yes, the expensiveness of the function was my concern also but, to address that I have exported this variable from [config.ts](https://github.com/microsoft/rnx-kit/blob/main/packages/align-deps/src/config.ts) file i.e., ``` export const allPackages = getPackageInfos(process.cwd()); ``` which then will be calculated only once and then passed on to further iterations. This solution is not at all increasing the runtime on my monorepo with approx 400 packages. Let me know if you are fine with this solution being merged.

tido64:
[...]

PR #2550, opened 2023-07-21
Knarwani/fix workspace resolution issue

krishna8770:
### Description

  - What issue are you trying to solve?
               Please check [this](https://github.com/microsoft/rnx-kit/issues/2457) issue for details

  - How does this change address the issue?
               This change will make sure that if align-deps is being run by a Monorepo then it will consider the workspaces else it will directly check inside node_modules.
  - If applicable, can you attach screenshots of before and after your
    change?

### Test plan
The feature added is not big, so no extra test has been added but all the available test suites are passed.

### DIFF
--- a/packages/align-deps/src/cli.ts
+++ b/packages/align-deps/src/cli.ts
@@ -16,6 +16,8 @@ import { defaultConfig } from "./config";
 import { printError, printInfo } from "./errors";
 import { isString } from "./helpers";
 import type { Args, Command } from "./types";
+import type { PackageInfos } from "workspace-tools";
+import { getPackageInfos } from "workspace-tools";
 
 export const cliOptions = {
   "exclude-packages": {
@@ -142,7 +144,7 @@ function reportConflicts(conflicts: [string, string][], args: Args): boolean {
   }, false);
 }
 
-async function makeCommand(args: Args): Promise<Command | undefined> {
+async function makeCommand(args: Args, allPackages: PackageInfos): Promise<Command | undefined> {
   const conflicts: [string, string][] = [
     ["init", "set-version"],
     ["init", args.write ? "write" : "no-write"],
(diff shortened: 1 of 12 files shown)


packages/align-deps/src/cli.ts
```diff
@@ -175,21 +177,22 @@ async function makeCommand(args: Args): Promise<Command | undefined> {
...
+    return makeSetVersionCommand(setVersion, options, allPackages);
   }
 
-  return makeCheckCommand(options);
+  return makeCheckCommand(options, allPackages);
 }
 
 export async function cli({ packages, ...args }: Args): Promise<void> {
-  const command = await makeCommand(args);
+  const allPackages = getPackageInfos(process.cwd());
```
tido64, line 194: Hi, thanks for working on this. I think it's a good start, but I have several issues with this:

1. We're scanning the whole monorepo and reading all packages before the command has been created. This work may be thrown away if we exit early here.
2. By introducing this call, we're now doing two scans instead of one. See `getManifests()`.
3. `getPackageInfos` eagerly reads all packages to return `PackageInfos`. We may not even need them. This should be done lazily.
4. We don't want to introduce new dependencies, i.e. we cannot add `parse-package-name` and `workspace-tools`. If we're missing anything, we should consider adding them to [`tools-node`](https://github.com/microsoft/rnx-kit/tree/main/packages/tools-node) or [`tools-workspaces`](https://github.com/microsoft/rnx-kit/tree/main/packages/tools-workspaces) instead.

Keep in mind that we have users running align-deps as a Git hook. This command cannot be reading the whole monorepo before the checks even start. We need to postpone work as much as possible until we actually need them.

We should be passing the list of manifests to the command itself and then to `loadPreset`. `loadPreset` will try finding the package from the monorepo if it's there. This result should be cached so that we only perform this once. If we are passing a "context" object with all the information, we can easily add caching to it. This will require us to modify the signature of commands, but we should be able to make it backwards compatible:

```diff
 export function checkPackageManifest(
-  manifestPath: string,
+  context: string | {
+    manifestPath: string;
+    manifests: string[];
+    "$cache": Record<string, string>
+  },
   options: Options,
   inputConfig = loadConfig(manifestPath, options),
   logError = error
 ): ErrorCode {
```

PR #2556, opened 2023-07-26
fix issue caused by Lerna 7+ `useWorkspaces` deprecation

### LINKED ISSUE #2555: Lerna 7+ breaks tools-workspaces functionality
#### Problem
Lerna 7.0.0 introduced a [breaking change](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#700-2023-06-08) that deprecates `useWorkspaces`. Instead Lerna now relies on a `npmClient` property to detect workspaces, unless package list is overriden by a `packages` property.

This in turn breaks the `@rnx-kit/metro-config`, since it is no longer able to correctly populate `watchFolders` property for metro config.

#### Workarounds
1) duplicate package list in `lerna.json` this "theoretically" should allow `tools-workspaces` to proceed as usual.
2) manually populate `watchFolders` metro config property.

#### Solution
My initial impulse was to simply cut everything lerna-related from `tools-workspaces`, as Lerna seemingly no longer considers itself a "sentinel". This presents 2 problems: a) how do we keep compatibility with Lerna pre-7 versions that still rely on `useWorkspaces`? b) how do we keep compatibility with Lernas ability to override package list through `packages` property?

Code in question (packages/tools-workspaces/src/lerna.ts):
```typescript
// https://lerna.js.org/docs/configuration
export async function findWorkspacePackages(
  configFile: string
): Promise<string[]> {
  const { packages, useWorkspaces } = await readJSON(configFile);
  if (!useWorkspaces) {
    return await findPackages(packages, path.dirname(configFile));
  }

  const root = path.dirname(configFile);
  const sentinels = filterSentinels();
  for (const sentinel of sentinels) {
    const filename = path.join(root, sentinel);
    if (fileExists(filename)) {
      const { findWorkspacePackages } = await getImplementation(filename);
      return await findWorkspacePackages(filename);
    }
  }

  return [];
}
```

so my way of thinking is that we:
1) check for `packages` first -> proceed as usual if found
2) do not check for `useWorkspaces` -> try to fallback to other sentinels regardless

like this:
```typescript
// https://lerna.js.org/docs/configuration
export async function findWorkspacePackages(
  configFile: string
): Promise<string[]> {
  const { packages } = await readJSON(configFile);
  // respect the Lerna `packages` property
  if (packages) {
    return await findPackages(packages, path.dirname(configFile));
  }

  // fallback to alternative sentinels regardless of `useWorkspaces`
  const root = path.dirname(configFile);
  const sentinels = filterSentinels();
  for (const sentinel of sentinels) {
    const filename = path.join(root, sentinel);
    if (fileExists(filename)) {
      const { findWorkspacePackages } = await getImplementation(filename);
      return await findWorkspacePackages(filename);
    }
  }

  return [];
}
```

The only uncovered case is when user deliberately sets `useWorkspaces` to false, does not provide lerna with `packages` but still expects things to somehow magically work. Please tell me if you're aware of such a setup.

### Affected Package

@rnx-kit/tools-workspaces

### Version

0.1.3

- [X] Android
- [X] iOS
- [X] macOS
- [X] Windows

### System Information

```true
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz        
    Memory: 18.37 GB / 31.93 GB
  Binaries:
    Node: 18.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.5.1 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK: Not Found
    Windows SDK:
      AllowAllTrustedApps: Enabled
      Versions: 10.0.19041.0
  IDEs:
    Android Studio: AI-222.4459.24.2221.10121639
    Visual Studio: 16.11.33801.447 (Visual Studio Community 2019)
  Languages:
    Java: 11.0.18
  npmPackages:
    @react-native-community/cli: ^10.0.0 => 10.2.5
    react: 18.2.0 => 18.2.0
    react-native: ^0.71.5 => 0.71.11
    react-native-windows: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

### Steps to Reproduce

Here's the minimal reproduction example. I'm not familiar with `yarn` workspaces, so the only limitation here is that it requires `pnpm`.

https://github.com/poniraq/lerna7-tools-workspaces-bug-example

### Code of Conduct

poniraq:
### Description

Lerna 7.0.0 introduced a [breaking change](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#700-2023-06-08) that deprecates useWorkspaces. Instead Lerna now relies on a npmClient property to detect workspaces, unless package list is overriden by a packages property.

This in turn breaks the @rnx-kit/metro-config, since it is no longer able to correctly populate watchFolders property for metro config.

Please refer to the [issue](#2555) for more details
Resolves #2555 

### Test plan

Minimal example: https://github.com/poniraq/lerna7-tools-workspaces-bug-example
Alternatively, any workspace with Lerna 7+ will reproduce the issue unless you manually populate `packages` for Lerna.

My only concern is that this change technically introduces an uncovered edge case:
1) Lerna pre-7
2) workspace setup via pnpm or yarn
3) `useWorkspaces` is deliberately set to `false` in `lerna.json`

### DIFF
--- a/packages/tools-workspaces/src/lerna.ts
+++ b/packages/tools-workspaces/src/lerna.ts
@@ -19,11 +19,13 @@ function filterSentinels(): string[] {
 export async function findWorkspacePackages(
   configFile: string
 ): Promise<string[]> {
-  const { packages, useWorkspaces } = await readJSON(configFile);
-  if (!useWorkspaces) {
+  const { packages } = await readJSON(configFile);
+  // respect the Lerna `packages` property
+  if (packages) {
     return await findPackages(packages, path.dirname(configFile));
   }
 
+  // fallback to alternative sentinels regardless of `useWorkspaces`
   const root = path.dirname(configFile);
   const sentinels = filterSentinels();
   for (const sentinel of sentinels) {
@@ -34,6 +36,7 @@ export async function findWorkspacePackages(
     }
   }
 
+  // return empty list if there's no alternative sentinel to fall back to
   return [];
 }
 
--- a/packages/tools-workspaces/src/lerna.ts
+++ b/packages/tools-workspaces/src/lerna.ts
@@ -19,13 +19,15 @@ function filterSentinels(): string[] {
 export async function findWorkspacePackages(
   configFile: string
 ): Promise<string[]> {
-  const { packages } = await readJSON(configFile);
-  // respect the Lerna `packages` property
-  if (packages) {
+  const { packages, useWorkspaces } = await readJSON(configFile);
+
+  // `useWorkspaces` was deprecated: https://github.com/lerna/lerna/releases/tag/7.0.0
+  // respect Lerna `packages` property
+  if (packages && useWorkspaces !== true) {
     return await findPackages(packages, path.dirname(configFile));
   }
 
-  // fallback to alternative sentinels regardless of `useWorkspaces`
+  // fall back to alternative sentinel regardless of `useWorkspaces`
   const root = path.dirname(configFile);
   const sentinels = filterSentinels();
   for (const sentinel of sentinels) {
@@ -36,7 +38,7 @@ export async function findWorkspacePackages(
     }
   }
 
-  // return empty list if there's no alternative sentinel to fall back to
+  // return empty list if there is no alternative to fall back to
   return [];
 }
 
--- a/.changeset/twelve-sheep-press.md
+++ b/.changeset/twelve-sheep-press.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/tools-workspaces": patch
+---
+
+fix issue caused by Lerna 7+ `useWorkspaces` deprecation


packages/tools-workspaces/src/lerna.ts
```diff
@@ -19,11 +19,13 @@ function filterSentinels(): string[] {
 export async function findWorkspacePackages(
   configFile: string
 ): Promise<string[]> {
-  const { packages, useWorkspaces } = await readJSON(configFile);
-  if (!useWorkspaces) {
+  const { packages } = await readJSON(configFile);
+  // respect the Lerna `packages` property
```
tido64, line 23: Could you add this link here for future reference: https://github.com/lerna/lerna/releases/tag/7.0.0

tido64 (approved):
I'd like to echo @kelset's sentiment that this is really exemplary. Thank you ❤️

packages/tools-workspaces/src/lerna.ts
```diff
@@ -19,11 +19,13 @@ function filterSentinels(): string[] {
...
+  if (packages) {
```
tido64, line 24: Would it make sense to do an explicit check here?

```suggestion
  if (packages && useWorkspaces !== true) {
```

poniraq: I honestly don't know what's the expected pre-7 Lerna behaviour here. What should happen if user expicitly sets `useWorkspaces` to false, but still provides `packages` list. I bet people at Lerna didnt know too so they just decided to deprecate `useWorkspaces`.

Could you please remind what's your preferred style is? Is it amends or do i just slap commits on top?

poniraq:
I've added requested changes as a new commit on top of "changeset" one, for the time being. let me know if you want me to squash/reorder them.

packages/tools-workspaces/src/lerna.ts
tido64: It looked like they had a use case for it a long time ago. I think it makes sense to keep it for now.

Can you please push additional commits? That way we have the history of changes here. We will squash before rebasing and merging with main.

tido64:
I've set this to automerge when CI passes. I will make a new release next week. Thank you ❤️

poniraq:
> I've set this to auto-merge when CI passes. I will make a new release next week. Thank you ❤️

@tido64 I think the merge is blocked by unresolved conversations. Could you please resolve them if you're satisfied with the changes?

tido64:
> @tido64 I think the merge is blocked by unresolved conversations. Could you please resolve them if you're satisfied with the changes?

Ah, the mobile client hides conversations so I didn't see it. In the future, feel free to resolve it yourself when you've addressed it. We always review the changes regardless.

PR #2591, opened 2023-08-08
Refactor saveAssets code to allow out of tree overrides

acoates-ms:
### Description
This change brings in most of https://github.com/react-native-community/cli/pull/2002 into rnx-kit.  Minus the config loading part, since that schema is controlled by the RN cli.

Currently the CLI provides logic to saveAssets for iOS/Android. -- with some combination of the two implementations working for the basic cases for other platforms. But as other platforms mature, they want to be able to provide more platform specific optimizations similar to how iOS provides logic to generate asset catalogs.

https://github.com/microsoft/react-native-windows/pull/11839 adds support for the new saveAssetsPlugin to successfully relocate assets - and this PR will automatically pick up the new plugin in @office-iss/react-native-win32, without going through rn-config.  Once it is added to the config we will be able to remove this special case logic.

### DIFF
--- a/packages/metro-service/src/asset/android.ts
+++ b/packages/metro-service/src/asset/android.ts
@@ -3,6 +3,7 @@
 import * as path from "path";
 import { getResourceIdentifier } from "./assetPathUtils";
 import type { PackagerAsset } from "./types";
+import type {AssetData} from 'metro';
 
 export function getAndroidAssetSuffix(scale: number): string {
   const tolerance = 0.01;
@@ -51,4 +52,20 @@ export function getAssetDestPathAndroid(
   const androidFolder = getAndroidResourceFolderName(asset, scale);
   const fileName = getResourceIdentifier(asset);
   return path.join(androidFolder, `${fileName}.${asset.type}`);
+}
+
+export function saveAssetsAndroid(
+  assets: ReadonlyArray<AssetData>,
+  _platform: string,
+  _assetsDest: string | undefined,
+  _assetCatalogDest: string | undefined,
+  addAssetToCopy: (
+    asset: AssetData,
+    allowedScales: number[] | undefined,
+    getAssetDestPath: (asset: AssetData, scale: number) => string,
+  ) => void,
+) {
+  assets.forEach((asset) =>
+    addAssetToCopy(asset, undefined, getAssetDestPathAndroid),
+  );
 }
--- a/packages/metro-service/src/asset/default.ts
+++ b/packages/metro-service/src/asset/default.ts
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import path from "path";
+import type {AssetData} from 'metro';
+import type { PackagerAsset } from "./types";
+
+export function getAssetDestPath(
+  asset: PackagerAsset,
+  scale: number
+): string {
+  const suffix = scale === 1 ? "" : `@${scale}x`;
+  const fileName = `${asset.name + suffix}.${asset.type}`;
+  return path.join(
+    // Assets can have relative paths outside of the project root.
+    // Replace `../` with `_` to make sure they don't end up outside of
+    // the expected assets directory.
+    asset.httpServerLocation.substr(1).replace(/\.\.\//g, "_"),
+    fileName
+  );
+}
+
+export function saveAssetsDefault(
+  assets: ReadonlyArray<AssetData>,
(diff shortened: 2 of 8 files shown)


packages/metro-service/src/bundle.ts
```diff
@@ -41,6 +45,42 @@ type RequestOptions = {
...
+      return require(saveAssetsPlugin);
+    }
+  } catch (_) {
+    /* empty */
+  }
+  return platform === "ios"
+    ? saveAssetsIOS
+    : platform === "android"
+    ? saveAssetsAndroid
+    : saveAssetsDefault;
```
tido64, line 81: Nit: Instead of nested ternary ops, can we use a `switch`? That way we can handle `win32`  as well.

packages/metro-service/src/asset/write.ts
```diff
@@ -58,11 +50,22 @@ function copyAll(filesToCopy: Record<string, string>) {
   });
 }
 
-export function saveAssets(
+export async function saveAssets(
```
tido64, line 53: Is this necessary? It doesn't look like we `await` in here?

packages/metro-service/src/asset/write.ts
```diff
@@ -58,11 +50,22 @@ function copyAll(filesToCopy: Record<string, string>) {
...
+    assets: ReadonlyArray<AssetData>,
+    platform: string,
+    assetsDest: string | undefined,
+    assetCatalogDest: string | undefined,
+    addAssetToCopy: (
+      asset: AssetData,
+      allowedScales: number[] | undefined,
+      getAssetDestPath: (asset: AssetData, scale: number) => string
+    ) => void
+  ) => void
```
tido64, line 68: Use the type?
```suggestion
  saveAssetsPlugin: SaveAssetsPlugin
```

packages/metro-service/src/asset/default.ts
```diff
@@ -0,0 +1,37 @@
...
+  assets: ReadonlyArray<AssetData>,
+  _platform: string,
+  _assetsDest: string | undefined,
+  _assetCatalogDest: string | undefined,
+  addAssetToCopy: (
+    asset: AssetData,
+    allowedScales: number[] | undefined,
+    getAssetDestPath: (asset: AssetData, scale: number) => string
+  ) => void
+) {
```
tido64, line 35: Nit: If we make this an arrow function, we can type it properly:
```suggestion
export const saveAssetsDefault: SaveAssetsPlugin = (
  assets,
  _platform,
  _assetsDest,
  addAssetToCopy
) => {
```

packages/metro-service/src/asset/ios.ts
```diff
@@ -68,17 +70,48 @@ export function writeImageSet(imageSet: ImageSet): void {
...
+  assets: ReadonlyArray<AssetData>,
+  _platform: string,
+  _assetsDest: string | undefined,
+  assetCatalogDest: string | undefined,
+  addAssetToCopy: (
+    asset: AssetData,
+    allowedScales: number[] | undefined,
+    getAssetDestPath: (asset: AssetData, scale: number) => string
+  ) => void
+) {
```
tido64, line 85: Nit: If we make this an arrow function, we can type it properly:
```suggestion
export const saveAssetsIOS: SaveAssetsPlugin = (
  assets,
  _platform,
  _assetsDest,
  addAssetToCopy
) => {
```

packages/metro-service/src/asset/android.ts
```diff
@@ -52,3 +53,19 @@ export function getAssetDestPathAndroid(
...
+  assets: ReadonlyArray<AssetData>,
+  _platform: string,
+  _assetsDest: string | undefined,
+  _assetCatalogDest: string | undefined,
+  addAssetToCopy: (
+    asset: AssetData,
+    allowedScales: number[] | undefined,
+    getAssetDestPath: (asset: AssetData, scale: number) => string
+  ) => void
+) {
```
tido64, line 67: Nit: If we make this an arrow function, we can type it properly:
```suggestion
export const saveAssetsAndroid: SaveAssetsPlugin = (
  assets,
  _platform,
  _assetsDest,
  addAssetToCopy
) => {
```

tido64 (approved):

packages/metro-service/src/asset/default.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
```
tido64, line 8: Can we remove this?
```suggestion
```

tido64 (approved):

PR #2621, opened 2023-08-22
chore(webapis): add tool for scanning for uses of Web APIs

tido64:
### Description

Implemented a tool for scanning for uses of Web APIs.

### Test plan

1. [Install Rust](https://www.rust-lang.org/learn/get-started)
2. Build and run:
    ```sh
    cd incubator/@react-native-webapis/scan
    cargo build -r
    cd ../../..
    ./incubator/@react-native-webapis/scan/target/release/scan
    ```


incubator/@react-native-webapis/scan/src/ignored_dirs.rs
```diff
@@ -0,0 +1,115 @@
...
+    "@babel",
+    "@changesets",
+    "@eslint-community",
+    "@haul-bundler",
+    "@jest",
+    "@pnpm",
+    "@react-native",
+    "@react-native-community",
+    "@react-native-webapis",
+    "@react-native-windows",
```
kelset, line 16:
```suggestion
    "@react-native-community",
    "@react-native-mac",
    "@react-native-webapis",
    "@react-native-windows",
```

if I'm understanding correctly, we need to add the new "@react-native-mac" entry here since it's what the RNmacos team is using for their monorepo (post 72 monorepo landing).

The tool was indeed annoyed at some files in it:
Screenshot 2023-09-19 at 15 20 30

(I'm not sure what we want/should do about the `Failed to parse source file: ./packages/typescript-service/test/__fixtures__/sample-snapshot-file.js` error)

incubator/@react-native-webapis/scan/merge.mjs
```diff
@@ -0,0 +1,669 @@
+// Last updated 2023-08-30
+const data = {
+  fabric: {
```
kelset, line 3:
```suggestion
  big-repo-1: {
```

incubator/@react-native-webapis/scan/merge.mjs
```diff
@@ -0,0 +1,669 @@
...
+    ShadowRoot: 1,
+    "navigator.replace": 1,
+    "navigator.push": 1,
+    "navigator.createHref": 1,
+    "navigator.msSaveOrOpenBlob": 1,
+    IDBRequest: 1,
+    "navigator.getUserMedia": 1,
+    Plugin: 1,
+  },
+  midgard: {
```
kelset, line 125:
```suggestion
  big-repo-2: {
```

incubator/@react-native-webapis/scan/merge.mjs
```diff
@@ -0,0 +1,669 @@
...
+    HTMLDListElement: 1,
+    TouchEvent: 1,
+    "navigator.cpuClass": 1,
+    HTMLAreaElement: 1,
+    HTMLMapElement: 1,
+    HTMLTableRowElement: 1,
+    HTMLOptGroupElement: 1,
+    HTMLScriptElement: 1,
+  },
+  odsp: {
```
kelset, line 348:
```suggestion
  big-repo-3: {
```

incubator/@react-native-webapis/scan/merge.mjs
```diff
@@ -0,0 +1,669 @@
...
+    IntersectionObserverEntry: 1,
+    FocusEvent: 1,
+    "navigator.permissions": 1,
+    MediaSource: 1,
+    "navigator.updateNavigatorSeries": 1,
+    "navigator.mouseUpHandler": 1,
+    "navigator.mouseMoveHandler": 1,
+    "navigator.navigatorOptions": 1,
+  },
+  ooui: {
```
kelset, line 477:
```suggestion
  big-repo-4: {
```

incubator/@react-native-webapis/scan/src/web_apis.rs
```diff
@@ -0,0 +1,615 @@
+/**
+ * THIS ARRAY MUST BE SORTED! We perform binary searches on it.
+ *
+ * Names extracted from `lib.dom.d.ts` in TypeScript 5.1.6, subtracting names
+ * in `react-native/types/modules/globals.d.ts`.
+ */
+pub static WEB_APIS: &'static [&str] = &[
+    //"AbortController", // Polyfilled by React Native
```
kelset, line 8: could we maybe have the WebAPIs that are polyfilled by RN core as a separate file/data structure and when the tool spins up, it "diffs" the whole list minus the RN one and gets at runtime the one to use? I think it will make it easier for overall maintenance

kelset (changes requested):
thanks for working on this!

I love that by going down the Rust route, the tool is blazing fast ⚡️

I left a few comment; aside from those, can we add a README.md to the root of scan that explains:
1. how to use it/build it
2. what the merge.mjs file is
3. what is the counting logic (if I read correctly, we count all instances across all files we find minus the ones in the ignored_dirs?)

incubator/@react-native-webapis/scan/src/ignored_dirs.rs
tido64: These are usually due to the source file being Flow typed. Ideally, we should strip out the types and try again. But it doesn't skew the numbers that much, so I will address it later.

incubator/@react-native-webapis/scan/src/web_apis.rs
tido64: Would you mind if we addressed this another time? Right now, this is a compile-time list. Making the changes you suggest means that we have to make a copy of this with the polyfilled modules filtered out. It's doable, but it means additional code.

kelset: understood

kelset (approved):
thanks for addressing the feedback!

PR #2623, opened 2023-08-24
fix(esbuild-bundle-analyzer): cleaner output 

AAHAbbas:
### Description
Fixing few of the issues and TODOs

- ~Searching for which line an import is defined in a file is more generic and searches under `lib/`, `src/`, `dist/`, and `build/`, and support more file extensions~
-  ~`userRequest` outputs cleaner and more accurate import statements~
- Added support to remove any arbitrary namespace prefix and output a relative module name
- Added support for multiple chunks - all chunks are disconnected
- Cleaned the code a bit and fixed some issues and logical errors
- Switched to using CJS module

### Test plan
- Generate metafile
- Run `yarn rnx-analyze meta.json --output result.json`
- Upload `result.json` [here](https://happy-water-0887b0b1e.azurestaticapps.net/#/)


incubator/esbuild-bundle-analyzer/package.json
```diff
@@ -11,13 +11,8 @@
   "files": [
     "lib/*"
   ],
-  "type": "module",
```
tido64, line 14: What was the reason for switching back to CJS? Can you include it in the change logs?

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -40,39 +45,70 @@ function getLine(filePath: string, keyword: string): string {
...
+  return userRequest;
+}
+
+function removePrefix(filePath: string): string {
+  let parts = filePath.split(":");
+
+  while (parts.length > 1) {
+    filePath = parts.slice(1).join(":");
+    parts = filePath.split(":");
+  }
```
tido64, line 71: This looks really odd to me. Why wouldn't you just take the last bit instead of splitting and joining in a loop? Also, does this handle Windows paths e.g., `D:\\Source\\`?

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -40,39 +45,70 @@ function getLine(filePath: string, keyword: string): string {
...
+
+  const patternToRemoveRegex =
+    /\/(src|lib|dist|build)\/index(\.[^.]+)?(\.[^.]+)?$/;
+
+  const match = userRequest.match(patternToRemoveRegex);
+  if (match) {
+    return userRequest.replace(match[0], "");
+  }
+
+  return userRequest;
```
tido64, line 62: I think you can simplify this:
```suggestion
  return userRequest.replace(patternToRemoveRegex, "");
```

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -1,36 +1,41 @@
...
+      // import { ensureDir } from "@rnx-kit/cli/bundle/metro";
+      if (line.includes("/lib/")) {
+        return line.replace("/lib/", "/src/").includes(keyword);
+      } else if (line.includes("/dist/")) {
+        return line.replace("/dist/", "/src/").includes(keyword);
+      } else if (line.includes("/build/")) {
+        return line.replace("/build/", "/src/").includes(keyword);
+      } else {
+        return false;
       }
```
tido64, line 38: I think this is making a lot of assumptions of how the project is structured. We shouldn't be changing the path.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -1,36 +1,41 @@
...
   WebpackStats,
-} from "./types.js";
+} from "./types";
 
-// TODO: Update to be more generic and configurable to it supports
-// more than just lib and src folders
 function getLine(filePath: string, keyword: string): string {
   try {
-    const file = fs.readFileSync(filePath, "utf-8");
+    const file = fs.readFileSync(path.resolve(filePath), "utf-8");
```
tido64, line 15: Why is `path.resolve` necessary here? It doesn't change the file path.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -40,39 +45,70 @@ function getLine(filePath: string, keyword: string): string {
...
-function getSimplePath(dir: string, file: string): string {
+function getSimplePath(file: string): string {
+  if (!file.startsWith("..")) {
+    file = removePrefix(file);
+  }
+
   const index = file.indexOf("node_modules");
-  return index >= 0 ? file.slice(index) : path.relative(dir, file);
+  return index >= 0 ? file.slice(index) : path.relative(process.cwd(), file);
 }
```
tido64, line 89: I don't think we should be modifying the path at all. That's for the presentation layer to decide whether it makes sense to do so. If there are multiple copies of a package, we cannot know which `node_modules` folder we're looking at or where it came from.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
AAHAbbas: Changed it to only map back if the user provide the source and destination folder.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -40,39 +36,73 @@ function getLine(filePath: string, keyword: string): string {
...
+  }
+
+  return result;
+}
+
+/**
+ * Returns a simpler and more readable path which starts from node_modules
+ * and removes the namespace prefix, e.g.:
+ * namespace:user/x/.store/@test-library@0.0.1-b55dbe3d1aed7a6c074d/node_modules/@test-library/a/b/test.js
  * becomes node_modules/@test-library/a/b/test.js
```
tido64, line 72: I am not convinced this is the correct thing to do. It's fundamentally changing the path. The previous path tells users exactly which copy of the package we're looking at. The transformed path is ambiguous.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -40,39 +36,73 @@ function getLine(filePath: string, keyword: string): string {
...
+ * Returns a cleaner userRequest by removing the src/index.ts(x) and lib/index.js(x)
+ * from the import path.
+ */
+export function getCleanUserRequest(userRequest?: string): string {
+  if (!userRequest) return "";
+
+  const patternToRemoveRegex =
+    /\/(src|lib|dist|build)\/index(\.[^.]+)?(\.[^.]+)?$/;
+
+  return userRequest.replace(patternToRemoveRegex, "");
```
tido64, line 48: Should this be looking at the `main` field instead of guessing what the main entry file is?

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -40,39 +36,73 @@ function getLine(filePath: string, keyword: string): string {
...
+  const isWindowsPath = /\\[\\\S|*\S]?.*$/.test(filePath);
+  const last = namespace.split(":").pop();
+
+  if (isWindowsPath && last?.length === 1) {
+    const driveLetter = /[a-zA-Z]$/.test(last) ? `${last}:` : "";
+    return `${driveLetter}${result}`;
+  }
+
+  return result;
+}
```
tido64, line 66: If we're already doing regex, maybe try `/((\b[A-Za-z]:)?[^:]*)$/`?

image

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
AAHAbbas: How about just removing the namespace and making the path relative, would that be better? There needs to be some distinction between the `reasons.module` and `reasons.moduleName`, `module.identifier` and `module.name`.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
AAHAbbas: Great, thanks! This is better.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
AAHAbbas: Yes, I can changed it to use the `src` and `dest` parameters, and if arguments are not passed then use the `main` field. Is that better?

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
tido64: Just making it relative sounds fine to me.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
tido64: On second thought, I don't think we should be doing this at all. It's often useful to see whether we correctly imported a package via ESM vs CJS.

AAHAbbas: Sounds good, I'll just remove this function then.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -85,45 +88,39 @@ export function webpackStats(
...
-              module: cleanInput,
-              moduleName: getSimplePath(metafileDir, cleanInput),
+              module: input,
+              moduleName: getSimplePath(input),
               userRequest: imp.original,
               loc:
                 !skipLineNumber && imp.original
-                  ? getLine(cleanInput, imp.original)
+                  ? getLine(removeNamespace(input), imp.original, src, dest)
                   : "",
```
tido64, line 124: Without parsing the source map, blindly remapping src to build/dist/lib/etc **will** result in the wrong line being used.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -39,40 +35,47 @@ function getLine(filePath: string, keyword: string): string {
   }
 }
 
+export function removeNamespace(filePath: string): string {
+  const match = filePath.match(/((\b[A-Za-z]:)?[^:]*)$/);
+  if (!match) return filePath;
+
+  return match[0];
```
tido64, line 42: Nit:
```suggestion
  return match?.[0] ?? filePath;
```

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
```diff
@@ -39,40 +35,47 @@ function getLine(filePath: string, keyword: string): string {
...
  */
-function getSimplePath(dir: string, file: string): string {
-  const index = file.indexOf("node_modules");
-  return index >= 0 ? file.slice(index) : path.relative(dir, file);
+function getSimplePath(file: string): string {
+  if (!file.startsWith("..")) {
+    file = removeNamespace(file);
+  }
+
+  return path.relative(process.cwd(), file);
```
tido64, line 53: Is `process.cwd()` always correct? What happens if this tool is used in a different folder?

AAHAbbas: We've only been using it from the same folder where we run rnx-bundle to generate the metafile. However, I can extend this to let the user pass a parameter defining the root folder where the bundle and metafile has been generated.

tido64: Can we assume `metafile` resides in the correct project directory?

AAHAbbas: The `metafile` is written to the [location of Metro's sourceMapUrl](https://github.com/microsoft/rnx-kit/blob/950fcf0a3d673df17238bf1a1e15082a45b2bfaa/packages/metro-serializer-esbuild/src/index.ts#L426C1-L430C1). Do you know if `sourceMapUrl` can be configured to point to a different location? For our projects `sourceMapUrl` = `index.mobile.jsbundle.map`. So, the `metafile` resides in the correct project directory.

tido64: You can, but I think they should always reside within the project. You can use [pkg-dir](https://github.com/sindresorhus/pkg-dir) (use 5.x as 6.x is ESM only) to find the project root.

incubator/esbuild-bundle-analyzer/src/webpackStats.ts
AAHAbbas: Ohh right, I'll remove the mapping for now then. I can fix using the source map sometime later 🙂

tido64 (approved):

tido64 (approved):

PR #2636, opened 2023-08-28
Fix: RNXAccountType was wrong on iOS

sveinar:
Proper RNX account type for iOS

### Description

The problem that I am trying to solve here is a crash with MSA accounts on iOS.

In this screenshot, we can perhaps see this little bug show itself with the Typescript string "MicrosoftAccount" does not entirely match up with the iOS string compare on the native side coming in as "MicrosoftAccount" but we only accept the string "Microsoft" (without the suffix Account).

### Test plan

Running locally with an MSA account on iOS.

This does not seem to be an issue on Android.

### DIFF
--- a/.changeset/tender-feet-knock.md
+++ b/.changeset/tender-feet-knock.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/react-native-auth": patch
+---
+
+fix: use correct ios account type string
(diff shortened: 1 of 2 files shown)


tido64:
image

Thanks, @sveinar! Can you run `yarn change` and add a change log entry?

tido64 (approved):

tido64:
Published with [@rnx-kit/react-native-auth@0.2.4](https://github.com/microsoft/rnx-kit/releases/tag/%40rnx-kit/react-native-auth%400.2.4)

PR #2656, opened 2023-09-11
chore: allow `rnx-align-deps.mjs` to skip the build step

tido64:
### Description

- Added [exports map](https://nodejs.org/api/packages.html#package-entry-points) to packages depended on by align-deps
- Reconfigured align-deps' `tsconfig.json` to emit declarations only
- Reconfigured `rnx-align-deps` to skip build

### Test plan

```
yarn clean
yarn
yarn rnx-align-deps
```

`yarn rnx-align-deps` should be fairly instant since it no longer builds all dependencies.

#### Before:

image

#### After:

image

### DIFF
--- a/.changeset/proud-bulldogs-hide.md
+++ b/.changeset/proud-bulldogs-hide.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/babel-plugin-import-path-remapper": patch
+---
+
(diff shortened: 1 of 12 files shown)


kelset:
looks like some tests are failing, can you check them?

PR #2669, opened 2023-09-14
chore(align-deps): update `update-profile.mjs` script

tido64:
### Description

Update `update-profile` script to handle the recent shuffling of dependencies.

### Test plan

```
cd packages/align-deps

# Ensure we can resolve Metro version for 0.73
yarn update-profile 0.73

# Ensure we can resolve Metro version for 0.72 (and older)
yarn update-profile --force 0.72
```


packages/align-deps/scripts/update-profile.mjs
```diff
@@ -22,11 +20,19 @@ import { isMetaPackage } from "../lib/capabilities.js";
...
  * }} PackageInfo
  */
 
+/**
+ * This wrapper is mostly for coercing TypeScript into inferring the correct
+ * type.
+ * @param {any} v
+ * @returns {Record<string, string> | undefined}
+ */
+const Optional = (v) => v;
```
kelset, line 34: image

kelset (approved):
(just as a side note, let's remember that this command "is mistaken" with versions older than 72 (71, etc) and will try to add `"@react-native/metro-config"` even if it doesn't exists. It's not a big deal for us so we'll be fine)

tido64:
> (just as a side note, let's remember that this command "is mistaken" with versions older than 72 (71, etc) and will try to add `"@react-native/metro-config"` even if it doesn't exists. It's not a big deal for us so we'll be fine)

I'll make a mental note of removing the "backwards compatibility" code whenever 0.73 becomes a thing. Things are a bit weird right now since I'm going between 0.72 and 0.73.

PR #2670, opened 2023-09-14
fix(metro-service): fix reporter not being correctly overridden

tido64:
### Description

Fix `yarn start` not showing help on startup.

### Test plan

```
cd packages/test-app
yarn build --dependencies
yarn start
```

A help message should appear.


tido64:
Actually, in order to match what upstream does, we need to do it in a different way. Canceling this.

PR #2693, opened 2023-09-21
fix(eslint-plugin): cache access to `NODE_ENV`

tido64:
### Description

Cache access to `NODE_ENV`. `process.env` is apparently a special object and is slow to access. We don't expect `NODE_ENV` to change here, so we should cache it.

### Test plan

n/a

### DIFF
--- a/.changeset/tasty-stingrays-add.md
+++ b/.changeset/tasty-stingrays-add.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/eslint-plugin": patch
+---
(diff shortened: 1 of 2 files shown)


kelset (approved):
I don't even know how you managed to figure this one out but :shipit:

tido64:
> I don't even know how you managed to figure this one out but :shipit:

😅 Someone else did the work: https://github.com/jestjs/jest/pull/14455

PR #2698, opened 2023-09-22
chore(test-app): reconfigure so our `babel.config.js` gets picked

tido64:
### Description

Metro looks for a Babel config in `projectRoot`, which is currently set to `./src`. Created a root `index.js` that imports `./src/index` and removed `projectRoot` from server config.

### Test plan

```
cd packages/test-app
yarn build --dependencies
yarn start

# In a separate terminal
curl -v 'http://localhost:8081/index.bundle?platform=ios'
```


kelset (approved):
it works

PR #2722, opened 2023-10-09
Pass minify CLI option to metro-serializer-esbuild

AAHAbbas:
### Description
Running `yarn react-native rnx-bundle --dev false --minify false` doesn't generate a prod bundle that's not minified. Passing the `minify` CLI option to the `metro-serializer-esbuild `so the outputted bundle will be minified or not depending on the passed minify flag. 

### Test plan

Prepare:
```
cd packages/test-app
yarn build --dependencies
```

Generating a prod bundle that's minified:
```
yarn bundle+esbuild --dev false
```

Generating a prod bundle that's **not** minified (this would still generate a minified bundle before):
```
yarn bundle+esbuild --dev false --minify false
```

### DIFF
--- a/packages/cli/src/bundle/metro.ts
+++ b/packages/cli/src/bundle/metro.ts
@@ -34,10 +34,11 @@ export async function metroBundle(
 
   if (!dev && bundleConfig.treeShake) {
     if (minify != null) {
-      bundleConfig.treeShake =
-        typeof bundleConfig.treeShake === "object"
-          ? { ...bundleConfig.treeShake, minify }
-          : { minify };
+      if (typeof bundleConfig.treeShake === "object") {
+        bundleConfig.treeShake.minify = minify;
+      } else {
+        bundleConfig.treeShake = { minify };
+      }
     }
   } else {
     bundleConfig.treeShake = false;
--- a/packages/cli/src/bundle/metro.ts
+++ b/packages/cli/src/bundle/metro.ts
@@ -35,6 +35,7 @@ export async function metroBundle(
   customizeMetroConfig(metroConfig, {
     ...bundleConfig,
     treeShake: !dev && bundleConfig.treeShake,
+    minify,
   });
 
   const metroBundleArgs: MetroBundleArgs = {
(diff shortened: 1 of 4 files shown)


packages/cli/src/bundle/metro.ts
```diff
@@ -35,6 +35,7 @@ export async function metroBundle(
   customizeMetroConfig(metroConfig, {
     ...bundleConfig,
     treeShake: !dev && bundleConfig.treeShake,
+    minify,
```
tido64, line 38: Can we create the `EsbuildOptions` object and pass it here instead of passing on `minify`?  `BundleParameters.treeShake` already has `minify` options, so adding another one at root is confusing.

tido64 (changes requested):
Thanks for adding this. I think we want to avoid having two `minify` options inside `BundleParameters`. See the comment below.

packages/cli/src/bundle/metro.ts
```diff
@@ -32,10 +32,18 @@ export async function metroBundle(
 ): Promise<void> {
   info(`Bundling ${bundleConfig.platform}...`);
 
-  customizeMetroConfig(metroConfig, {
-    ...bundleConfig,
-    treeShake: !dev && bundleConfig.treeShake,
-  });
+  if (!dev && bundleConfig.treeShake) {
+    if (minify !== undefined) {
```
tido64, line 36: Just to be on the safe side, let's do coerced comparison to catch both `null` and `undefined`:
```suggestion
    if (minify != null) {
```

packages/cli/src/bundle/metro.ts
```diff
@@ -32,10 +32,18 @@ export async function metroBundle(
...
+      bundleConfig.treeShake =
+        typeof bundleConfig.treeShake === "object"
+          ? { ...bundleConfig.treeShake, minify }
+          : { minify };
```
tido64, line 40: Let's avoid the copy if we can:
```suggestion
      if (typeof bundleConfig.treeShake === "object") {
        bundleConfig.treeShake.minify = minify;
      } else {
        bundleConfig.treeShake = { minify };
      }
```

tido64 (approved):
This looks good to me now. I left a couple of nits.

tido64:
Thanks again 👍

tido64:
FYI: https://github.com/microsoft/rnx-kit/releases/tag/%40rnx-kit/cli%400.16.17

PR #2730, opened 2023-10-12
fix(metro-service): ensure correct metro packages are used

tido64:
### Description

Ensure correct Metro dependencies are used by traversing the dependency chain starting from `react-native`.

### Test plan

```
cd packages/test-app
yarn build --dependencies
yarn android  # or your platform choice

# In a separate terminal
yarn start
```


kelset (approved):
Looks good. Tested locally and it works.

Just for future reference for my silly brain: basically this PR moves the "source of truth" for everything Metro into `@rnx-kit/tools-react-native/metro` so that we can more easily manage it all from there. (aside form types)

PR #2769, opened 2023-10-25
chore(metro-service): update `assets-registry` only when needed

tido64:
### Description

Update `assets-registry` only when needed.

### Test plan

```
cd packages/metro-service
yarn prepare
# No output expected

echo "" > src/assets-registry/path-support.js
echo "" > src/assets-registry/registry.js
yarn prepare
# Two files should be updated
```


kelset (approved):
LGTM

PR #2772, opened 2023-10-27
chore: restrict use of `export default`

tido64:
### Description

Exports should be explicitly named where possible.

### Test plan

n/a


kelset (approved):
giphy

PR #2775, opened 2023-10-31
chore: bump Nx to v17

tido64:
### Description

Bump Nx to [v17](https://blog.nrwl.io/nx-17-0-has-landed-1dad19176ff3)

### Test plan

n/a


package.json
```diff
@@ -56,7 +56,7 @@
     "typescript": "^5.0.0"
   },
   "engines": {
-    "node": ">=14.18"
+    "node": ">=16.13"
```
kelset, line 59: do we have some internal customer still on 16? would be nice to keep it at LTS level but I guess we have a bunch of people still at 18

kelset (approved):
overall look good

package.json
kelset: right, so can't we bump it to at least 18? is there some CI VM or something that still runs on 16 so we need to keep it around?

tido64: Yup, we run Node versions 16, 18, 20 on CI:

https://github.com/microsoft/rnx-kit/blob/7aa6708cdf3bb1a71f4fd7ea602ee57731883f63/.github/workflows/pr.yml#L50-L54

kelset: got it - go ahead with merging 👍

PR #2776, opened 2023-10-31
chore: fix `lint` command not linting all files

tido64:
### Description

`lint` currently only looks at the files that live in `src/` and ignores sub-directories.

### Test plan

CI should pass.


kelset (approved):
nice catch, thanks for fixing!

PR #2779, opened 2023-11-02
fix(metro-config): fix `@react-native/metro-config` not being resolved

tido64:
### Description

Fix `@react-native/metro-config` not being resolved correctly in a pnpm setup.

### Test plan

n/a


kelset (approved):
makes sense 👍

Issue #2783, opened 2023-11-02
Enable pnpm mode to catch issues earlier


tido64:
Despite our best efforts, we cannot catch all pnpm related issues before they hit users. We need to be more proactive and catch them earlier. This means that we _need_ to enable pnpm mode.

To enable pnpm mode:

```diff
diff --git a/.yarnrc.yml b/.yarnrc.yml
index f59a253e..b2cfdf1f 100644
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -7,7 +7,7 @@ logFilters:
     level: discard
   - code: YN0013 # X can't be found in the cache and will be fetched from the remote registry
     level: discard
-nodeLinker: node-modules
+nodeLinker: pnpm
 npmRegistryServer: "https://registry.npmjs.org"
 packageExtensions:
   "@fluentui/utilities@*":
```

### Work items

- [x] https://github.com/microsoft/rnx-kit/pull/2778
- [x] https://github.com/microsoft/rnx-kit/pull/2779
- [x] https://github.com/microsoft/rnx-kit/pull/2780
- [x] https://github.com/microsoft/rnx-kit/pull/2781
- [x] https://github.com/microsoft/rnx-kit/pull/2782
- [x] https://github.com/microsoft/rnx-kit/pull/2784
- [x] https://github.com/microsoft/rnx-kit/pull/2787
- [x] https://github.com/microsoft/rnx-kit/pull/2791
- [x] https://github.com/microsoft/rnx-kit/pull/2792
- [x] https://github.com/microsoft/rnx-kit/pull/2793
- [x] https://github.com/microsoft/rnx-kit/pull/2794
- [x] https://github.com/microsoft/rnx-kit/pull/2795
- [x] https://github.com/microsoft/rnx-kit/pull/2797
- [x] https://github.com/microsoft/rnx-kit/pull/2798
- [x] https://github.com/microsoft/rnx-kit/pull/2799
- [x] https://github.com/microsoft/rnx-kit/pull/2800
- [x] https://github.com/microsoft/rnx-kit/pull/2806
- [x] https://github.com/microsoft/rnx-kit/pull/2808
- [x] https://github.com/microsoft/rnx-kit/pull/2809

### Upstream fixes

- [x] https://github.com/facebook/react-native/pull/41398
  - [ ] 0.73: https://github.com/facebook/react-native/pull/41444
  - [ ] 0.72: https://github.com/facebook/react-native/pull/41445
[...]

PR #2785, opened 2023-11-03
enable pnpm mode

tarunrajput:
### Description

 
Enable pnpm mode to catch issues earlier
Resolves: https://github.com/microsoft/rnx-kit/issues/2783


tido64:
Hi @tarunrajput, thanks for trying to help out. We are not quite ready to enable this yet. As you can see in #2783, there are a number of things that needs to be fixed before we can do that. I will be closing this for now.

PR #2796, opened 2023-11-06
Create fork of chromium-edge-launcher with fix for windows

acoates-ms:
### Description

Recently metro took a dependency on chromium-edge-launcher to launch edge and attach it to a RN instance.

Unfortunately, there is a bug in chromium-edge-launcher which prevents it from correctly finding Edge on Windows machines.  There has not been response from the maintainer, so we're forking it here to allow metro to bring in the fix.

Includes some typings and formatting fixes to allow compilation within the rnx-kit repo.

### DIFF
--- a/incubator/chromium-edge-launcher/package.json
+++ b/incubator/chromium-edge-launcher/package.json
@@ -1,0 +1,56 @@
+{
+  "private": true,
+  "name": "@rnx-kit/chromium-edge-launcher",
+  "version": "0.0.1",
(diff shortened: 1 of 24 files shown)


incubator/chromium-edge-launcher/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "private": true,
```
tido64, line 2: Don't we need to publish this package?
```suggestion
```

tido64 (approved):

PR #2831, opened 2023-11-16
ci: use Node 16 for install/build as well

tido64:
### Description

Otherwise Yarn will detect inconsistencies when we try to publish. This will also reorder things so that all steps related to the website use latest Node LTS.

See also https://github.com/microsoft/rnx-kit/pull/2829, https://github.com/microsoft/rnx-kit/pull/2830

### Test plan

n/a


tido64:
Bypassing policies.

PR #2877, opened 2023-12-06
docs: words about general maintenance

tido64:
### Description

Like https://github.com/microsoft/react-native-test-app/pull/1721, but for rnx-kit.

### Test plan

n/a


kelset (approved):
🚢

PR #2886, opened 2023-12-18
feat(tsconfig): base TypeScript configs for working with Node

tido64:
### Description

Base TypeScript configurations for working with Node. This makes them reusable in external projects, like [`react-native-test-app`](https://github.com/microsoft/react-native-test-app/blob/trunk/tsconfig.json).

### Test plan

CI should pass.


kelset (approved):
:shipit:

PR #2916, opened 2024-01-11
fix(react-native-host): handle refactorings in 0.74

tido64:
### Description

Handle refactorings in 0.74

### Test plan

In `react-native-test-app`:

```
npm run set-react-version 0.74
yarn
cd example
yarn ios

# New Architecture
RCT_NEW_ARCH_ENABLED=1 pod install --project-directory=ios
yarn ios
```

| Old Arch | New Arch |
| :-: | :-: |
| Simulator Screenshot - iPhone 15 Pro - 2024-01-11 at 14 12 20 | Simulator Screenshot - iPhone 15 Pro - 2024-01-11 at 14 08 57 |

### DIFF
--- a/packages/react-native-host/ReactNativeHost.podspec
+++ b/packages/react-native-host/ReactNativeHost.podspec
@@ -8,7 +8,7 @@ version = package['version']
 repository = package['repository']
 repo_dir = repository['directory']
 
-new_arch_enabled = ENV['RCT_NEW_ARCH_ENABLED'] == '1'
+new_arch_enabled = ENV['RCT_NEW_ARCH_ENABLED'] == '1' && ENV['USE_NEW_ARCH'] != '0'
 preprocessor_definitions = [
   'FOLLY_CFG_NO_COROUTINES=1',
   'FOLLY_HAVE_CLOCK_GETTIME=1',
--- a/.changeset/late-kids-fry.md
+++ b/.changeset/late-kids-fry.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/react-native-host": patch
+---
+
+Handle refactorings in 0.74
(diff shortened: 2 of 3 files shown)


packages/react-native-host/ReactNativeHost.podspec
```diff
@@ -8,7 +8,7 @@ version = package['version']
 repository = package['repository']
 repo_dir = repository['directory']
 
-new_arch_enabled = ENV['RCT_NEW_ARCH_ENABLED'] == '1'
+new_arch_enabled = ENV['RCT_NEW_ARCH_ENABLED'] == '1' && ENV['USE_NEW_ARCH'] != '0'
```
JasonVMo, line 11: What's the relationship between these two environment variables? Is one new and one old? As it is the first one can turn it on (RCT_NEW_ARCH_ENABLED) and the second can block it (USE_NEW_ARCH) but is that the intent?

tido64: Short story: Meta did a [change](https://github.com/facebook/react-native/commit/4a35f2512e21deacf3e26e811875a9b8e4388a58) which basically hardcoded the value of `RCT_NEW_ARCH_ENABLED=1`, then introduced `USE_NEW_ARCH` as a runtime flag to enable/disable new architecture.

Starting with 0.72, they did a bunch of work to no longer use `RCT_NEW_ARCH_ENABLED` to differentiate between old/new arch. With 0.74, this work is considered complete: the two will _always_ be built together. This means that both old and new arch will end up in the binaries. As part of that, they decided that `RCT_NEW_ARCH_ENABLED` should always be true. But the consequence is that it's breaking basically all third party libraries out there. We're in talks with Meta to unbreak `RCT_NEW_ARCH_ENABLED`.

JasonVMo (approved):

PR #2939, opened 2024-01-29
fix(react-native-host): Add visionOS as a supported platform

Saadnajmi:
### Description

Add visionOS as a supported platform to ReactNativeHost, so that react-native-test-app can support visionOS. React Native host doesn't import UIKit anywhere, so I don't expect any of its' code to fail to compile on visionOS.

### Test plan

Once react native test app is unblocked, I can update the test app here.


tido64 (approved):

tido64:
Thanks ❤️

PR #2946, opened 2024-02-01
ci: fix `@react-native-webapis` packages not being built

tido64:
### Description

Packages that don't have a `test` target are ignored when running `nx run-may --target test`, causing published packages to miss build artifacts.

### Test plan

1. Run `build:ci:all`
2. Verify that `@react-native-webapis/*` packages are included


kelset (approved):
Yup 👍 good catch

PR #2962, opened 2024-02-09
fix(metro-serializer-esbuild): include `ReactFabric-prod.js` by default

tido64:
### Description

Include `ReactFabric-prod.js` by default.

### Test plan

n/a


kelset (approved):
:shipit:

PR #2984, opened 2024-02-22
feat(align-deps) allow exact version within range according to semver…

### LINKED ISSUE #2983: align-deps: new flag to not throw an error for exact dependencies that are within bounds
Both myself and the company I work for started using `@rnx-kit/align-deps` and it would be great to add it to CI as well, potentially even blocking PR merge if it throws an error. However we use exact versions for certain dependencies, which even if are within the "bounds" of what align-deps checks for results in an error for not using the exact sytnax for versioning.

While I understand that the current behavior should be the default as it is much more common in _React Native_ to not use exact versions, it would be awesome if the `rnx-align-deps` script would provide a flag which if specified, it woud still be checked that the version declared in `package.json` is within the bound but if it is, won't throw an error.

I've seen https://github.com/microsoft/rnx-kit/issues/2130 which is somewhat similar (though the reporter didn't really ellaborate on their use-case) and the suggestion there to use a custom preset/config, but in my (our) case we would very much love to use the community config which is a tremendous help in e.g. when updating to a newer _React Native_ version. If align-deps would throw an error if the declared version is outside of the bounds would be very useful in these cases, while going forward, if the dependency is kept within it, won't error out.

Thank you very much for considering my request!
Cheers!

balazsgerlei:
…Satisfies

### Description

Add a new option (flag) `--diff-mode` to allow specifying two modes for diffing:

- `strict`: Same as before
- `allow-exact-versions`: Allow (don't throw an error) when exact versions are declared in `package.json` that are within range according to `semverSatisfies`.

Resolves #2983 

Remaining tasks:

- [x] Change to `diff-mode` options from a boolean flag
- [ ] Add information about what changed (_Changeset_)
- [ ] Add unit tests
- [ ] Update documentation

### Test plan

- CI should pass, including updated unit tests
- Declare a dependency to `package.json` with an _exact version_ that is within the range of what the built-in capabilities define. Run `align-deps` with the new  `--diff-mode` set to `allow-exact-version` and verify that **no error is thrown**.
- Declare a dependency to `package.json` with an _exact version_ that is within the range of what the built-in capabilities define. Run `align-deps` with the new  `--diff-mode` set to `allow-exact-version` and `--write` flag set and verify that **nothing changes** in package.json.
- Declare a dependency to `package.json` with an _exact version_ that is outside of the range of what the built-in capabilities define. Run `align-deps` with the new  `--diff-mode` set to `allow-exact-version` and verify that **an error is thrown** for not correct version.
- Declare a dependency to `package.json` with an _exact version_ that is within the range of what the built-in capabilities define. Run `align-deps` with the new  `--diff-mode` set to `strict` and verify that **an error is thrown** for not correct version.
- Declare a dependency to `package.json` with an _exact version_ that is within the range of what the built-in capabilities define. Run `align-deps` with the new  `--diff-mode` set to `strict` but **with** `--write` flag and verify that **the versions is overwritten** with what is defined in built-in capabilities.
- Declare a dependency to `package.json` with an _exact version_ that is outside of the range of what the built-in capabilities define. Run `align-deps` with the new  `--diff-mode` set to `strict` and verify that **an error is thrown** for not correct version.
- Declare a dependency to `package.json` with an _exact version_ that is within the range of what the built-in capabilities define. Run `align-deps` **without** the new  `--diff-mode` set and verify that **an error is thrown** for not correct version.
- Declare a dependency to `package.json` with an _exact version_ that is within the range of what the built-in capabilities define. Run `align-deps` **without** the new  `--diff-mode` set but **with** `--write` flag and verify that **the versions is overwritten** with what is defined in built-in capabilities.
- Declare a dependency to `package.json` with an _exact version_ that is outside of the range of what the built-in capabilities define. Run `align-deps` **without** the new  `--diff-mode` set and verify that **an error is thrown** for not correct version.

### DIFF
--- a/packages/align-deps/src/cli.ts
+++ b/packages/align-deps/src/cli.ts
@@ -15,7 +15,20 @@ import { makeSetVersionCommand } from "./commands/setVersion";
 import { defaultConfig } from "./config";
 import { printError, printInfo } from "./errors";
 import { isString } from "./helpers";
-import type { Args, Command } from "./types";
+import type { Args, Command, DiffMode } from "./types";
+
+function validateDiffMode(mode: string | undefined): DiffMode {
+  switch (mode) {
+    case "allow-exact-version":
+    case "strict":
+      return mode;
+    default:
+      if (mode) {
+        warn("Unknown diff mode:", mode);
+      }
+      return "strict";
+  }
+}
 
 export const cliOptions = {
   "exclude-packages": {
@@ -75,6 +88,12 @@ export const cliOptions = {
     default: false,
     description: "Writes changes to the specified 'package.json'.",
     type: "boolean",
+  },
+  "diff-mode": {
+    default: "strict",
+    description:
+      "Determines the diff mode that align-deps uses to check version declarations in 'package.json'. Valid values are 'strict', meaning only strict equality with capability requirements (usually a range) are allowed and 'allow-exact-version' which allows versions that are within the required range, making it possible to declare exact versions.",
+    choices: ["strict", "allow-exact-version"],
   },
 };
 
@@ -169,6 +188,7 @@ async function makeCommand(args: Args): Promise<Command | undefined> {
     "set-version": setVersion,
     verbose,
     write,
+    "diff-mode": diffMode,
   } = args;
 
   const options = {
@@ -180,6 +200,7 @@ async function makeCommand(args: Args): Promise<Command | undefined> {
     write,
     excludePackages: excludePackages?.toString()?.split(","),
     requirements: requirements?.toString()?.split(","),
+    diffMode: validateDiffMode(diffMode),
   };
 
   if (typeof init !== "undefined") {
--- a/packages/align-deps/src/diff.ts
+++ b/packages/align-deps/src/diff.ts
@@ -1,10 +1,14 @@
 import type { PackageManifest } from "@rnx-kit/tools-node/package";
 import { dependencySections } from "./helpers";
+import semverSatisfies from "semver/functions/satisfies";
 import type { Changes } from "./types";
+
+export type DiffMode = "strict" | "allow-exact-version";
 
 export function diff(
   manifest: PackageManifest,
-  updatedManifest: PackageManifest
+  updatedManifest: PackageManifest,
+  mode: DiffMode = "strict"
 ): Changes | undefined {
   const allChanges: Changes = {
     dependencies: [],
@@ -13,6 +17,7 @@ export function diff(
     capabilities: [],
   };
 
+  const allowExactVersion = mode === "allow-exact-version";
   const numChanges = dependencySections.reduce((count, section) => {
     const changes = allChanges[section];
     const currentDeps = manifest[section] ?? {};
@@ -23,7 +28,9 @@ export function diff(
       if (!current) {
         changes.push({ type: "added", dependency, target });
       } else if (current !== target) {
-        changes.push({ type: "changed", dependency, target, current });
+        if (!allowExactVersion || !semverSatisfies(current, target)) {
+          changes.push({ type: "changed", dependency, target, current });
+        }
       }
     }
 
--- a/packages/align-deps/src/types.ts
+++ b/packages/align-deps/src/types.ts
@@ -19,6 +19,8 @@ export type Changes = {
   capabilities: { type: "unmanaged"; dependency: string; capability: string }[];
 };
 
+export type DiffMode = "strict" | "allow-exact-version";
+
 export type Options = {
   presets: string[];
   loose: boolean;
@@ -28,6 +30,7 @@ export type Options = {
   write: boolean;
   excludePackages?: string[];
   requirements?: string[];
+  diffMode?: DiffMode;
 };
 
 export type Args = Pick<Options, "loose" | "verbose" | "write"> & {
@@ -39,6 +42,7 @@ export type Args = Pick<Options, "loose" | "verbose" | "write"> & {
   packages?: (string | number)[];
   presets?: string | number;
   requirements?: string | number;
+  "diff-mode"?: string;
 };
 
 export type DependencyType = "direct" | "development" | "peer";
--- a/packages/cli/src/align-deps.ts
+++ b/packages/cli/src/align-deps.ts
@@ -25,6 +25,7 @@ export function rnxAlignDeps(
     verbose: Boolean(args.verbose),
     write: Boolean(args.write),
     packages: argv,
+    "diff-mode": String(args.diffMode),
   });
 }
 
@@ -73,6 +74,10 @@ export const rnxAlignDepsCommand = {
     {
       name: "--write",
       description: cliOptions.write.description,
+    },
+    {
+      name: `--diff-mode [${cliOptions["diff-mode"].choices?.join("|")}]`,
+      description: cliOptions["diff-mode"].description,
     },
   ],
 };
--- a/packages/align-deps/src/commands/check.ts
+++ b/packages/align-deps/src/commands/check.ts
@@ -84,7 +84,7 @@ export function checkPackageManifest(
     kitType
   );
 
-  const allChanges = diff(manifest, updatedManifest);
+  const allChanges = diff(manifest, updatedManifest, options.diffMode);
   if (allChanges) {
     if (options.write) {
       // The config object may be passed to other commands, so we need to
--- a/packages/align-deps/test/check.app.test.ts
+++ b/packages/align-deps/test/check.app.test.ts
@@ -12,6 +12,7 @@ const defaultOptions = {
   noUnmanaged: false,
   verbose: false,
   write: true,
+  allowExactVersion: false,
 };
 
 function fixturePath(name: string) {
--- a/packages/align-deps/test/initialize.test.ts
+++ b/packages/align-deps/test/initialize.test.ts
@@ -11,6 +11,7 @@ const defaultOptions = {
   noUnmanaged: false,
   verbose: false,
   write: false,
+  diffMode: "strict",
 };
 
 describe("initializeConfig()", () => {
--- a/packages/align-deps/test/setVersion.test.ts
+++ b/packages/align-deps/test/setVersion.test.ts
@@ -38,6 +38,7 @@ describe("makeSetVersionCommand()", () => {
     noUnmanaged: false,
     verbose: false,
     write: false,
+    diffMode: "strict",
   };
 
   afterEach(() => {
--- a/packages/align-deps/test/vigilant.test.ts
+++ b/packages/align-deps/test/vigilant.test.ts
@@ -450,6 +450,7 @@ describe("checkPackageManifestUnconfigured()", () => {
     noUnmanaged: false,
     verbose: false,
     write: false,
+    diffMode: "strict",
   };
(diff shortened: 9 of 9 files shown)


balazsgerlei:
I marked the PR as Draft as I haven't finished with the implementation, e.g. I still need to add unit tests but I wanted to ask the project maintainers' opinion (especially @tido64 especially considering this is my first feature addition to this repo. I appreciate any help and suggestion!

packages/align-deps/src/diff.ts
```diff
@@ -1,17 +1,20 @@
 import { keysOf } from "@rnx-kit/tools-language";
 import type { PackageManifest } from "@rnx-kit/tools-node/package";
+import semverSatisfies from "semver/functions/satisfies";
 import type { Changes } from "./types";
 
 export function diff(
   manifest: PackageManifest,
-  updatedManifest: PackageManifest
+  updatedManifest: PackageManifest,
+  mode: "exact" | "allow-exact-version" = "exact"
```
balazsgerlei, line 9: I based this on your patch @tido64 but TBH I'm not sure "exact" is the right name for the default mode, especially that the other option has "exact" in its name as well. I don't have a better suggestion yet, but I'll try to come up with something. You know, naming is one of the hard problem in programming 😆.

kelset:
thanks for sending it over!
Super small thing, can you run [yarn change locally to generate the changesets](https://github.com/microsoft/rnx-kit/blob/main/CONTRIBUTING.md#change-logs) ? that will make CI go green

packages/align-deps/src/cli.ts
```diff
@@ -70,6 +70,12 @@ export const cliOptions = {
     description: "Writes changes to the specified 'package.json'.",
     type: "boolean",
   },
+  "allow-exact-version": {
+    default: false,
+    description:
+      "Determines whether align-deps should allow exact versions in 'package.json' that are within the range and not throw an error for them.",
+    type: "boolean",
+  },
 };
```
kelset, line 79: nit: once everything else is handled and we're good on the implementation, we should add mention of this in the readme: https://github.com/microsoft/rnx-kit/blob/main/packages/align-deps/README.md#usage

balazsgerlei:
> thanks for sending it over! Super small thing, can you run [yarn change locally to generate the changesets](https://github.com/microsoft/rnx-kit/blob/main/CONTRIBUTING.md#change-logs) ? that will make CI go green

Sure, that's definitely on my list. You know what, I'll update the PR description with checkboxes for the remaining tasks and if I forgot something you can call that out!

Getting back to changeset, what kind of version bump you think this change warrants?

kelset:
> Getting back to changeset, what kind of version bump you think this change warrants?

@tido64 any preferences? If we want to be close to the [semver definition](https://semver.org/#summary) I'd say we should go with minor:

> MINOR version when you add functionality in a backward compatible manner

since it's a new feature

packages/align-deps/src/cli.ts
```diff
@@ -70,6 +70,12 @@ export const cliOptions = {
     description: "Writes changes to the specified 'package.json'.",
     type: "boolean",
   },
+  "allow-exact-version": {
```
tido64, line 73: Speaking of names, I wonder if this should be `--diff-algorithm` with valid values being `strict-equality` and `allow-exact-version`.

(The rest looks fine, btw)

balazsgerlei: I like this approach more, if you look into the code it's converting back and forth between a boolean and the two modes so it makes sense to use these modes from the beginning (with a default value that keeps chrrent behavior).

Not to be too nitpicky (especially considering it's your library so OFC ultimately it's your choice) - what about `- - diff-mode`? It's hardly an algorithm IMHO 😅

tido64: I'm fine with `--diff-mode`. It's shorter and more concise 👍

tido64:
@balazsgerlei: Are you unblocked to continue working on this? If you don't have the time, let me know. I wouldn't mind taking over and pushing it over the finish line. 😄

balazsgerlei:
> @balazsgerlei: Are you unblocked to continue working on this? If you don't have the time, let me know. I wouldn't mind taking over and pushing it over the finish line. 😄 

I've been indeed very occupied with personal stuff so I had little to no free time after work to attend to this. It would be still a nice challenge to finish this so I'll try to allocate some time for it and if I couldn't do it this week I'll let you know, but thanks for offering to take over!

tido64:
> I've been indeed very occupied with personal stuff so I had little to no free time after work to attend to this. It would be still a nice challenge to finish this so I'll try to allocate some time for it and if I couldn't do it this week I'll let you know, but thanks for offering to take over!

Take the time you need to finish this. No pressure. Just making sure you're not waiting to hear from us 😉

balazsgerlei:
Okay, it took a while, I've finally had some time to update this. 

I implemented the two distinct diff mode options, what I'm not sure about is how and where to handle if the option is not set correctly. It should not happen since the only two valid options are delcared in cli.ts though I feel like it should still be handled in e.g. check.ts. I could add a simple `isDIffMode` function similarly to `isKitType` in `initialize.ts` before diff is called in `checkPackageManifest` but I don't see any `ErrorCode` (which is what `checkPackageManifest` returns) that would be a good fit and I was hesitant to add a new `ErrorCode` for this, though that might be the best choice.

Can I ask for some guidence on this @tido64 @kelset ? Thank you!

kelset:
Hey @balazsgerlei Tommy is currently on a break; I'll throw in my 2C$ in the meantime.

Because we are setting the options via `.choices` from [yargs directly](https://yargs.js.org/docs/#api-reference-choiceskey-choices) I don't think we necessarily need to handle the error case you are mentioning directly because it should be handled by yargs itself.

What happens currently if you try to run the command and passing an invalid option?

balazsgerlei:
> Hey @balazsgerlei Tommy is currently on a break; I'll throw in my 2C$ in the meantime.
> 
> Because we are setting the options via `.choices` from [yargs directly](https://yargs.js.org/docs/#api-reference-choiceskey-choices) I don't think we necessarily need to handle the error case you are mentioning directly because it should be handled by yargs itself.
> 
> What happens currently if you try to run the command and passing an invalid option?

The available options are listed and at the end it also tells you what you did wrong:

> Invalid values:
  Argument: diff-mode, Given: "xyz", Choices: "strict", "allow-exact-version"

That's what I also meant is that when using the script in Terminal you basically cannot pass the wrong argument to the actual logic, just that it would be more precise to handle that case somewhere e.g. in `check.ts`. That said, I'm more than happy to move forward like this, I would really like if I/we can get to the finish line. 😄 

So I think I will continue with the unit tests like this and in the time being, let me know if you have another solution in mind or if @tido64 comes back and has a different opinion!

packages/align-deps/src/cli.ts
```diff
@@ -180,6 +187,7 @@ async function makeCommand(args: Args): Promise<Command | undefined> {
     write,
     excludePackages: excludePackages?.toString()?.split(","),
     requirements: requirements?.toString()?.split(","),
+    diffMode,
```
tido64, line 190: I think this would be a good place to validate that we have a correct value (or none) for diff mode:
```suggestion
    diffMode: validateDiffMode(diffMode),
```

packages/cli/src/align-deps.ts
```diff
@@ -25,6 +25,7 @@ export function rnxAlignDeps(
     verbose: Boolean(args.verbose),
     write: Boolean(args.write),
     packages: argv,
+    "allow-exact-version": Boolean(args.allowExactVersion),
```
tido64, line 28: Shouldn't this be `diff-mode`?

packages/align-deps/src/diff.ts
```diff
@@ -1,17 +1,22 @@
 import { keysOf } from "@rnx-kit/tools-language";
 import type { PackageManifest } from "@rnx-kit/tools-node/package";
+import semverSatisfies from "semver/functions/satisfies";
 import type { Changes } from "./types";
 
+export type DiffMode = "strict" | "allow-exact-version";
```
tido64, line 6: We should move this into `./types.ts`

packages/align-deps/src/types.ts
```diff
@@ -27,6 +27,7 @@ export type Options = {
   write: boolean;
   excludePackages?: string[];
   requirements?: string[];
+  diffMode?: string;
```
tido64, line 30: I think we should have the correct type by the time this gets created:
```suggestion
  diffMode?: DiffMode;
```

packages/cli/src/align-deps.ts
balazsgerlei: Can you help me a little bit here? What does this declaration do? Does diff-mode need to be listed here to work? And what it's type should be. I'm especially confused as it was working with this still being incorrectly allow-exact-version so maybe it shouldn't be added here? Thx!

packages/align-deps/src/cli.ts
balazsgerlei: Can you help me a little bit with the validation though?

My initial idea was (when I brought up validation earlier) was:
```ts
function isDiffMode(mode: string): mode is DiffMode {
  return mode === "strict" || mode === "allow-exact-version";
}
```

However this expects a string and we have `string | undefined` here.

tido64: I would do something like this:

```ts
function validateDiffMode(mode: string | undefined): DiffMode {
  switch (mode) {
    case "allow-exact-version":
    case "strict":
      return mode;
    default:
      if (mode) {
        warn("Unknown diff mode:", mode);
      }
      return "strict";
  }
}
```

packages/cli/src/align-deps.ts
tido64: align-deps is standalone. You can depend on this one package and get going. `@rnx-kit/cli` is out drop-in replacement for `@react-native-community/cli`, and it kinda includes "everything" the repo has to offer. One of its commands, `align-deps`, calls into `align-deps`. We should reuse the type to ensure these are always up-to-date, but for now, let's just sync the config here.

packages/align-deps/src/cli.ts
balazsgerlei: Thanks, applied it via rebase.

timostroehlein:
Thank you for the PR, we also need that feature in our team. I was about to implement that myself until i saw this PR 😄.

I have one small suggestion that would make this ideal for us, currently the `diff-mode` `allow-exact-version` only checks for the exact version, as the name suggests. It would be great if there would be another option, e.g. `allow-in-range`, which would allow all versions which satisfy the version range of the capability.

The following would be valid then:
capability: `^0.73.0`
valid: `^0.73.1` `^0.73.9` `0.73.2`
invalid: `^0.74.0` `0.74.0`

tido64:
It's been some time since this has received any updates so I'll bring it to the finish line. The new PR will also address @timostroehlein's request.

balazsgerlei:
> It's been some time since this has received any updates so I'll bring it to the finish line. The new PR will also address @timostroehlein's request.
> 
> Thanks for all your efforts!

Hi! Thank you and I'm sorry I could not carve out time to finish my PR. 😢 
I'm glad that the feature got added though and I'll use it in the future for sure!

PR #2989, opened 2024-02-27
fix(web-storage): migrate away from `com.facebook.react.turbomodule.core.interfaces.TurboModule`

tido64:
### Description

Migrate away from `com.facebook.react.turbomodule.core.interfaces.TurboModule` as it has been moved in 0.74 and will break TurboModule detection.

### Test plan

```
cd packages/test-app
yarn build --dependencies
yarn android

# In a separate terminal
yarn start
```

### DIFF
--- a/.changeset/early-zebras-laugh.md
+++ b/.changeset/early-zebras-laugh.md
@@ -1,0 +1,5 @@
+---
+"@react-native-webapis/web-storage": patch
+---
+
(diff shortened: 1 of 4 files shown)


kelset:
tested against RNTA (didn't think about the fact that via the other PR we basically have the same thing here lol)

I can confirm it does the trick, tested on RNTA on nightly.

Before (manually applying) the fix:
Screenshot 2024-02-28 at 13 00 57

After:
Screenshot 2024-02-28 at 13 04 56

kelset (approved):
:shipit:

PR #2995, opened 2024-03-01
fix(metro-config): handle apps in nested folders

tido64:
### Description

Handle apps in nested folders e.g.:

```
my-awesome-component
├── example  <-- Note no `package.json`
│   └── App.js
├── package.json
└── src
```

### Test plan

n/a


kelset (approved):
:shipit:

PR #3007, opened 2024-03-15
ci: bump Node to v20 to satisfy Yarn

tido64:
### Description

Publishing CI currently fails:

```
/usr/local/bin/yarn --version
Usage Error: This tool requires a Node version compatible with >=18.12.0 (got 16.20.2). Upgrade Node, or set `YARN_IGNORE_NODE=1` in your environment.
```

### Test plan

n/a


incubator/chromium-edge-launcher/package.json
tido64, line 1: @acoates-ms: Can we remove this package now that your patch was accepted?

kelset (approved):
LGTM :shipit:

PR #3030, opened 2024-03-23
fix: backticks in README

Rexogamer:
### Description

this PR fixes a mismatched backtick in the main README, which was causing the text to not render properly.

### Test plan

N/A


kelset (approved):
nice catch, thanks!

PR #3036, opened 2024-03-27
docs(react-native-host): update example to prevent premature dealloc

tido64:
### Description

Updated the example to make it clear that `ReactNativeHost` needs to be retained.

### Test plan

n/a


kelset (approved):
good stuff 👍

PR #3042, opened 2024-04-02
fix(babel-preset): deprecate `looseClassTransform`

tido64:
### Description

Deprecate `looseClassTransform` and redirect to compiler assumptions.

### Test plan

n/a


kelset (approved):
good one 👍

PR #3043, opened 2024-04-02
fix(console): inline colors and drop `chalk`

tido64:
### Description

Inline the colors we use and drop `chalk`

### Test plan

image


kelset (approved):
Love this type of improvement 👍

PR #3046, opened 2024-04-02
fix(typescript-service): unset unused loggers

tido64:
### Description

I'm not entirely sure what the loggers are used for. It doesn't look like they get used upstream either: https://github.com/search?q=repo%3Amicrosoft%2FTypeScript+createLanguageService&type=code

### Test plan

```sh
cd packages/test-app
yarn build --dependencies
yarn start

# In a separate terminal, run the following command every time you make changes to `App.native.tsx`
curl -v 'http://localhost:8081/index.bundle?platform=ios' > /dev/null
```

Example output:

image

### DIFF
--- a/.changeset/late-cheetahs-raise.md
+++ b/.changeset/late-cheetahs-raise.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/typescript-service": patch
+---
+
+Unset unused loggers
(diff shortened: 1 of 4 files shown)


kelset (approved):
yeah not sure why we had them in the first place for `languageServiceHost` - maybe some leftover old code.

PR #3073, opened 2024-04-05
feat(align-deps): Add core-visionos capability

Saadnajmi:
### Description

Add a `core-visionos` capability, and update the packages in the 0.73 profile to recommend versions with visionOS support.  

### Test plan

CI should pass


packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
```diff
@@ -126,7 +131,7 @@ export const profile: Profile = {
   },
   svg: {
     name: "react-native-svg",
-    version: "^14.0.0",
+    version: "^15.0.0",
```
kelset, line 134: this is a major bump, why is it needed?

Saadnajmi: 15.0.0 is the first version that supports visionOS

Saadnajmi: So I thought I could make something like `profile-0.73-visionos` but that doesn't seem to be a thing. To update internal repos (which error if align-deps isn't happy), I wanted to add the `core-visions` capability and have it recommend the newer packages. It seems that not how capabilities work?

Saadnajmi: Related, I've been holding off on this PR till some of my other PRs to add visionOS support land. But to unblock, I may split off the `0.73` profile updates to a separate PR

Saadnajmi: https://github.com/microsoft/rnx-kit/pull/3085/

packages/align-deps/README.md
```diff
@@ -294,6 +294,7 @@ resolve to:
...
+| core-visionos                        | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               | Meta package for installing `react`                               |
```
tido64, line 297: Shouldn't this point to `react-native-visionos` at some point?

Saadnajmi: Hmm. I think this is because I put `#meta` for `profile-0.61.ts`

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
```diff
@@ -33,6 +33,11 @@ export const profile: Profile = {
     version: "^0.73.0",
     capabilities: ["react"],
   },
+  "core-visionos": {
+    name: "#meta",
```
tido64, line 37: Did we agree to use Callstack's as default?
```suggestion
    name: "@callstack/react-native-visionos",
```

packages/align-deps/src/presets/microsoft/react-native/profile-0.74.ts
```diff
@@ -33,6 +33,11 @@ export const profile: Profile = {
     version: "^0.74.0",
     capabilities: ["react"],
   },
+  "core-visionos": {
+    name: "#meta",
```
tido64, line 37:
```suggestion
    name: "@callstack/react-native-visionos",
```

tido64 (changes requested):

packages/test-app/package.json
```diff
@@ -197,6 +197,7 @@
         "core-android",
         "core-ios",
         "core-macos",
+        "core-visionos",
```
tido64, line 200: Reverting this change should make CI happy:
```suggestion
```

tido64 (approved):

packages/test-app/package.json
Saadnajmi: Oops, sorry I missed it. Thanks for checking!

PR #3077, opened 2024-04-09
docs(changeset): fix findWorkspacePackagesSync lerna 7+ incompatibility

weizs:
### Description
Sync findWorkspacePackages Compatibility Adapter for lerna 7+ to findWorkspacePackagesSync


weizs:

PR #3078, opened 2024-04-09
fix(tools-workspaces): fix findWorkspacePackagesSync lerna 7+ incompatibility

weizs:
### Description
Sync findWorkspacePackages Compatibility Adapter for lerna 7+ to findWorkspacePackagesSync


tido64 (approved):
Thanks for fixing this!

PR #3085, opened 2024-04-12
chore(align-deps): Update `profile-0.73` to recommend visionOS compatible versions

Saadnajmi:
### Description

Update our 0.73 profile to recommend the latest version of a couple of packages that have added visionOS support. Given 0.74 is not out yet, these versions should be safe to update to. 

### Test plan

CI should pass. I confirmed an internal project that depends on these projects builds (at least for the apple platforms).


.changeset/chilled-ghosts-think.md
```diff
@@ -0,0 +1,45 @@
...
+"@rnx-kit/tools-workspaces": patch
+"@rnx-kit/tools-language": patch
+"@rnx-kit/eslint-config": patch
+"@rnx-kit/eslint-plugin": patch
+"@rnx-kit/metro-service": patch
+"@rnx-kit/metro-config": patch
+"@rnx-kit/bundle-diff": patch
+"@rnx-kit/jest-preset": patch
+"@rnx-kit/polyfills": patch
+"@rnx-kit/align-deps": patch
```
Saadnajmi, line 35: I only updated align-deps.. I assume all the others get patched because they depend on align-deps.

tido64: Is this happening because you have to remotes in your clone. It looks like Changesets is comparing against your `main` instead of upstream.

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
```diff
@@ -110,7 +114,7 @@ export const profile: Profile = {
   },
   netinfo: {
     name: "@react-native-community/netinfo",
-    version: "^11.0.1",
+    version: "^11.3.1",
```
tido64, line 117: Strictly speaking, this should be 11.3.0. If we keep 11.3.1, we should remove it from the 0.74 profile:
```suggestion
    version: "^11.3.0",
```

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
```diff
@@ -49,6 +49,10 @@ export const profile: Profile = {
     version: "^0.73.0",
     devOnly: true,
   },
+  clipboard: {
+    name: "@react-native-clipboard/clipboard",
+    version: "^1.10.0",
```
tido64, line 54: Shouldn't this be 1.14? Also, can you move it to the bottom section?
```suggestion
    version: "^1.14.0",
```

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
```diff
@@ -122,11 +126,11 @@ export const profile: Profile = {
   },
   storage: {
     name: "@react-native-async-storage/async-storage",
-    version: "^1.21.0",
+    version: "^1.23.0",
```
tido64, line 129:
```suggestion
    version: "^1.22.0",
```

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
```diff
@@ -122,11 +126,11 @@ export const profile: Profile = {
...
   },
   svg: {
     name: "react-native-svg",
-    version: "^14.0.0",
+    version: "^15.0.0",
```
tido64, line 133: This seems to bump minimum iOS version to 12.4. Are we okay with a breaking change on the 0.73 profile?

kelset: I raised the same concern here https://github.com/microsoft/rnx-kit/pull/3073#discussion_r1555550313 and I don't think we should do this breaking change. If internal codebases need the major bump, that should be handled with a custom local profile that overrides the default profile.

.changeset/chilled-ghosts-think.md
Saadnajmi: Oh yeah, Beachball handles that case gracefully not it seems changesets does not. I'll update

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
Saadnajmi: Isn't the minimum OS version of 0.73 already iOS 13.4?

tido64: That's right. So this should be fine then.

kelset: good catch - can confirm the bump: https://github.com/facebook/react-native/commit/610b14e4f3f606cd4c49518b9a42e3290fd52aeb

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
Saadnajmi: Derp, yeah. Thank you for catching all of my typos. I must have committed the wrong version as I was looking for where to put this. I thought we'd want it alphabetical but I can move to bottom.

.changeset/chilled-ghosts-think.md
Saadnajmi: Updated.

Saadnajmi:
Sorry, my last merge of upstream/main seemed to break CI, so I did a rebase/force push, hoping that would fix it

tido64 (approved):
Thanks ❤️

PR #3101, opened 2024-04-18
fix(rnta-msal): address deprecations

tido64:
### Description

Migrate away from deprecated `acquireTokenSilent(Array<(out) String!>, IAccount, String): IAuthenticationResult!`

### Test plan

n/a


kelset (approved):
LGTM

PR #3110, opened 2024-04-25
feat(tools-apple): introduce new package for Apple tools

tido64:
### Description

`@rnx-kit/tools-apple` is a collection of functions for deploying apps on iOS or macOS

### Test plan

n/a


packages/tools-apple/src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  available: boolean;
+  platform:
+    | "com.apple.platform.appletvos"
+    | "com.apple.platform.appletvsimulator"
+    | "com.apple.platform.driverkit"
+    | "com.apple.platform.iphoneos"
+    | "com.apple.platform.iphonesimulator"
+    | "com.apple.platform.macosx"
+    | "com.apple.platform.watchos"
+    | "com.apple.platform.watchsimulator";
```
Saadnajmi, line 14: Can we add visionOS too? Should be `xros` and `xrsimulator`

kelset (approved):
:shipit:

PR #3112, opened 2024-04-25
chore(align-deps): Recommend datetime-picker version compatible with new architecure & visionOS

Saadnajmi:
### Description

https://github.com/react-native-datetimepicker/datetimepicker/releases/tag/v8.0.0

This version supports the new architecture on 073+, and visionOS. Let's update our 0.73 profile preset

### Test plan

CI should pass.


.changeset/little-tips-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rnx-kit/align-deps": patch
+---
+
+git push
```
tido64, line 5: Can you write a better change log entry than this? 😆

packages/align-deps/src/presets/microsoft/react-native/profile-0.73.ts
```diff
@@ -102,7 +102,7 @@ export const profile: Profile = {
   },
   "datetime-picker": {
     name: "@react-native-community/datetimepicker",
-    version: "^7.7.0",
+    version: "^8.0.0",
```
tido64, line 105: For profile 0.72 (and below), I think we need to change them to `>=7.7.0 <8.0.0`

image

.changeset/little-tips-jog.md
Saadnajmi: oof

tido64 (approved):
Thanks ❤️

PR #3124, opened 2024-05-02
fix(react-native-host): prevent unintentional autolinking

tido64:
### Description

Prevent unintentional autolinking.

### Test plan

```
cd packages/test-app
yarn react-native config
```

Verify that `@rnx-kit/react-native-host` is not present.


kelset (approved):
:shipit:

PR #3130, opened 2024-05-06
fix(react-native-host): fix `SurfaceRegistryBinding::startSurface` failed

tido64:
### Description

Fix `SurfaceRegistryBinding::startSurface` failed on startup with New Architecture

### Test plan

See repro steps in https://github.com/microsoft/react-native-test-app/issues/2015


kelset (approved):
It works ([details here](https://github.com/microsoft/react-native-test-app/issues/2015))! Let's ship 👍

PR #3167, opened 2024-05-29
fix(tools-shell): allow `once` functions to take arguments

tido64:
### Description

Allow `once` functions to take arguments

### Test plan

n/a

### DIFF
--- a/.changeset/spicy-rabbits-smash.md
+++ b/.changeset/spicy-rabbits-smash.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/tools-shell": patch
+---
+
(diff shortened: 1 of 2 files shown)


JasonVMo (approved):
Looks good, though if I'm reading it right it does look like once<> won't work right with functions that return undefined as a result. Not related to this change, just an observation though maybe that is by design.

tido64:
> Looks good, though if I'm reading it right it does look like once<> won't work right with functions that return undefined as a result. Not related to this change, just an observation though maybe that is by design.

No, I think you're right. I've fixed it by adding `NonNullable`:

image

PR #3168, opened 2024-05-29
fix(build): fix `rnx-build-apple` not being resolved

tido64:
### Description

Fix `rnx-build-apple` not being resolved correctly in pnpm setups

### Test plan

n/a

### DIFF
--- a/.changeset/strange-donuts-beg.md
+++ b/.changeset/strange-donuts-beg.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/build": patch
+---
+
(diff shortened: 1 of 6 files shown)


JasonVMo (approved):
This looks reasonable, though should the /bin/bash actually be /bin/sh, or are the scripts written specifically for bash shells?

tido64:
> This looks reasonable, though should the /bin/bash actually be /bin/sh, or are the scripts written specifically for bash shells?

We're using some convenience features that only exist in `/bin/bash` (and compatible shells).

PR #3171, opened 2024-05-30
feat(build): implement OAuth device flow

tido64:
### Description

Implement [OAuth device flow](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#device-flow).

### Test plan

```sh
cd incubator/build
yarn build --dependencies
yarn rnx-build -p ios --device-type simulator ../../packages/test-app
```

- `rnx-build` should ask the user to type in a code in a browser
- Build should continue once it's done
- Build will fail because Microsoft does not allow OAuth apps by default (see https://github.com/microsoft/rnx-kit/tree/main/incubator/build#personal-access-token for how to create a PAT)
- If the access token has expired, we should try fetching a new one (test by revoking the token created above)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2828,6 +2828,18 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@octokit/auth-oauth-device@npm:^7.1.1":
+  version: 7.1.1
+  resolution: "@octokit/auth-oauth-device@npm:7.1.1"
+  dependencies:
+    "@octokit/oauth-methods": "npm:^5.0.0"
+    "@octokit/request": "npm:^9.0.0"
+    "@octokit/types": "npm:^13.0.0"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -2886,17 +2910,17 @@ __metadata:
   linkType: hard
 
 "@octokit/plugin-rest-endpoint-methods@npm:^13.0.0":
-  version: 13.2.2
-  resolution: "@octokit/plugin-rest-endpoint-methods@npm:13.2.2"
+  version: 13.2.1
```
tido64, line 2913: This is an intentional downgrade. 13.2.2 contains a regression for ESM users.

JasonVMo (approved):
Looks good as far as I can tell though I'm no expert on oauth flows.

PR #3172, opened 2024-05-30
feat(build): redefine plugin interface

tido64:
### Description

Redefine plugin interface so that one plugin can serve multiple purposes

### Test plan

n/a


incubator/build-plugin-firebase/src/index.ts
```diff
@@ -29,12 +31,17 @@ function validateConfig(config: Partial<FirebaseConfig>): void {
...
 }
 
-// `export default` required for plugin interface
-// eslint-disable-next-line no-restricted-exports
-export default function (config: Partial<FirebaseConfig>): DistributionPlugin {
-  validateConfig(config);
+export const distribution: PluginInterface["distribution"] = (
+  config?: FirebaseConfig
+) => {
+  if (!validateConfig(config)) {
```
JasonVMo, line 41: Small nit but validateConfig basically returns true unless it throws an error, while this checks the result and throws and error. I suppose this is future proofing in the case where it eventually returns false without throwing but as it stands today this error will never get triggered.

incubator/build/src/types.ts
```diff
@@ -71,8 +71,13 @@ export type Remote = {
...
     context: Context,
     inputs: BuildParams,
-    spinner: Ora
+    logger: Ora
   ): Promise<string | null>;
   cancelBuild(context: Context): Promise<void>;
   install(): Promise<number>;
 };
+
+export type PluginInterface = Partial<{
```
JasonVMo, line 80: I'm curious why this is a partial. It looks like the distribution entry below is required if I'm reading the code above. Wouldn't it be more accurate to just add a ? to the remote options to denote that as optional?

JasonVMo (commented):
Only one real comment, the other is more of a nit you can choose to ignore.

incubator/build-plugin-firebase/src/index.ts
tido64: The if-check is here to help TypeScript to understand that `config` is not `undefined`:

| Nullable | Nonnullable |
| :-: | :-: |
| image | image |

incubator/build/src/types.ts
tido64: What I wanted to express here is that at least one of these are nonnull. You can implement a distribution plugin or a remote plugin, or both. To be completely accurate, one would have to write something like:

```typescript
export type PluginInterface =
  | { distribution: (options?: JSObject) => DistributionPlugin }
  | { remote: Remote }
  | {
      distribution: (options?: JSObject) => DistributionPlugin;
      remote: Remote;
    };
```

But this isn't scalable if we make more things pluginable (like platforms or source control). In terms of usage, we need to check whether any of these are set before we can use them in either case, and that's why I thought it's simpler to use `Partial` here.

JasonVMo (approved):

PR #3180, opened 2024-06-03
fix(third-party-notices): use license URL if no copyright info is found

AAHAbbas:
### Description
If no copyright or package author information is found then set license and license URL information instead of the default "No copyright notice" placeholder. This ensures that the JSON and text information is more similar. 

### Test plan
Unit tests

| Before | After  |
| ------ | --------- |
| <pre>{<br>  "name": "deprecated-react-native-prop-types",<br>  "version": "2.3.0",<br>  "license": "MIT",<br>  "copyright": "No copyright notice"<br>}</pre>    | <pre>{<br>  "name": "deprecated-react-native-prop-types",<br>  "version": "2.3.0",<br>  "license": "MIT",<br>  "copyright": "MIT (https://spdx.org/licenses/MIT.html)"<br>}</pre>|

### DIFF
--- a/packages/third-party-notices/src/output/json.ts
+++ b/packages/third-party-notices/src/output/json.ts
@@ -27,7 +27,7 @@ function parseCopyright(
       return packageAuthor;
     }
 
-    if (licenseURLs.length > 0) {
+    if (licenseURLs && licenseURLs.length > 0) {
       return `${license} (${licenseURLs.join(" ")})`;
     }
 
--- a/packages/third-party-notices/src/output/json.ts
+++ b/packages/third-party-notices/src/output/json.ts
@@ -16,11 +16,22 @@ function getPackageAuthor(modulePath: string): string | undefined {
 
 function parseCopyright(
   modulePath: string,
-  licenseText: string | undefined
+  licenseText: string | undefined,
+  license: string | undefined,
+  licenseURLs: string[]
 ): string {
   const m = licenseText?.match(/^Copyright .*$/m);
   if (!m) {
-    return getPackageAuthor(modulePath) || "No copyright notice";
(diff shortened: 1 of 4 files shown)


packages/third-party-notices/src/output/json.ts
```diff
@@ -16,11 +16,22 @@ function getPackageAuthor(modulePath: string): string | undefined {
 
 function parseCopyright(
   modulePath: string,
-  licenseText: string | undefined
+  licenseText: string | undefined,
+  license: string | undefined,
+  licenseURLs: string[]
```
tido64, line 21: Is `licenseURLs` always defined?

AAHAbbas: I assumed it will: https://github.com/microsoft/rnx-kit/blob/bad9c1e7782e873fe02c0fe893bd4bc8817eb699/packages/third-party-notices/src/extractors.js#L60C1-L61C1. I can add a check just to be safe.

packages/third-party-notices/src/output/json.ts
```diff
@@ -16,11 +16,22 @@ function getPackageAuthor(modulePath: string): string | undefined {
...
 ): string {
   const m = licenseText?.match(/^Copyright .*$/m);
   if (!m) {
-    return getPackageAuthor(modulePath) || "No copyright notice";
+    const packageAuthor = getPackageAuthor(modulePath);
+    if (packageAuthor) {
+      return packageAuthor;
+    }
+
+    if (licenseURLs && licenseURLs.length > 0) {
```
tido64, line 30:
```suggestion
    if (licenseURLs?.length > 0) {
```

tido64 (approved):

PR #3184, opened 2024-06-06
feat(align-deps): allow version range subsets

### LINKED ISSUE #2983: align-deps: new flag to not throw an error for exact dependencies that are within bounds

tido64:
### Description

Allow exact and version range subsets with `--diff-mode allow-subset`.

Resolves #2983
Resolves #2984

### Test plan

n/a

### DIFF
--- a/packages/align-deps/src/cli.ts
+++ b/packages/align-deps/src/cli.ts
@@ -21,7 +21,7 @@ export const cliOptions = {
   "diff-mode": {
     default: "strict",
     description:
-      "Determines what align-deps should do when versions differ. Valid values are 'strict' (version strings must be equal) and 'allow-subset' (allow version ranges that are a subset).",
+      "Sets the algorithm used to determine if versions differ. Valid values are 'strict' (version strings must be equal) and 'allow-subset' (allow version ranges that are a subset).",
     choices: ["strict", "allow-subset"],
   },
   "exclude-packages": {
--- a/packages/align-deps/src/cli.ts
+++ b/packages/align-deps/src/cli.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { error } from "@rnx-kit/console";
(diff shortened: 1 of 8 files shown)


packages/align-deps/src/cli.ts
```diff
@@ -15,9 +15,15 @@ import { makeSetVersionCommand } from "./commands/setVersion";
 import { defaultConfig } from "./config";
 import { printError, printInfo } from "./errors";
 import { isString } from "./helpers";
-import type { Args, Command } from "./types";
+import type { Args, Command, DiffMode } from "./types";
 
 export const cliOptions = {
+  "diff-mode": {
```
tido64, line 21: New flag needs to be documented in `README`

tido64:
> Looks like a good change. One question is whether this will lead to more duplicate packages within the repo and whether we need to do something to counteract that.

It's more likely that dupes will be introduced. I've added a warning label in the docs.

PR #3220, opened 2024-07-11
[align-deps] Update react-native-svg min versions

marcshilling:
Referencing [react-native-svg's supported versions](https://github.com/software-mansion/react-native-svg#supported-react-native-versions), updating some values here.

I'm on RN 0.72 for example and react-native-svg 15.4.0 works fine for me, but align-deps tells me I need to be using ^13.14.0


marcshilling:

marcshilling:
Maybe I was wrong. Having some Android build issues on CI after upgrading react-native-svg. Will revisit if necessary.

PR #3223, opened 2024-07-17
fix(eslint-plugin): handle packages missing `default` in their exports map

tido64:
### Description

Properly handle packages that are missing a default entry in their exports map.

### Test plan

1. In [Fluent UI](https://github.com/microsoft/fluentui), create a file, `packages/react-components/react-components/src/icons.ts`, with the following content:
    ```ts
    export * from "@fluentui/react-icons";
    ```
2. Run ESLint: `yarn workspace @fluentui/react-components eslint --ext .ts src/icons.ts`
3. Verify that at least 1 error is fixable with `--fix`

### DIFF
--- a/.changeset/twelve-fireants-hammer.md
+++ b/.changeset/twelve-fireants-hammer.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/eslint-plugin": patch
+---
+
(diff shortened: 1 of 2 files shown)


Hotell:
`🚢.it()`

thx a lot @tido64 !

JasonVMo (approved):

PR #3241, opened 2024-07-24
chore: refactor `build-android` command to use common spawn options

tido64:
### Description

Refactor `build-android` command to use common spawn options

### Test plan

CI should pass.


scripts/src/process.js
```diff
@@ -25,24 +18,28 @@ function isRunningNx(command, args) {
   return command.startsWith("yarn") && args[0] === "nx";
 }
 
+function workspaceRoot() {
+  return fileURLToPath(new URL("../../", import.meta.url));
```
JasonVMo, line 22: Is this a reliable way to get to the workspace root? This approach feels fragile.

JasonVMo (approved):

scripts/src/process.js
JasonVMo: Resolving based on discussion.

PR #3244, opened 2024-07-25
feat(cli)!: introduce standalone `rnx-cli`

tido64:
### Description

Introduce standalone `rnx-cli`.

The minimum required Node version has also been bumped to 16.17.

### Test plan

CI should pass.

The following commands aren't covered by CI and must be tested manually:

- `yarn rnx doctor`
- `yarn rnx info`


packages/cli/package.json
```diff
@@ -50,11 +53,9 @@
     "@rnx-kit/tools-language": "^2.0.0",
     "@rnx-kit/tools-node": "^2.1.1",
     "@rnx-kit/tools-react-native": "^1.3.5",
-    "fs-extra": "^10.0.0",
-    "node-fetch": "^2.6.7",
+    "commander": "^11.1.0",
```
tido64, line 56: We're not using latest 12.x because they've dropped support for Node 16 and we've said we'll support >=16.

acoates-ms (approved):

PR #3255, opened 2024-07-30
fix(config): fix modules not being resolved from correct path

tido64:
### Description

Fix module not being resolved from correct path

### Test plan

CI should pass.


packages/config/package.json
```diff
@@ -42,7 +42,6 @@
   },
   "devDependencies": {
     "@rnx-kit/eslint-config": "*",
-    "@rnx-kit/jest-preset": "*",
```
Saadnajmi, line 45: duplicate change from https://github.com/microsoft/rnx-kit/pull/3252/files ?

packages/config/package.json
```diff
@@ -42,7 +42,6 @@
...
     "@rnx-kit/metro-plugin-cyclic-dependencies-detector": "*",
```
Saadnajmi, line 45: Sidenote, Should all these be `workspace:*`?

packages/config/package.json
tido64: No, I left this package out:
image

packages/config/package.json
tido64: Actually, I'm not sure whether npm will fail if it encounters `workspace:`. Regardless of whether it's under `devDependencies`. Reverting until we can figure that out.

PR #3261, opened 2024-08-06
feat(tools-react-native): add caching layer on top of `loadConfig`

tido64:
### Description

Add caching layer on top of `loadConfig`. This will be used in a number of modules, including in https://github.com/microsoft/rnx-kit/pull/3244. And since the caching code was lifted from [RNTA](https://github.com/microsoft/react-native-test-app/blob/trunk/android/autolink.mjs), we'll have to replace it there as well.

### Test plan

Tests were added.


packages/tools-react-native/README.md
```diff
@@ -24,14 +24,16 @@ import * as platformTools from "@rnx-kit/tools-react-native/platform";
...
 
-| Category | Function                                               | Description                                                                                     |
-| -------- | ------------------------------------------------------ | ----------------------------------------------------------------------------------------------- |
-| metro    | `findMetroPath(projectRoot)`                           | Finds the installation path of Metro.                                                           |
-| metro    | `getMetroVersion(projectRoot)`                         | Returns Metro version number.                                                                   |
-| platform | `expandPlatformExtensions(platform, extensions)`       | Returns a list of extensions that should be tried for the target platform in prioritized order. |
-| platform | `getAvailablePlatforms(startDir)`                      | Returns a map of available React Native platforms. The result is cached.                        |
-| platform | `getAvailablePlatformsUncached(startDir, platformMap)` | Returns a map of available React Native platforms. The result is NOT cached.                    |
```
acoates-ms, line 33: I'm trying to remember what the usage might have been, but I suspect there might be a case where a script is modifying the config or something, and wanted to be able to requery the list?  I dont know off hand of such a use, just pinging it incase it rings any bells for you.

acoates-ms (approved):

packages/tools-react-native/README.md
tido64: I couldn't find instances of it anywhere so I kinda assumed it was only used in tests and therefore is safe to remove. We can always bring it back it something breaks?

PR #3264, opened 2024-08-07
fix(tools-node): avoid additional stat call when resolving symlinks

tido64:
### Description

Avoid an additional stat call when trying to resolve symlinks

### Test plan

Existing tests should pass.


tido64:
Never mind. I forgot how `fs.stat` differs from `fs.lstat`

PR #3305, opened 2024-08-23
feat(tools-apple): add primitives for build and run commands

tido64:
### Description

- Added primitives for building 'build' and 'run' commands.
- Added experimental commands for building and running iOS/macOS apps. These need more testing, preferably outside of rnx-kit, and as such, will not be publicly available. But if people are willing to keep patches, there are ways to access them.

### Test plan

n/a


packages/cli/src/run.ts
```diff
@@ -0,0 +1,31 @@
...
+      return Promise.resolve();
+  }
+}
+
+export const rnxRunCommand = {
+  name: "rnx-run",
+  description:
+    "Build and run your native app for testing in emulator/simulator or on device",
+  func: rnxRun,
+  options: rnxBuildCommand.options,
```
tido64, line 30: We probably also need `--simulator` and `--without-building`?

packages/cli/src/build/ios.ts
```diff
@@ -0,0 +1,30 @@
+import type { Config } from "@react-native-community/cli-types";
```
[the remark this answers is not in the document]
tido64: The config is something that will stay in rnc-cli for a while. But this is just a type import.

packages/tools-apple/package.json
```diff
@@ -49,7 +49,8 @@
     "update-readme": "rnx-kit-scripts update-api-readme"
   },
   "dependencies": {
-    "@rnx-kit/tools-shell": "^0.1.2"
+    "@rnx-kit/tools-shell": "^0.1.2",
```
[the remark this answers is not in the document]
tido64: No, I haven't checked whether Changesets handles them correctly yet and I'd rather not find out yet.

packages/tools-apple/src/ios.ts
```diff
@@ -1,11 +1,22 @@
 import { retry } from "@rnx-kit/tools-shell/async";
 import { ensure, makeCommand } from "@rnx-kit/tools-shell/command";
 import * as readline from "node:readline";
-import type { Device, DeviceType, Logger, Simulator } from "./types.js";
+import { open } from "./macos.js";
```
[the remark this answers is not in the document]
tido64: This just imports the `open` command for Simulator.app

packages/tools-apple/src/xcode.ts
```diff
@@ -1,9 +1,30 @@
...
+import { spawn } from "node:child_process";
+import * as path from "node:path";
+import { iosSpecificBuildFlags } from "./ios.js";
+import { macosSpecificBuildFlags } from "./macos.js";
+import { findSchemes } from "./scheme.js";
+import type { BuildParams, JSObject } from "./types.js";
 
 export const xcrun = makeCommand("xcrun");
 
+function remoteSpecificBuildFlags(
```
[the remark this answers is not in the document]
tido64: Cloud build stuff. Because of the Xcode version mismatch, sanitizers may crash on startup.

packages/cli/src/build.ts
```diff
@@ -0,0 +1,54 @@
...
+export function rnxBuild(
+  _argv: string[],
+  config: Config,
+  buildParams: InputParams
+) {
+  switch (buildParams.platform) {
+    case "ios":
+      return buildIOS(config, buildParams);
+    case "macos":
+      return buildMacOS(config, buildParams);
```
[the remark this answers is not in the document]
tido64: Windows? 😂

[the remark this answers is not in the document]
Saadnajmi: visionOS, specifically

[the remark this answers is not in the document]
tido64: I added it in the last iteration. I think I still want to get this published and test in RNTA though. It's harder to switch versions here.

packages/cli/src/build.ts
```diff
@@ -0,0 +1,89 @@
...
+) {
+  switch (buildParams.platform) {
+    case "ios":
+    case "visionos":
+      return buildIOS(config, buildParams);
+    case "macos":
+      return buildMacOS(config, buildParams);
+    default:
+      // @ts-expect-error Safe guard against user input
+      console.error(`Unsupported platform: ${buildParams.platform}`);
```
Saadnajmi, line 49: The nicest use of ts-expect-error I've ever seen :D

packages/cli/src/build.ts
```diff
@@ -0,0 +1,89 @@
...
+  func: rnxBuild,
+  options: [
+    {
+      name: "--platform <string>",
+      description: "Target platform",
+    },
+    {
+      name: "--workspace <string>",
+      description:
+        "Path, relative to project root, of the Xcode workspace to build (macOS only)",
```
Saadnajmi, line 68: Not iOS or visionOS?

packages/cli/src/build.ts
tido64: Hehehe... `tsc` inferred that `platform` is type `never` here but this is user input. I probably should tighten the types instead.

packages/cli/src/run.ts
```diff
@@ -0,0 +1,38 @@
...
+  name: "rnx-run",
+  description:
+    "Build and run your native app for testing in emulator/simulator or on device",
+  func: rnxRun,
+  options: [
+    ...rnxBuildCommand.options,
+    {
+      name: "--device <string>",
+      description: "The name of the device to launch the app in",
+    },
```
Saadnajmi, line 36: Why is this extra?

packages/tools-apple/src/xcode.ts
```diff
@@ -1,9 +1,84 @@
...
+
+  const buildSettingsArgs = ["-workspace", xcworkspace];
+  for (const flag of reusedFlags) {
+    const i = params.lastIndexOf(flag);
+    if (i >= 0) {
+      buildSettingsArgs.push(params[i], params[i + 1]);
+    }
+  }
+
+  buildSettingsArgs.push("-showBuildSettings", "-json");
```
Saadnajmi, line 46: ... you can show build settings as a JSON? I want update every pipeline I've ever touched now...

Saadnajmi (approved):

packages/cli/src/build.ts
tido64: `react-native config` doesn't include a `macos` section. But it does include `ios` and `visionos`. If we can fix that, we no longer need this flag.

packages/cli/src/run.ts
tido64: When we build, we don't need to target specific devices. Just generic device or simulator. But the run command actually needs to know which device/simulator you prefer.

packages/cli/src/build.ts
tido64: Letting Commander validate the input is better:
```
% yarn react-native rnx-run --platform windows
error: option '--platform <string>' argument 'windows' is invalid. Supported platforms: 'ios', 'macos', 'visionos'.
```

packages/tools-apple/src/xcode.ts
tido64: I think most Xcode tools accept a `-json` or `--json`.

PR #3306, opened 2024-08-23
Add typescript as a peerDependency in @rnx-kit/eslint-plugin

sverrejoh:
### Description

Strict package managers need all dependencies to be fulfilled. This carries the typescript dependency from the consuming package over to `@typescript-eslint/eslint-plugin`, which would otherwise fail on missing typescript dependency.

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -55,6 +55,11 @@
     "eslint": ">=8.56.0",
     "typescript": "^5.0.0"
   },
+  "peerDependenciesMeta": {
+    "typescript": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@microsoft/eslint-plugin-sdl": "^0.2.0",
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin/package.json
```diff
@@ -52,7 +52,8 @@
     "typescript-eslint": "^7.5.0"
   },
   "peerDependencies": {
-    "eslint": ">=8.56.0"
+    "eslint": ">=8.56.0",
+    "typescript": "^5.0.0"
```
tido64, line 56: Can we make it optional? Not all of our users are using TypeScript:
```suggestion
    "typescript": "^5.0.0"
  },
  "peerDependenciesMeta": {
    "typescript": {
      "optional": true
    }
```

sverrejoh: Yes, that's a good idea. It's optional in the @typescript-eslint package already.

PR #3326, opened 2024-09-02
feat(cli): enable `build` and `run` commands

tido64:
### Description

Make `build` and `run` commands available for public use

### Test plan

CI should pass.


Saadnajmi (approved):

.github/workflows/pr.yml
```diff
@@ -104,6 +104,10 @@ jobs:
           if [[ "$(yarn show-affected --base origin/${{ github.base_ref }})" = *"@rnx-kit/test-app"* ]]; then
             echo 'android=true' >> $GITHUB_OUTPUT
           fi
+      - name: Build @rnx-kit/cli
+        if: ${{ steps.affected-projects.outputs.android != '' }}
```
Saadnajmi, line 108: Why Android specifically?

tido64: This is the "build Android" job. If we don't need to build the app, we don't need to build the CLI.

PR #3355, opened 2024-09-14
fix(jest-preset): add `babel-plugin-syntax-hermes-parser` if needed

tido64:
### Description

Add `babel-plugin-syntax-hermes-parser` if we detect that React Native needs it

### Test plan

n/a


tido64:
No longer needed. `jest-preset` should already be picking up `@react-native/babel-preset`, which contains `babel-plugin-syntax-hermes-parser`.

PR #3392, opened 2024-10-16
fix(metro-plugin-typescript): fix beta versions of TypeScript failing version check

tido64:
### Description

Users are unable to use beta versions of TypeScript because `semver.satisfies("5.7.0-beta", ">=4.7.0")` returns `false`.

### Test plan

Tests were added.


tido64:
cc @sverrejoh

sverrejoh (approved):

PR #3424, opened 2024-11-20
fix(tools-react-native): add `loadContextAsync` to exported types

tido64:
### Description

Add missing `loadContextAsync` to exported types

### Test plan

n/a

### DIFF
--- a/packages/tools-react-native/context.d.ts
+++ b/packages/tools-react-native/context.d.ts
@@ -1,1 +1,5 @@
-export { loadContext, resolveCommunityCLI } from "./lib/context";
+export {
+  loadContext,
+  loadContextAsync,
+  resolveCommunityCLI,
+} from "./lib/context";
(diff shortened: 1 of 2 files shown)


Saadnajmi (approved):

packages/tools-react-native/context.d.ts
```diff
@@ -1 +1,5 @@
-export { loadContext, resolveCommunityCLI } from "./lib/context";
+export {
+  loadContext,
+  loadContextAsync,
+  resolveCommunityCLI,
+} from "./lib/context";
```
Saadnajmi, line 5: For my own knowledge: I always get a bit irked when I see imports/exports from `lib` since it's generated and not source, but I think that comes from inexperience in the JS world. Is there a reason we're importing from `lib` and not `src`? Or

tido64: Mostly because we don't ship src. But that will probably change in the near future when we fully support export maps and Node enables type stripping by default.

PR #3425, opened 2024-11-20
feat(config): add support for `extends` field

tido64:
### Description

Add support for `extends` field

### Test plan

TODO: Add tests

### DIFF
--- a/packages/config/src/getKitConfig.ts
+++ b/packages/config/src/getKitConfig.ts
@@ -1,9 +1,13 @@
-import { readPackage } from "@rnx-kit/tools-node";
-import path from "path";
+import type { PackageManifest } from "@rnx-kit/tools-node/package";
+import { readPackage } from "@rnx-kit/tools-node/package";
+import merge from "lodash.merge";
+import * as fs from "node:fs";
+import * as path from "node:path";
 import type { KitConfig } from "./kitConfig";
 
 /**
- * Options for retrieving a kit config. The default is equivalent to passing { cwd: process.cwd() }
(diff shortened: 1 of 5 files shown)


packages/config/src/getKitConfig.ts
```diff
@@ -17,27 +21,43 @@ export type GetKitConfigOptions = {
   cwd?: string;
 };
 
+function findPackageDir({ module, cwd = process.cwd() }: GetKitConfigOptions) {
+  if (!module) {
+    return cwd;
+  }
+
+  const p = require.resolve(module + "/package.json", { paths: [cwd] });
```
NikoAri, line 29: (nit) can you expand, instead of "p" to something like "packageJsonFilePath"?

packages/config/src/getKitConfig.ts
```diff
@@ -17,27 +21,43 @@ export type GetKitConfigOptions = {
...
+  return path.dirname(p);
+}
+
+function loadBaseConfig(
+  config: PackageManifest["rnx-kit"],
+  packageDir: string
+): PackageManifest["rnx-kit"] {
+  const base = config?.extends;
+  if (typeof base !== "string") {
```
NikoAri, line 38: (nit) can't we typify this config instead of explicitly checking for types 😱

NikoAri (approved):

packages/config/src/getKitConfig.ts
tido64: The input is a JSON file. It's inherently unsafe.

acoates-ms (approved):

PR #3490, opened 2025-01-22
feat(tools-react-native): add additional platform helpers

JasonVMo:
### Description

This adds several helpers to the `@rnx-kit/tools-react-native` package for dealing with platform names and extensions. These are needed by the new typescript related package I'm working on and will also help centralize some things into a core package. In particular the `@rnx-kit/metro-plugin-typescript` package uses some of this functionality and that will be adjusted as a separate change.

Update: This also adds visionos to the list of valid platforms supported in @rnx-kit/tools-react-native.

### Test plan

I've been testing these changes as part of a larger change, but given this is purely additive the test plan for this PR is just to pass the standard tests.


packages/tools-react-native/src/platform.ts
```diff
@@ -8,6 +8,15 @@ import { readReactNativeConfig } from "./context";
...
 export type AllPlatforms = "android" | "ios" | "macos" | "win32" | "windows";
 
+// Possible values for AllPlatforms
+const allValues: AllPlatforms[] = [
+  "android",
+  "ios",
+  "macos",
+  "win32",
+  "windows",
+];
```
Saadnajmi, line 18: There's also visionOS technically. And tvOS

JasonVMo: Good point. I assume that is effectively:
- platform: "visionos"
- file naming: "filename.visionos.ts"
- platform package: "react-native-visionos"
Is that correct? Is there typically casing in it like "visionOs" or anything? I can add that as a separate PR.

packages/tools-react-native/src/platform.ts
```diff
@@ -134,22 +162,57 @@ export function platformExtensions(platform: string): string[] {
...
+}
+
+// Map of platform to npm package name
+const platformToPackage: Record<AllPlatforms, string> = {
+  android: "react-native",
+  ios: "react-native",
+  macos: "react-native-macos",
+  win32: "@office-iss/react-native-win32",
+  windows: "react-native-windows",
+};
```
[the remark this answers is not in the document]
JasonVMo: Yep, good callout, it is already using @rnx-kit/tools-react-native so I will update it once I add the visionOS changes.

packages/tools-react-native/src/platform.ts
JasonVMo: Looks like it is "@callstack/react-native-visionos". I went ahead and added vision OS to the platform list in this change, seems silly to not do it as a single update.

packages/tools-react-native/src/index.ts
```diff
@@ -13,7 +13,12 @@ export {
   expandPlatformExtensions,
   getAvailablePlatforms,
   getAvailablePlatformsUncached,
+  getModuleSuffixes,
+  getPlatformPackageName,
+  getPlatformsForPackage,
   parsePlatform,
+  parsePlatformValue,
   platformExtensions,
+  platformValues,
```
tido64, line 22: Do we need all these exports? We shouldn't expose a bigger surface than is necessary. This makes it harder to avoid breaking changes in the future.

packages/tools-react-native/src/platform.ts
```diff
@@ -134,22 +169,58 @@ export function platformExtensions(platform: string): string[] {
...
-    case "macos":
-    case "win32":
-    case "windows":
-      return val;
-
-    default:
-      throw new Error("Invalid platform '" + val + "'");
+  const platform = parsePlatformValue(val);
+  if (!platform) {
+    throw new Error(`Invalid platform '${val}'`);
```
tido64, line 190: We should change this to say unknown instead of invalid:
```suggestion
    throw new Error(`Unknown platform: ${val}`);
```

packages/tools-react-native/src/platform.ts
```diff
@@ -134,22 +169,58 @@ export function platformExtensions(platform: string): string[] {
...
+
+// Map of platform to npm package name
+const platformToPackage: Record<AllPlatforms, string> = {
+  android: "react-native",
+  ios: "react-native",
+  macos: "react-native-macos",
+  win32: "@office-iss/react-native-win32",
+  windows: "react-native-windows",
+  visionos: "@callstack/react-native-visionos",
+};
```
tido64, line 210: Not a fan of hard-coding packages like this. Can we use `getAvailablePlatforms()` instead?

tido64 (changes requested):

packages/tools-react-native/src/platform.ts
JasonVMo: In the forthcoming tools-typescript package this table is used to do a reverse lookup of what platforms to typecheck based on declared dependencies (when it is being asked to detect platforms). That being said I agree, the whole way platforms are declared feels overly hardcoded. It seems better to have the base platform setup declared in the root kit-config and have that drive overall usage.

Maybe something like:
- root config -> available platforms with this information provided. Similar to how the default platforms are set up in the RNTA root package.json. Or maybe android and ios are built-in with an ability to declare additional platforms.
- per platform -> ability to override platforms.

JasonVMo: I just removed these functions from the changes I just pushed and will figure out a different approach.

packages/tools-react-native/src/platform.ts
JasonVMo: Done

packages/tools-react-native/src/index.ts
JasonVMo: Yes, though I removed the getPlatformPackageName and getPlatformsForPackage calls given the below discussion.

packages/tools-react-native/src/platform.ts
```diff
@@ -134,22 +169,32 @@ export function platformExtensions(platform: string): string[] {
   }
 }
 
+/**
+ * @returns the given string as a platform value or undefined if it is not a valid platform.
+ */
+export function parsePlatformValue(val: string): AllPlatforms | undefined {
```
tido64, line 175: Nit: I feel like this should be called `tryParsePlatform` or something since it's basically a non-throwing version of `parsePlatform`.

tido64 (approved):

packages/tools-react-native/src/platform.ts
JasonVMo: Done

PR #3492, opened 2025-01-22
feat(config): add helper for loading kit config in the case where package.json has been parsed already

JasonVMo:
### Description

This is a quick helper function to load the kit config (resolving with the base config) in the case where package.json has already been loaded for a different purpose. This just routes to the internal call `loadBaseConfig` but makes it more generic by taking the `PackageManifest` type instead.

### Test plan

Purely additive code, routing to existing codepaths that are already in use. Just unit tests.

### DIFF
--- a/packages/config/src/getKitConfig.ts
+++ b/packages/config/src/getKitConfig.ts
@@ -76,7 +76,7 @@ export function getKitConfig(
  * @param packageDir The directory containing the package.json file.
  * @returns The rnx-kit configuration for this package, merged with any base configuration.
  */
-export function getKitConfigFromPackageJson(
+export function getKitConfigFromPackageManifest(
   packageJson: PackageManifest,
   packageDir: string
 ): KitConfig | undefined {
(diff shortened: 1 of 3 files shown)


packages/config/src/getKitConfig.ts
```diff
@@ -69,3 +69,16 @@ export function getKitConfig(
...
   }
 }
+
+/**
+ * Helper for loading the KitConfig for cases where the package.json has already been loaded for other reasons.
+ * @param packageJson The loaded and parsed package.json file for this package.
+ * @param packageDir The directory containing the package.json file.
+ * @returns The rnx-kit configuration for this package, merged with any base configuration.
+ */
+export function getKitConfigFromPackageJson(
```
tido64, line 79: Nit: Manifest > package.json
```suggestion
export function getKitConfigFromPackageManifest(
```

tido64 (approved):

packages/config/src/getKitConfig.ts
JasonVMo: Good idea, made the switch

PR #3503, opened 2025-02-01
Add an experimental tools-typescript package for better react-native builds with typescript

JasonVMo:
### Description

A package that helps with building typescript, particularly for react-native. It
leverages the @rnx-kit/typescript-service package to build using the language
service rather than using the compiler directly. This allows for custom
resolution strategies (among other things). The compilation and type-checking
are still done by typescript but this drives some convenient custom
configurations.

In particular the `buildTypescript` command can do things like:

- multiplex a build in a directory, to build for multiple react-native platforms
  at the same time. The files will be split such that the minimal build can be
  performed.
- detect which react-native platforms should be built based on rnx-kit bundle
  configs or react-native.config.js settings.
- successfully build with the module suffixes without throwing up a ton of bogus
  unresolved reference errors.
- share caches where possible to speed up compilations within the same node
  process.

### Motivation

The current story for building typescript for react-native is sub-par, and
typescript itself is particularly restrictive with module resolution. This
addresses the ability to build react-native better right now, but also creates a
framework for experimenting with different resolvers.

### Test plan

- There are some unit tests already but more will be added in the future.
- Tested via repo linking in the FURN repo since it has complex build needs
- Planning to add a test repo for another change and will run builds in there and compare results with tsc as part of validation.


incubator/tools-typescript/src/reporter.ts
```diff
@@ -0,0 +1,101 @@
...
+    logging ? console.log(`${name}:`, ...args) : () => {};
+  };
+  const error = (...args: unknown[]) => {
+    errors++;
+    console.error(...args);
+  };
+  const trace = (...args: unknown[]) => {
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    tracing ? console.log(`${name}:`, ...args) : () => {};
+  };
```
tido64, line 79: You can return `undefined` to suppress the empty function warning. Also, since `logging` is constant, let's move it outside the function:
```suggestion
  const trace = tracing
    ? (...args: unknown[]) => { console.log(`${name}:`, ...args); }
    : () => undefined;
```

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
+import fs from "node:fs";
+import path from "node:path";
+import ts from "typescript";
+import { parseSourceFileDetails } from "./platforms";
+
+/**
+ * Map of file resolution lookups, this is global as it remembers module resolutions for
+ * a given source file
+ */
```
tido64, line 9: Will this be used in watch mode? What happens if files are added/deleted?

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
...
+
+/**
+ * @param root the root of the project (package root path)
+ * @returns a typescript module resolution cache
+ */
+export function getTsModuleCache(root: string) {
+  const rootCache = getRootCache(root);
+  rootCache.moduleCache ??= ts.createModuleResolutionCache(
+    root,
+    (x) => x,
```
tido64, line 61: Can we define an identity function above and reference it here instead of creating a function per root?

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * @param root the root of the project (package root path)
+ * @returns a typescript type reference directive resolution cache
+ */
+export function getTsTypeRefCache(root: string) {
+  const rootCache = getRootCache(root);
+  return (rootCache.typeRefCache ??=
+    ts.createTypeReferenceDirectiveResolutionCache(
+      root,
+      (x) => x,
```
tido64, line 79: Can we define an identity function above and reference it here instead of creating a function per root?

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
...
+        suffix: suffixes ? suffixes[i] : "",
+        resolved: lookupModuleForFile(module, file),
+      };
+    });
+
+    // find the first unchecked or found value, if it is found return it
+    const first = entries.find((e) => !e.resolved || e.resolved.resolvedModule);
+    if (first && first.resolved?.resolvedModule) {
+      return first.resolved.resolvedModule;
+    }
```
tido64, line 154:
```suggestion
    const resolvedModule = first?.resolved?.resolvedModule;
    if (resolvedModule) {
      return resolvedModule;
    }
```

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    // ensure the base entry has been resolved, this will be used to do file lookups
+    let result = entries.pop()!;
```
tido64, line 157: Don't force nonnull. Prefer throwing an error.

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
...
+ * - for './folder/file.js' returns './folder/file.ios.js', './folder/file.native.js', './folder/file.js'
+ *
+ * @param fileRef the file reference to multiplex with suffixes, can be with or without extension
+ * @param suffixes suffixes to search
+ * @returns either just the original file ref or a search list in suffix order
+ */
+export function getFilesToSearch(
+  fileRef: string,
+  suffixes?: string[]
+): [string, ...string[]] {
```
tido64, line 206: Isn't this just `string[]`?
```suggestion
): string[] {
```

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
...
+  if (suffixes && suffixes.length > 0) {
+    // split off the extension if it exists, then add the suffixes to try
+    const { base, ext = "" } = parseSourceFileDetails(fileRef, true);
+    return suffixes.map((s) => (s !== "." ? `${base}${s}${ext}` : fileRef)) as [
+      string,
+      ...string[],
+    ];
```
tido64, line 213: Isn't this just `string[]`? If you fix the return type, we can drop the type casting:
```suggestion
    return suffixes.map((s) => (s !== "." ? `${base}${s}${ext}` : fileRef));
```

incubator/tools-typescript/src/resolver.ts
```diff
@@ -0,0 +1,254 @@
...
+  module: string,
+  remap?: Record<string, string>
+): string {
+  if (remap) {
+    const match = /^(@[^/]+\/)?([^/]+)(\/.*)?$/.exec(module);
+    if (match && match[2]) {
+      const pkg = match[1] ? match[1] + match[2] : match[2];
+      const trailing = match[3] || "";
+      return (remap[pkg] || pkg) + trailing;
+    }
```
tido64, line 233: Should we use `parseModuleRef` from `@rnx-kit/tools-node`?

incubator/tools-typescript/src/task.ts
```diff
@@ -0,0 +1,74 @@
...
+  platforms?: PlatformInfo[]
+): Promise<void>[] {
+  const { asyncWrites, target } = options;
+  const promises: Promise<void>[] = [];
+  if (asyncWrites) {
+    context.writer = createAsyncWriter(target!);
+  }
+
+  const tasks = multiplexForPlatforms(context, platforms);
+  tasks.forEach((taskContext) => promises.push(buildTask(taskContext)));
```
tido64, line 68: Avoid `.forEach`. In this case, we should use `.map` instead:
```suggestion
  const promises = tasks.map((taskContext) => buildTask(taskContext));
```

incubator/tools-typescript/src/task.ts
```diff
@@ -0,0 +1,74 @@
...
+    );
+
+    // emit files that need to be built
+    if (build.length > 0) {
+      reporter.time(`emit: ${build.length} files`, () => {
+        build.forEach((file) => {
+          if (!project.emitFile(file)) {
+            reporter.error(`unable to build ${file}`);
+          }
+        });
```
tido64, line 29:
```suggestion
        for (const file of build) {
          if (!project.emitFile(file)) {
            reporter.error(`unable to build ${file}`);
          }
        }
```

incubator/tools-typescript/src/task.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    // check files that need to be type-checked
+    if (check.length > 0) {
+      reporter.time(`validate: ${check.length} files`, () => {
+        check.forEach((file) => {
+          if (!project.validateFile(file)) {
+            reporter.error(`type errors in ${file}`);
+          }
+        });
```
tido64, line 40:
```suggestion
        for (const file of check) {
          if (!project.validateFile(file)) {
            reporter.error(`type errors in ${file}`);
          }
        }
```

tido64 (approved):

tido64 (changes requested):
Sorry, didn't mean to approve this already.

incubator/tools-typescript/src/resolver.ts
JasonVMo: I certainly could, it would make this two steps. `parsePackageRef` pulls apart a reference and will return:
- `@rnx-kit/tools-react-native` -> `{ scope: '@rnx-kit', name: 'tools-react-native' }`
- `react-native/sub-path` -> `{ scope: '', name: 'react-native/sub-path' }`
- `@foo/bar/sub-baz` -> `{ scope: 'foo', name: 'bar/sub-baz' }`
What this function is doing is effectively splitting the sub-paths off so the module name can be compared for remapping. So...
- `@rnx-kit/tools-react-native` -> `[ '@rnx-kit/tools-react-native', '' ]`
- `react-native/sub-path` -> `[ 'react-native', 'sub-path' ]` 
- `@foo/bar/sub-baz` -> `[ '@foo/bar', 'sub-baz' ]`

I was trying to keep the changes isolated for now, but I think this function, or an equivalent should likely move to @rnx-kit/tools-node but I'd prefer to do that as a separate change. I'd likely create a common ancestor that picks apart the package ref but also splits the sub-path portion as well. Then make both parsePackageRef and the new function depend on that one.

incubator/tools-typescript/src/task.ts
JasonVMo: Good catch, had a few incarnations :)

JasonVMo:
> Sorry, didn't mean to approve this already.

No worries, wouldn't submit without hearing back anyways. I just submitted changes in response to your comments. Two still warrant a bit of discussion.

incubator/tools-typescript/src/resolver.ts
JasonVMo: Do you want me to go ahead and make that change to tools-node? I would probably make the new base function:
- split the string on '/'
- treat the first term as a scope if it starts with '@'
- treat the next term as the package name
- join the trailing entries on '/' as the sub-path (is there a more official name for this? Sub-module?)
Thinking about it that is likely easier than using an elaborate regex to split the string.

JasonVMo: Actually looks like `parseModuleRef` does what I want, opened a separate PR to make that one a little more efficient but for this change I'll switch over.

incubator/tools-typescript/src/files.ts
```diff
@@ -0,0 +1,169 @@
...
+      this.active++;
+      next().then(() => this.runFinished());
+    }
+  }
+
+  /** shift the queue to start at 0 again, just hygiene so it doesn't grow unbounded */
+  private rebalanceQueue() {
+    if (this.first > this.rebalanceAt) {
+      this.pending = this.pending.slice(this.first);
+      this.last -= this.first;
```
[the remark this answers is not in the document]
JasonVMo: Yep, made some adjustments here, legacy of a previous implementation.

incubator/tools-typescript/src/files.ts
```diff
@@ -0,0 +1,169 @@
...
+      this.dirs.add(dir);
+    }
+    const current = this.next++;
+    const fn = () =>
+      fs.promises.writeFile(filePath, content).catch((e) => {
+        this.reporter?.error(`Error writing ${filePath}`, e);
+        this.errors++;
+      });
+    this.active[current] = this.throttler.run(fn).then(() => {
+      delete this.active[current];
```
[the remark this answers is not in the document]
JasonVMo: I originally had it written using the original callback based fs.writeFile rather than the newer fs.promises.writeFile version but the code was actually much more messy. In particular the waiting for finish ended up having to use setTimeout which was much less elegant.

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,268 @@
...
+function createPlatformInfo(platform: string, npmName?: string): PlatformInfo {
+  // a list of built-in platforms to use as a fallback
+  const fallbackPackages: Record<string, string> = {
+    android: "react-native",
+    ios: "react-native",
+    macos: "react-native-macos",
+    win32: "@office-iss/react-native-win32",
+    windows: "react-native-windows",
+    visionos: "@callstack/react-native-visionos",
+  };
```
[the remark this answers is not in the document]
JasonVMo: Not too many times but still a good idea.

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,268 @@
...
+  return {
+    name: platform,
+    pkgName: npmName || fallbackPackages[platform] || "",
```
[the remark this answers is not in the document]
JasonVMo: Probably a good idea in the case where it is a platform unknown to us. Really only necessary if we are trying to remap the platform.

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,268 @@
...
+export function parseSourceFileDetails(
+  file: string,
+  ignoreSuffix?: boolean
+): ParsedFileName {
+  // valid last extensions, will manually check for .d.ts and .d.tsx
+  const validExtensions = [".js", ".jsx", ".ts", ".tsx", ".json"];
+  const match = /^(.*?)(\.[^./\\]+)?(\.[^./\\]+)?(\.[^./\\]+)?$/
+    .exec(file)
+    ?.slice(1)
+    .filter((s) => s);
```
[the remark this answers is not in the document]
JasonVMo: That isn't quite right but close. I reworked this a bit and used an approach similar to yours that is quite a bit easier to understand.

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,268 @@
...
+ */
+export function multiplexForPlatforms(
+  context: BuildContext,
+  platforms?: PlatformInfo[]
+): BuildContext[] {
+  const { cmdLine, reporter, writer, root } = context;
+  const checkOnly = !!cmdLine.options.noEmit;
+  // no platforms then we have nothing to do
+  if (!platforms || platforms.length === 0) {
+    context.platform = platforms?.[0];
```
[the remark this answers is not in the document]
JasonVMo: Good catch. Originally I was handling the single platform case here but the issue with that is it's better to route through the map below so platform modifiers live in a single place.

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,268 @@
...
+    return [context];
+  }
+
+  // set up the tasks for each platform, build array is undefined in checkOnly mode
+  const tasks: BuildContext[] = platforms.map((platform) => {
+    return {
+      root,
+      cmdLine: copyParsedCmdLine(cmdLine),
+      platform,
+      build: checkOnly ? undefined : [],
```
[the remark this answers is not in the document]
JasonVMo: Not sure which you are referring to but if you are talking about the copyParsedCmdLine it essentially copies both the levels of the cmdLine options that we are likely to modify. It is equivalent to:
`return { ...cmdLine, options: { ...cmdLine.options } }`

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,268 @@
...
+    };
+  });
+
+  // parse the files into file entries with suffixes already split out
+  const entries = processFileList(cmdLine.fileNames);
+
+  // now iterate through the tasks in reverse order setting the default task as last
+  for (let i = tasks.length - 1; i >= 0; i--) {
+    addFilesToContext(entries, i === 0, tasks[i]);
+  }
```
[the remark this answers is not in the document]
JasonVMo: Because we are essentially picking off files, then building any leftover files with the first platform which we are treating as primary.

incubator/tools-typescript/src/reporter.ts
```diff
@@ -0,0 +1,101 @@
...
+ * @returns a new timer object that will not capture any timings but will still execute the calls
+ */
+export function getNullTimer(): Timer {
+  return {
+    time<T>(_s: string, fn: () => T): T {
+      return fn();
+    },
+    async timeAsync<T>(_s: string, fn: () => Promise<T>): Promise<T> {
+      return await fn();
+    },
```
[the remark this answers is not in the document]
JasonVMo: Good catch, true for the non-timing case.

incubator/tools-typescript/src/resolver.ts
JasonVMo: I haven't gone through watch mode yet. Going to do that when I start looking at routing the metro-plugin-typescript code through here because it is similar logic. Added some todo items in the README.md

incubator/tools-typescript/src/resolver.ts
JasonVMo: Can you explain what you mean in more detail?

incubator/tools-typescript/src/resolver.ts
JasonVMo: Same comment as above

incubator/tools-typescript/src/resolver.ts
JasonVMo: This is actually a failure of typescripts type-assertions. It is guaranteed to be non-null because the string array is that [string, ...string[]] type, it always returns at least one. But when that type gets mapped above it flattens the output and typescript doesn't know it is at least size one. Taking away the non-null assertion there has a cascade of effects throughout the function and requires about four unnecessary null-checks.

I could do a typecast after the map using the same trick, which is likely safer than the non-null assertion since it is happening when it is set without any room for code to get in the middle.

incubator/tools-typescript/src/resolver.ts
JasonVMo: See above, it forces the type-checker to assert that we are returning at least one value.

incubator/tools-typescript/src/resolver.ts
JasonVMo: Removed all of these since it still doesn't go through the map statement and it is just easier to add a single null check in the right place.

incubator/tools-typescript/src/resolver.ts
JasonVMo: I just changed all the [string, string[]] stuff as it wasn't transferring through the map statement.

incubator/tools-typescript/src/files.ts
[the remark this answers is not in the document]
tido64: What do you think of using events instead of keeping a map of promises?

incubator/tools-typescript/src/platforms.ts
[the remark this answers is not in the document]
tido64: This line in particular:
```js
      build: checkOnly ? undefined : [],
```
Do we need a new array per platform?

incubator/tools-typescript/src/resolver.ts
tido64: Example:
```ts
function identity<T>(x: T): T {
  return x;
}

export function getTsModuleCache(root: string) {
  const rootCache = getRootCache(root);
  rootCache.moduleCache ??= ts.createModuleResolutionCache(
    root,
    identity,
    ...
```

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,302 @@
...
+  ignoreSuffix?: boolean
+): ParsedFileReference {
+  let ext = path.extname(file);
+  let suffix = "";
+  let base = file;
+
+  if (extensions.has(ext)) {
+    if (ext === ".ts" || ext === ".tsx") {
+      const dtsExt = ".d" + ext;
+      if (extensions.has(dtsExt) && file.endsWith(dtsExt)) {
```
tido64, line 119: Can't we assume `extensions.has(dtsExt)` is always true here?
```suggestion
      if (file.endsWith(dtsExt)) {
```
And on that note, do we even need `.d.tsx` in the set?

incubator/tools-typescript/src/platforms.ts
```diff
@@ -0,0 +1,302 @@
...
+    suffix = path.extname(base);
+  } else {
+    suffix = ext;
+    ext = "";
+  }
+
+  if (suffix && !ignoreSuffix) {
+    base = base.slice(0, -suffix.length);
+  } else {
+    suffix = "";
```
tido64, line 133: Unnecessary. It was initialized with an empty string.
```suggestion
```

incubator/tools-typescript/src/files.ts
```diff
@@ -0,0 +1,172 @@
...
+  private addPending(p: () => Promise<void>) {
+    this.pending.push(p);
+  }
+
+  private async processNextPending() {
+    this.rebalanceQueue();
+    if (this.first < this.pending.length) {
+      const next = this.pending[this.first++];
+      this.active++;
+      return next(); //.then(() => this.runFinished());
```
tido64, line 61: Remove comment?

incubator/tools-typescript/src/platforms.ts
JasonVMo: I don't want to make that assumption because it eventually depends on the parsing set. In some cases it may be configuration driven. I am removing .d.tsx, I don't think they are a thing.

incubator/tools-typescript/src/files.ts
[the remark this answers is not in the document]
JasonVMo: I can try it but I'd rather split that into a separate change. I think my understanding of how to use async correctly is better now than when I first tried the emitter implementation. I'd prefer to not block this change on that though.

incubator/tools-typescript/src/platforms.ts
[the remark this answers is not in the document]
JasonVMo: Yes, that array will have elements pushed to it by the sub function `pushToBuild` in `addFilesToContext` so it isn't just a throwaway array.

incubator/tools-typescript/src/resolver.ts
JasonVMo: Thanks, made the switch.

incubator/tools-typescript/src/files.ts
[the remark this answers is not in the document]
JasonVMo: Actually went ahead and made this change. I also added more tests for this, ensured it was reusable, restartable, and so forth.

tido64 (approved):

PR #3507, opened 2025-02-05
test(test-repos): Add a test monorepo under /test-repos/yarn-pnpm

JasonVMo:
### Description

There are several tools built in rnx-kit that it would be beneficial to test in a separate repo. This creates a separate monorepo, set up with yarn in pnpm mode under /test-repos/yarn-pnpm. This will allow us to leverage tools built in rnx-kit and test them out in various configurations.

- In particular this is essential for being able to test the yarn plugin for external workspaces currently active as a draft PR.

Currently this test repo isn't included in CI loops, as it is pulling in rnx-kit packages via npm, but that will be added once it is in a nice state and is actually testing out some of the tools directly. Ideally it will reference the rnx-kit packages via the external workspaces plugin.

For the repo itself:
- it has an eslint-config package, two controls copied and adapted from FURN (button and text), then the test-app code split into three packages, test-app-shared, test-app-desktop, and test-app-mobile.
- scripts is set up in much the same fashion as for rnx-kit itself.
- yarn is in pnpm mode
- right now only the build/lint scripts are working. The bundle commands route to the rnx-cli correctly but I haven't worked through the issues.
- It uses lage for builds rather than nx since it let's us test the tool that is more common internally.


test-repos/yarn-pnpm/packages/test-app-mobile/package.json
```diff
@@ -0,0 +1,187 @@
...
+    "@rnx-kit/metro-resolver-symlinks": "^0.2.1",
+    "@rnx-kit/metro-serializer": "^2.0.0",
+    "@rnx-kit/metro-serializer-esbuild": "^0.2.1",
+    "@rnx-kit/polyfills": "^0.2.0",
+    "@rnx-kit/react-native-auth": "^0.3.0",
+    "@rnx-kit/react-native-host": "^0.5.3",
+    "@rnx-kit/react-native-test-app-msal": "^5.0.1",
+    "@rnx-kit/third-party-notices": "^2.0.0",
+    "@rnx-repo-yarn-pnpm/eslint-config": "workspace:*",
+    "@rnx-repo-yarn-pnpm/scripts": "workspace:*",
```
tido64, line 60: Can we use `portal:` instead of copying these two packages?

test-repos/yarn-pnpm/packages/Button/babel.config.js
tido64, line 1: The package folder should be all lower-case.

test-repos/yarn-pnpm/packages/test-app-mobile/package.json
JasonVMo: The scripts will be not quite the same but I can make that adjustment for the eslint-config package.

tido64 (approved):

PR #3518, opened 2025-02-13
fix(eslint-plugin): Fix (dev)dependency cycle in the repo between eslint-config and eslint-plugin

JasonVMo:
### Description

Quick fix to what is technically a cycle in the repo.
- `@rnx-kit/eslint-config` - leverages `@rnx-kit/eslint-plugin` to lint files in our repo, this is private and internal.
- `@rnx-kit/eslint-plugin` - has a devDependency on `@rnx-kit/eslint-config` which to use for linting files in the plugin package.

This breaks the cycle by using the same config that eslint-config uses but directly referencing the configuration code in the plugin itself.

### Test plan

Only affects repo builds and linting of files. Linting run passes.

### DIFF
--- a/.changeset/forty-doors-itch.md
+++ b/.changeset/forty-doors-itch.md
@@ -1,0 +1,6 @@
+---
+"@rnx-kit/eslint-plugin": patch
+---
+
+Tooling only change to use locally defined eslint.config.js to break dependency
+cycle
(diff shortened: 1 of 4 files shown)


.changeset/forty-doors-itch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rnx-kit/eslint-plugin": patch
+---
+
+Tooling only change to use locally defined eslint.config.js to break dependency
+cycle
```
tido64, line 6: If you run `yarn change --empty`, an empty change file will be generated:
```suggestion
---
---
```
We don't need to publish a new version because of this internal change.

tido64 (approved):

PR #3534, opened 2025-02-28
fix(ios): 3rd party components on new arch 0.77+

janicduplessis:
### Description

On RN 0.77+ we need to setup the 3rd party component provider, or all components will end up using the interop layer, even if they have a new arch implementation.

### Test plan

Tested in the test-app by adding react-native-safe-area-context and checking that it uses the new arch implementation.

This was tested on both RN 0.76 (before the fix is needed) and 0.78 (after the fix is needed).

### DIFF
--- a/packages/react-native-host/cocoa/RNXBridgelessHeaders.h
+++ b/packages/react-native-host/cocoa/RNXBridgelessHeaders.h
@@ -21,6 +21,12 @@
 #define USE_FEATURE_FLAGS
 #endif  // __has_include(<react/featureflags/ReactNativeFeatureFlags.h>)
 
+#if __has_include(<ReactCodegen/RCTThirdPartyComponentsProvider.h>)
+#define USE_CODEGEN_PROVIDER 1
+#import <ReactCodegen/RCTThirdPartyComponentsProvider.h>
+#import <React/RCTComponentViewFactory.h>
+#endif // __has_include(<ReactCodegen/RCTThirdPartyComponentsProvider.h>)
+
 #if __has_include(<react/runtime/JSEngineInstance.h>)
 using SharedJSRuntimeFactory = std::shared_ptr<facebook::react::JSEngineInstance>;
 #else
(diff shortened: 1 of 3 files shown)


packages/react-native-host/cocoa/RNXBridgelessHeaders.h
```diff
@@ -21,6 +21,12 @@
 #define USE_FEATURE_FLAGS
 #endif  // __has_include(<react/featureflags/ReactNativeFeatureFlags.h>)
 
+#if __has_include(<ReactCodegen/RCTThirdPartyComponentsProvider.h>)
+#define USE_CODEGEN_PROVIDER 1
+#import <ReactCodegen/RCTThirdPartyComponentsProvider.h>
```
Saadnajmi, line 26: In what case would this be false? I.e: is there some min React Native version where even on the old arch, this header will always be there, this pound define is always true, and we could remove this code?

janicduplessis: On new arch in 0.76 the header is not there, it was added in 0.77.

packages/react-native-host/cocoa/ReactNativeHost.mm
```diff
@@ -1,13 +1,12 @@
...
 
-#include "FollyConfig.h"
-
 #import <React/RCTBridge.h>
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTCxxBridgeDelegate.h>
 #import <React/RCTDevLoadingViewSetEnabled.h>
 #import <React/RCTUtils.h>
 
+#include "FollyConfig.h"
```
tido64, line 9: This header needs to be moved back. It must be applied first so we can disable coroutines for backwards-compatibility.

We should probably have wrapped the line in `// clang-format off` and `// clang-format on` to prevent this.

For context, see https://github.com/microsoft/rnx-kit/pull/2883.

janicduplessis:
@tido64 Fixed the header issue and added clang format comments

tido64 (approved):

PR #3543, opened 2025-03-05
fix(yarn-plugin-dynamic-extensions): load user extensions relative to the source config file

tido64:
### Description

~~Correct config type for package extensions module path. With `SettingsType.ABSOLUTE_PATH`, we no longer have to resolve the path ourselves.~~ Update: see https://github.com/microsoft/rnx-kit/pull/3543#discussion_r1980933059

~~Allow plugin to be disabled with an empty string or null values.~~ Update: see https://github.com/microsoft/rnx-kit/pull/3543#discussion_r1981572978

Load user extensions relative to the source config file.

### Test plan

n/a

### DIFF
--- a/incubator/yarn-plugin-dynamic-extensions/index.js
+++ b/incubator/yarn-plugin-dynamic-extensions/index.js
@@ -28,10 +28,11 @@ exports.factory = (require) => {
       return;
     }
 
+    const { npath } = require("@yarnpkg/fslib");
     const { pathToFileURL } = require("node:url");
 
     // On Windows, import paths must include the `file:` protocol.
-    const url = pathToFileURL(packageExtensions);
+    const url = pathToFileURL(npath.fromPortablePath(packageExtensions));
     const external = await import(url.toString());
     return external?.default ?? external;
   }
--- a/incubator/yarn-plugin-dynamic-extensions/index.js
+++ b/incubator/yarn-plugin-dynamic-extensions/index.js
@@ -23,7 +23,7 @@ exports.factory = (require) => {
     const packageExtensions = configuration.get(DYNAMIC_PACKAGE_EXTENSIONS_KEY);
     if (
       typeof packageExtensions !== "string" ||
-      packageExtensions === "false"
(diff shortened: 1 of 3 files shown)


incubator/yarn-plugin-dynamic-extensions/index.js
```diff
@@ -15,34 +15,45 @@ exports.name = "@rnx-kit/yarn-plugin-dynamic-extensions";
...
+   * @returns {value is string}
+   */
+  function isEnabled(value) {
+    if (typeof value !== "string") {
+      return false;
+    }
+
+    const path = require("node:path");
+    const input = path.basename(value);
+    return !["NULL", "Null", "false", "null", "~"].includes(input);
```
tido64, line 29: Actually, I don't know if this is any better now because we have to support inheritance and Yarn config doesn't support OR-ed types. Using `path.basename` to get the input is a hack (it won't let us input an empty string, for instance) so I will revert, but keep this function instead.

Saadnajmi (approved):

incubator/yarn-plugin-dynamic-extensions/index.js
```diff
@@ -15,17 +15,26 @@ exports.name = "@rnx-kit/yarn-plugin-dynamic-extensions";
...
+  /**
+   * @param {unknown} value
+   * @returns {value is string}
+   */
+  function isEnabled(value) {
+    if (typeof value !== "string") {
+      return false;
+    }
+
+    return !["", "NULL", "Null", "false", "null", "~"].includes(value);
```
Saadnajmi, line 27: Is there really, not a better way to do this?

JasonVMo: Why not add a separate boolean setting for disabling the plugin?

tido64: Alternative: Disable this plugin if it was inherited from parent configs.

JasonVMo (approved):

PR #3560, opened 2025-03-21
Changes to prepare for running typescript directly in node

JasonVMo:
### Description

These are preparatory changes for being able to run typescript code directly in via node.

#### `@rnx-kit/tsconfig`

This adds a new `tsconfig.node.json` file and exports it from the project. This is similar to the `tsconfig.esm.json` entry but adds two additional compiler options:
- `rewriteRelativeImportExtensions: true` - this setting requires using explicit .ts file extensions in import statements but when transpiling they will be turned into the equivalent .js extensions. The explicit .ts file extensions are required to run typescript in node.
- `erasableSyntaxOnly: true` - this setting blocks the use of things like `namespace` and `enum` as these settings require an additional transpile strep vs. plain stripping. While there is a flag to do extra work to transpile these in node 22.6+, the automatic running in node 23+ will not include this flag. Better to be strict here and be able to run more places.

#### `@rnx-kit/tools-typescript`

This does the conversion work to switch this package to be able to be runnable directly via node. Note that the supporting packages need to be converted too but this was done as a proof of concept to start.

#### Typescript version upgrade

I also upgraded the version of typescript internally (and the supporting tools) to a 5.8+ version as this is required to have the `erasableSyntaxOnly` flag available.


incubator/tools-typescript/package.json
```diff
@@ -10,10 +10,18 @@
   },
   "files": [
     "lib/**/*.d.ts",
-    "lib/**/*.js"
+    "lib/**/*.js",
+    "src"
   ],
-  "main": "lib/index.js",
-  "types": "lib/index.d.ts",
+  "type": "module",
```
tido64, line 16: Is this intentional? This is a breaking change if it is.

incubator/tools-typescript/src/build.ts
```diff
@@ -3,10 +3,10 @@ import type { AllPlatforms } from "@rnx-kit/tools-react-native";
...
 import path from "node:path";
 import ts from "typescript";
-import { loadPackagePlatformInfo } from "./platforms";
-import { createReporter } from "./reporter";
-import { createBuildTasks } from "./task";
-import type { BuildContext, BuildOptions, PlatformInfo } from "./types";
+import { loadPackagePlatformInfo } from "./platforms.ts";
+import { createReporter } from "./reporter.ts";
+import { createBuildTasks } from "./task.ts";
+import type { BuildContext, BuildOptions, PlatformInfo } from "./types.ts";
```
tido64, line 9: What does `tsc` output? I don't think it changes the file extensions for the CJS output.

packages/tsconfig/tsconfig.node.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
```
tido64, line 2: This should extend `tsconfig.esm.json` instead. That way you only need to add `erasableSyntaxOnly` and `rewriteRelativeImportExtensions`.

incubator/tools-typescript/src/build.ts
JasonVMo: The types do resolve correctly from .js files, tested by mucking about in the scripts sources with both importing to a .cjs file and the .js file (which is esm) in our scripts.

tido64: Ah, I missed that `rewriteRelativeImportExtensions` actually rewrites import statements. TypeScript was opposed to this when Zearing suggested it to them a while back.

incubator/tools-typescript/package.json
tido64: `type: module` only tells Node to assume that this package is ESM by default. We currently only publish CJS and the raw TS files. I think it's wrong to enable this now.

tido64 (approved):

PR #3561, opened 2025-03-27
chore: use root Node min version for new packages

tido64:
### Description

Use root Node min version for new packages:
- Bump min version to 18.12
- Convert the new-package script to TypeScript

### Test plan

Generate a new package:

```
yarn new-package test-package
```

Inspect the contents of the package.


JasonVMo (approved):
The changes look good though I don't see anything about requiring node 22 for getting it working. Where is that set?

tido64:
> The changes look good though I don't see anything about requiring node 22 for getting it working. Where is that set?

We cannot require Node 22 for anything yet. We still test 20 on CI, and 18 is still in maintenance. We've only bumped min version to 18.

PR #3575, opened 2025-04-10
ci: actually allow error reporting directly on PRs

tido64:
### Description

`GITHUB_TOKEN` was never set so we don't get any error reporting on PRs.

### Test plan

n/a

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://unpkg.com/@changesets/config@1.6.2/schema.json",
+  "$schema": "https://unpkg.com/@changesets/config@3.1.1/schema.json",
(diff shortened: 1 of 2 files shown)


scripts/dependencies.config.js
```diff
@@ -56,6 +56,8 @@ const getDependencyVersion = (() => {
 export default function ({ cwd, manifest }) {
   let extensions = undefined;
 
+
+
```
tido64, line 60: Added newlines to test that reporting works.

tido64: Looks like it works:
image

Saadnajmi (approved):

JasonVMo (approved):

PR #3579, opened 2025-04-11
chore(yarn-plugin-external-workspaces): Change where the bundle is stored in the repo

JasonVMo:
### Description

Previously the produced bundle for the external-workspaces yarn plugin was built in the dist folder and published, but not tracked via git. A copy of it was tracked by git in the yarn plugin cache. This meant that to update the bundle used in the repository it would either need to be manually copied over or updated via the yarn plugin import command (which has the side effect of wiping out all our comments in the yarn configuration file.

Instead this adds the produced bundle to the repo, removes it from the plugin cache (effectively moving the file), then just references the dist folder directly. This avoids the multistep process and feels like a more standard pattern to follow.


Saadnajmi (approved):

incubator/yarn-plugin-external-workspaces/.gitignore
```diff
@@ -0,0 +1 @@
+!dist/*.cjs
```
tido64, line 1: Nit: Missing newline at end of file:
```suggestion
!dist/*.cjs

```

tido64 (approved):

JasonVMo:
Didn't realize this hadn't been merged, addressed the whitespace then merged from main.

tido64 (approved):

PR #3580, opened 2025-04-11
chore(scripts): Switch scripts package from yargs to clipanion

JasonVMo:
### Description

This switches the scripts package to using clipanion as the cli tool. This will be necessary for later changes because of the ability to more easily aggregate commands into a CLI from elsewhere.


scripts/src/commands/lint.js
```diff
@@ -1,8 +1,34 @@
...
+      This command lints the current package using eslint.
+    `,
+    examples: [[`Lint the current package`, `$0 lint`]],
+  });
+
+  args = Option.Proxy();
+
+  async execute() {
+    const args = this.args.length > 0 ? this.args : ["--no-warn-ignored"];
+    await lint(undefined, args);
```
tido64, line 28: Is there a reason for why we don't just inline the function and remove it?

Saadnajmi:
Why not the built in node one? Is it too basic?

tido64:
> Why not the built in node one? Is it too basic?

The built-in one only parses args. If you want anything beyond that, like `--help`, you have to build it yourself.

tido64 (approved):

PR #3591, opened 2025-04-23
test: use Node's built-in type support

tido64:
### Description

Replace `tsx` with Node's built-in type support

### Test plan

CI should pass.


tido64:
Still too early to enable this.

PR #3599, opened 2025-04-29
ci: generate tokens for publishing

tido64:
### Description

Generate short-lived tokens for publishing purposes.

### Test plan

n/a


tido64:
It looks like a token was successfully generated:
image

acoates-ms (approved):

PR #3610, opened 2025-05-09
feat(tools-scripts): Add tools-scripts package for common cli helpers and routines

JasonVMo:
### Description

Adds a tools-scripts package for common routines and helpers for building CLIs. This will be used to share code amongst our CLI implementations and provide helpers for implementing repo scripts folders.


scripts/src/experimental/test-lint.js
```diff
@@ -0,0 +1,3 @@
+import { createLintCommand } from "../../../incubator/tools-scripts/lib/index.js";
+
+export const LintCommand = createLintCommand([["lint"]], {});
```
tido64, line 3: Are users expected to know what a Clipanion path is? Can we make this simpler?

scripts/src/experimental/test-lint.js
```diff
@@ -0,0 +1,3 @@
+import { createLintCommand } from "../../../incubator/tools-scripts/lib/index.js";
```
tido64, line 1: Don't `..` out of the current package.

scripts/src/experimental/test-lint.js
tido64, line 1: Why is this file prefixed `test-`?

scripts/src/index.js
```diff
@@ -29,4 +29,11 @@ cli.register(Builtins.DefinitionsCommand);
 cli.register(Builtins.HelpCommand);
 cli.register(Builtins.VersionCommand);
 
-cli.runExit(process.argv.slice(2), Cli.defaultContext);
+const argsBase = process.argv.slice(2);
+const args = argsBase.filter((arg) => arg !== "--experimental");
```
tido64, line 33: We should only remove the first `--experimental`

scripts/src/index.js
```diff
@@ -29,4 +29,11 @@ cli.register(Builtins.DefinitionsCommand);
...
+const contextMixin = { experimental: args.length < argsBase.length };
+const context = {
+  ...Cli.defaultContext,
+  ...contextMixin,
+};
```
tido64, line 38: The temporary extra object is unnecessary:
```suggestion
const context = {
  ...Cli.defaultContext,
  experimental: args.length < argsBase.length,
};
```

scripts/src/runExperimental.js
```diff
@@ -0,0 +1,14 @@
...
+ * @returns
+ */
+export async function dynamicImport(modulePath) {
+  // eslint-disable-next-line no-new-func
+  return await new Function(`return import("${modulePath}")`)();
+}
+
+export async function runExperimentalCommand() {
+  const { runExperimentalCli } = await dynamicImport("./experimental/cli.js");
+  return await runExperimentalCli();
```
tido64, line 13: Why are we wrapping the import in an eval?
```suggestion
  const { runExperimentalCli } = await import("./experimental/cli.js");
```

scripts/tsconfig.json
```diff
@@ -3,5 +3,6 @@
   "compilerOptions": {
     "noEmit": true
   },
-  "include": ["src"]
+  "include": ["src"],
+  "exclude": ["src/experimental"]
```
tido64, line 7: Why are we not type-checking the experimental folder?

incubator/tools-scripts/src/lint-command.ts
```diff
@@ -0,0 +1,123 @@
...
+     */
+    filesOrPatterns = Option.Rest();
+
+    /**
+     * @returns {Promise<1 | void>}
+     */
+    async execute() {
+      const options: LintOptions = {
+        ...baseConfig,
+        cwd: this.context.cwd ?? process.cwd(),
```
tido64, line 107: Is `cwd` not `nonnull`?
```suggestion
        cwd: this.context.cwd,
```

incubator/tools-scripts/src/cli.ts
```diff
@@ -0,0 +1,32 @@
...
+export class RnxCli extends Cli<ScriptContext> {
+  private scriptDefaults: Pick<ScriptContext, "cwd" | "root">;
+
+  /**
+   * @param root root directory of the project
+   * @param options standard clipanion cli options
+   */
+  constructor(root: string, options: Partial<CliOptions> = {}) {
+    super(options);
+    this.scriptDefaults = { root, cwd: process.cwd() };
```
tido64, line 21: Is `process.cwd()` always the package root? Can these commands be run outside a package manager?

incubator/tools-scripts/src/lint-command.ts
```diff
@@ -0,0 +1,123 @@
...
+  const baseConfig = { ...defaultLintConfig, ...commandDefaults };
+
+  // create an anonymous command class for this command
+  return class extends BaseCommand {
+    static override paths = paths;
+    static override usage = createUsage(paths, baseConfig);
+
+    /**
+     * Fix the found issues where possible
+     */
```
tido64, line 77: Nit: These comments are unnecessary because we also have descriptions.

incubator/tools-scripts/src/lint.ts
```diff
@@ -0,0 +1,157 @@
...
+ * Run eslint in the specified directory
+ * @param {string} cwd
+ * @param {LintOptions} options
+ * @returns {Promise<number | void>}
+ */
+export async function runLint(options: LintOptions): Promise<number | void> {
+  const { cwd, fix, warnIgnored, ignorePatterns = [] } = options;
+
+  // dynamically import eslint and the pretty formatter since these are optional peer dependencies
+  const { ESLint } = await import("eslint");
```
tido64, line 77: We should catch this and output an actionable error message e.g.:
```ts
function importESLint() {
  try {
    const { ESLint } = import("eslint");
    return ESLint;
  } catch (_) {
    process.exitCode = 1;
  }
}

...

  // dynamically import eslint and the pretty formatter since these are optional peer dependencies
  const ESLint = importESLint();
  if (!ESLint) {
    console.error("Could not find `eslint`; did you forget to add it as a dev dependency?");
    return;
  }
```

incubator/tools-scripts/src/lint.ts
```diff
@@ -0,0 +1,157 @@
...
+ */
+export async function loadConfig(
+  options: LintOptions
+): Promise<
+  [string | boolean | undefined, LinterConfig | LinterConfig[] | undefined]
+> {
+  const { fallbackConfig, cwd } = options;
+
+  // default to the local config in cwd if it exists
+  const eslintConfigPath = path.join(cwd, "eslint.config.js");
```
tido64, line 125: ESLint supports many more types today: https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file

incubator/tools-scripts/src/lint.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/**
+ * Use git's ls-files to find files in the specified directory matching the filePatterns
+ */
+export function findFiles(options: LintOptions) {
+  const { cwd, filePatterns = [], globalIgnores = [] } = options;
+
+  // use git to list the files in the directory
+  const { stdout } = spawnSync("git", ["ls-files", ...filePatterns], {
```
tido64, line 149: `git ls-files` can also exclude files with `:!:<pattern>`:
```suggestion
  const { stdout } = spawnSync("git", ["ls-files", ...filePatterns, globalIgnores.map((p) => `:!:${p}`)], {
```

scripts/src/experimental/cli.js
```diff
@@ -0,0 +1,19 @@
...
+function workspaceRoot() {
+  return fileURLToPath(new URL("../../../", import.meta.url));
+}
+
+export async function runExperimentalCli() {
+  const cli = new RnxCli(workspaceRoot(), {
+    binaryLabel: "rnx-kit-scripts experimental CLI",
+    binaryName: "rnx-kit-scripts",
+    binaryVersion: "0.0.0",
+    enableColors: true,
```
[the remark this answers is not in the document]
tido64: Actually, it looks like Clipanion will detect this value from the env if you don't set it. We should remove this line.

incubator/tools-scripts/src/lint-command.ts
```diff
@@ -0,0 +1,123 @@
...
+     * Treat the trailing arguments as file patterns instead of files
+     */
+    patterns = Option.Boolean("--patterns", false, {
+      description: `Treat what comes after as file patterns to use for file matching`,
+    });
+
+    /**
+     * either a list of files to lint, or a list of patterns to use for file matching
+     */
+    filesOrPatterns = Option.Rest();
```
tido64, line 99: Another thing I've just recalled is that this pattern basically blackboxes the command i.e., you can no longer pass flags directly to ESLint. Sometimes, this is desirable because we cannot keep adding an entry for every new flag that gets added and it prevents people from learning a generic tool like ESLint. In 1JS, we started out like this and it was a pretty bad experience until we opened up the commands. We should not make the same mistake here.

PR #3645, opened 2025-06-12
feat(lint-lockfile): standalone lockfile linter

tido64:
### Description

`@rnx-kit/lint-lockfile` is a standalone tool for scanning your lockfiles for potential issues.

### Test plan

This replaces the one-off script, `scripts/lint-lockfile.js`. CI should pass.


.changeset/clean-poems-impress.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rnx-kit/lockfile-lint": minor
```
JasonVMo: I guess one aside is that it feels like @rnx-kit/lint-lockfile would make more sense. Particularly as we might eventually also have something like @rnx-kit/lint-package or other things of that nature.

JasonVMo: Actually looking at this it might also be called @rnx-kit/lint-yarn-lockfile as I don't know that unifying the API between lockfiles is worth the complexity.

packages/metro-serializer-esbuild/metro.config.js
```diff
@@ -37,8 +37,8 @@ const extraNodeModules = (() => {
 })();
 
 module.exports = makeMetroConfig({
-  reporter: { update: () => undefined },
-  resetCache: true,
+  cacheStores: [], // Avoids issues with deleting the cache on Windows
```
JasonVMo: What does this change do?

incubator/lockfile-lint/test/yarn.test.ts
```diff
@@ -0,0 +1,125 @@
...
+function mockLockfile(result: string): NodeFileSystem {
+  return { readFileSync: () => result } as unknown as NodeFileSystem;
+}
+
+describe("yarn:loadLockfile()", () => {
+  it("parses classic 'yarn.lock'", () => {
+    const lockfile = loadLockfile(
+      "yarn.lock",
+      mockLockfile(`
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
JasonVMo: Consider moving these to fixtures, and maybe using snapshot testing rather than having the deep equality hard-coded. Obviously not called yarn.lock as that would throw off the dependency scanners.

incubator/lockfile-lint/src/yarn/lockfile.ts
```diff
@@ -0,0 +1,14 @@
+import * as nodefs from "node:fs";
+import type { Lockfile } from "../types.ts";
+import { parseLockfileV1 } from "./classic.ts";
+import { parseLockfileV2 } from "./modern.ts";
+
+export function loadLockfile(lockfilePath: string, fs = nodefs): Lockfile {
+  const data = fs.readFileSync(lockfilePath, { encoding: "utf-8" });
+  try {
+    return parseLockfileV2(data);
```
JasonVMo: Are you sure that the yarn parser will always throw an exception on the v1 lockfiles? You could add a version check for the opening header to throw if you encounter that.

incubator/lockfile-lint/README.md
```diff
@@ -0,0 +1,30 @@
...
+[![Build](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml/badge.svg)](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml)
+[![npm version](https://img.shields.io/npm/v/@rnx-kit/lockfile-lint)](https://www.npmjs.com/package/@rnx-kit/lockfile-lint)
+
+🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧
+
+### THIS TOOL IS EXPERIMENTAL — USE WITH CAUTION
+
+🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧
+
+`@rnx-kit/lockfile-lint` is a standalone tool for scanning your lockfiles for
```
JasonVMo: Some more documentation of how to set up and enable rules would be helpful.

.changeset/clean-poems-impress.md
tido64: I named it `lockfile-lint` because of ESLint. But I don't mind either way.

packages/metro-serializer-esbuild/metro.config.js
tido64: You can ignore this. It's to make tests less flaky on Windows. It has already been merged on `main`.

incubator/lockfile-lint/src/yarn/lockfile.ts
tido64: v1 lockfile isn't YAML and will always throw.

incubator/lockfile-lint/README.md
tido64: There isn't much to set up at the moment. This is currently not configurable. There's only one rule. You either need it or you don't run this tool at all.

In the next PR, I will introduce a second rule and some rudimentary configuration: https://github.com/microsoft/rnx-kit/compare/tido/lockfile-lint...tido/lockfile/no-dupes-rule?expand=1

incubator/lockfile-lint/test/yarn.test.ts
tido64: Snapshot testing was only [introduced in Node 22.3.0](https://nodejs.org/en/learn/test-runner/using-test-runner#snapshot-tests). I'll revisit these when it is the min supported version.

JasonVMo (approved):
Approving this, one thing to consider for future iterations is to rename useWorkspacePackage to something else, since use[Function] is typically for react hooks.

tido64:
> Approving this, one thing to consider for future iterations is to rename useWorkspacePackage to something else, since use[Function] is typically for react hooks.

https://github.com/microsoft/rnx-kit/pull/3651

PR #3654, opened 2025-06-20
feat(reporter): Add new reporter for common logging, timing, and eventing

JasonVMo:
# Create @rnx-kit/reporter

This is a common package for logging output to the console and/or logfiles,
timing tasks and operations, and listening for events for task start, finish,
and errors.

It also contains a performance tracker that can be enabled to dump performance
information from execution of tasks. Enabling performance tracing will chain to
child processes as well to handle script tracking.

## Motivation

Standardizing the reporter used in our packages allows for easier high level
perf analysis of how our tools are behaving. By adding the various events it
also gives a common framework for people to add listeners for telemetry if
desired.

## Usage

Reporters have two roles, the base Reporter role, and a Task role. Reporters are
effectively at the root level, whereas Tasks are parented to a reporter or
another Task.

Reporters are created by calling
`createReporter<T>(options: ReporterOptions<T>)` with any options specified.

- The generic parameter sets the type of the `data` property.
- This allows passing additional information through reporters and tasks and
  will be surfaced in events.

The reporter interface is comprised of several parts:

### Logging Functions

These include `error`, `warn`, `log`, and `verbose`. These are structured like
the console logging functions in that they have variable parameters, which will
be serialized into a single message string by using node's `inspect`. This is
the same internal mechanism used by console.log, at least in the node
implementation.

- The `LogLevel` set in either the global settings, or overridden in the
  reporter settings dictates whether anything is output from the functions.
- `"log"` is the default level, which will enable the `error`, `warn`, and `log`
  functions. The `verbose` function will do nothing.
- File logging can be enabled by configuring `OutputOptions` when creating a
  reporter or by calling `updateDefaultOutput`.
- File logging will share the same log level as the console, unless the level is
  set specifically in the file settings.
- A prefix for a type of message can be set in settings. This is prepended to
  all messages of this type. For instance the default error prefix contains
  `"Error:"`
- A label can be set for the reporter, which will prepend all output for all log
  types.

An additional `throwError` function is provided which will log the error and
then throw with that message. It will send an event for the error, and log it
under the reporter/task.

### Timing Functions

The task functions wrap a function call, either async or synchronous, creating a
new sub reporter in the Task role which is passed as a parameter. The task data
type, output, and formatting options are inherited from the parent reporter.
Events will be sent when the task is started and when it completes, with errors
and sub-operation timing recorded within.

```ts
  task<T>(
    name: string | TaskOptions<TData>,
    fn: (reporter: Reporter<TData>) => T
  ): T;
  taskAsync<T>(
    name: string | TaskOptions<TData>,
    fn: (reporter: Reporter<TData>) => Promise<T>
  ): Promise<T>;
```

The `time` and `timeAsync` functions are helpers for high frequency operation
timing. The results of these operations will be aggregated within the given
reporter or task, and will record the total elapsed time and number of calls.
These will be available in the complete event.

```ts
  time<T>(label: string, fn: () => T): T;
  timeAsync<T>(label: string, fn: () => Promise<T>): Promise<T>;
```

### Formatting Functions

These functions are part of the `ReporterFormatting` interface and provide
helpers which will format or color text using the settings for the given
reporter.


incubator/reporter/src/formatting.ts
```diff
@@ -0,0 +1,272 @@
...
+    highlight2: 38, // cyan-blue
+    highlight3: 45, // cyan-blue light
+    label: 36, // blue-green,
+    errorPrefix: "red", // chalk red value
+    package: 208, // orange light
+    path: 43, // blue-green
+    scope: 166, // orange mid
+    warnPrefix: "yellowBright", // chalk yellow bright value
+    verboseText: "dim", // chalk dim setting
+  },
```
tido64, line 54: I can't help but think this looks over-engineered. It would be simpler if we used the chalk colors directly:
```ts
  colors: {
    duration: chalk.greenBright,
    durationUnits: chalk.dim,
    ...
  },
```
chalk takes care of disabling colors when [the terminal doesn't support it](https://github.com/chalk/chalk/blob/main/source/vendor/supports-color/index.js), supporting the various [ANSI modes, and many other things](https://github.com/chalk/chalk/blob/main/source/vendor/ansi-styles/index.js). We shouldn't have to handle them ourselves. Some of the colors you list, I couldn't find in chalk either so I'm curious where you got them.

It will also make customization easier as well: Rather than specifying hex numbers, consumers can use named values.

incubator/reporter/src/events.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @param name friendly name of the event, will be turned into a symbol for containment
+ * @returns a set of functions that can be used to listen for and send events
+ */
+export function createEventHandler<T>(name: string) {
+  const eventName = Symbol(name);
+  const eventChannel = channel(eventName);
+
+  return {
+    subscribe: (callback: (event: T) => void) => {
+      const handler = (event: unknown, name: symbol | string) => {
```
tido64: Should we use `ChannelListener` exported from `node:diagnostics_channel`?
```suggestion
      const handler: ChannelListener = (event, name) => {
```

incubator/reporter/src/output.ts
```diff
@@ -0,0 +1,180 @@
...
+export function getFileStream(
+  settings: OutputSettings["file"]
+): fs.WriteStream | undefined {
+  if (settings?.target) {
+    const { target, writeFlags } = settings;
+    if (typeof target === "string") {
+      // if the target is a string, create a write stream, updating the settings so we don't open it twice
+      // Resolve the log path relative to the current file's directory
+      const logPath = path.isAbsolute(target) ? target : path.resolve(target);
+      fs.mkdirSync(path.dirname(logPath), { recursive: true });
```
tido64: We need to add mode. By default, Node will give everyone read/write access.
```suggestion
      fs.mkdirSync(path.dirname(logPath), { recursive: true, mode: 0o755 });
```

incubator/reporter/src/output.ts
```diff
@@ -0,0 +1,180 @@
...
+ */
+export function getFileStream(
+  settings: OutputSettings["file"]
+): fs.WriteStream | undefined {
+  if (settings?.target) {
+    const { target, writeFlags } = settings;
+    if (typeof target === "string") {
+      // if the target is a string, create a write stream, updating the settings so we don't open it twice
+      // Resolve the log path relative to the current file's directory
+      const logPath = path.isAbsolute(target) ? target : path.resolve(target);
```
tido64: Can't we always use `path.resolve`?
```suggestion
      const logPath = path.resolve(target);
```

incubator/reporter/src/output.ts
```diff
@@ -0,0 +1,180 @@
...
+      fs.mkdirSync(path.dirname(logPath), { recursive: true });
+      settings.target = fs.createWriteStream(logPath, {
+        encoding: "utf8",
+        flags: writeFlags || "w",
+      });
+      return settings.target;
+    } else {
+      // if the target is already a write stream, return it
+      return target;
```
tido64, line 176: How do we know this is a write stream? Do we need to validate?

incubator/reporter/src/performance.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+    // in file only mode suppress the console output
+    if (mode === "file-only") {
+      settings.level = "error";
+    }
+    // now create the reporter instance
+    this.reporter = new ReporterImpl({
+      name: "PerformanceTracker",
+      label: "PERF:",
+      packageName: "@rnx-kit/reporter",
```
tido64: We should get the name from the `package.json`

incubator/reporter/src/performance.ts
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/**
+ * @internal
+ */
+export function checkPerformanceEnv() {
+  if (!checkedEnv) {
+    checkedEnv = true;
+    const env = process.env[PERF_TRACKING_ENV_KEY];
+    if (env && typeof env === "string") {
```
tido64: Isn't `env` always a string if it evaluates to true?

incubator/reporter/src/performance.ts
```diff
@@ -0,0 +1,195 @@
...
+ * @internal
+ */
+export function decodePerformanceOptions(
+  serialized?: string
+): [PerformanceTrackingMode, string | undefined] {
+  if (serialized) {
+    const parts = serialized.split(",");
+    const mode = validMode(parts[0]);
+    const file = parts.length > 1 ? parts[1] : undefined;
+    return [mode, file];
```
tido64: In JS, out of bounds access will always return `undefined`. We can simplify this:
```suggestion
    const [mode, file] = serialized.split(",");
    return [validMode(mode), file];
```

incubator/reporter/src/performance.ts
```diff
@@ -0,0 +1,195 @@
...
+): string {
+  return file ? `${mode},${file}` : mode;
+}
+
+function validMode(mode: string): PerformanceTrackingMode {
+  const enabledModes = new Set(["enabled", "verbose", "file-only"]);
+  if (enabledModes.has(mode)) {
+    return mode as PerformanceTrackingMode;
+  }
+  return "disabled";
```
tido64: Allocating a `Set` seems unnecessary. A static array or a switch should be sufficient. The latter option doesn't need type casting:
```suggestion
  switch (mode) {
    case "enabled":
    case "verbose":
    case "file-only":
      return mode;
    default:
      return "disabled";
  }
```

incubator/reporter/src/reporter.ts
```diff
@@ -0,0 +1,264 @@
...
+  implements Reporter<T>
+{
+  private static activeReporters: ReporterImpl[] = [];
+  private static processReporter: ReporterImpl | undefined = undefined;
+
+  static handleProcessExit() {
+    // send process exit event to all reporters and tasks
+    while (ReporterImpl.activeReporters.length > 0) {
+      ReporterImpl.activeReporters.pop()?.finish(undefined, "process-exit");
+    }
```
tido64: Not a fan of popping an array in a while loop:
```suggestion
    // send process exit event to all reporters and tasks
    const activeReporters = ReporterImpl.activeReporters;
    ReporterImpl.activeReporters = [];
    for (const reporter of activeReporters) {
      reporter.finish(undefined, "process-exit");
    }
```

incubator/reporter/src/reporter.ts
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  private buildMsgPrepFunctions(label?: string): Record<LogLevel, PrepareMsg> {
+    // color the label if requested
+    if (label) {
+      label = this.color(label, "label");
+    }
+    const serialize = this.serializeArgs;
+    const colorText = this.color;
+    const prefixes = {} as Record<LogLevel, PrepareMsg>;
```
tido64, line 245:
```suggestion
    const prefixes: Record<LogLevel, PrepareMsg> = {};
```

incubator/reporter/src/types.ts
```diff
@@ -0,0 +1,274 @@
...
+    colors?: boolean;
+  };
+};
+export type OutputOptions = Partial<OutputSettings>;
+
+export type ColorSettings = Record<ColorTypeValues, ColorValue | ColorValue[]>;
+
+export type FormattingSettings = {
+  // whether colors should be disabled for the console, will override inspect options if disabled
+  disableColors?: boolean;
```
tido64: There are known conventions for this. Node supports `FORCE_COLOR`. chalk supports `--color`/`--no-color` in addition to the env var. We should not deviate from this. If we're already depending on chalk, we should use it for what it's worth.

incubator/reporter/src/output.ts
tido64: Speaking of which, I'm adding filesystem utility functions here: https://github.com/microsoft/rnx-kit/pull/3657

incubator/reporter/src/formatting.ts
JasonVMo: I played with this a bit, also looked at the way yarn handles the chalk colors as well. Couple of points here:
- this supports setting a color to an ansi 256 number, a hex value like "#ccbbaa", or a named value like "yellowBright".
- the colors are rendered via chalk's color APIs providing appropriate downsampling.
- ANSI 256 is actually a very common terminal color mode, when hex values are downsampled to ANSI values it isn't a great mapping, and even if you use the direct hex values they don't always get mapped precisely.
- The ANSI codes are more compact in the strings.
- The built in chalk colors are a bit bright and garish.
- I picked ANSI colors that look distinct, can be seen on both black and white backgrounds, and aren't too bright and garish.

We can tweak the values in the future as we like, but using the ansi color values builds in a bit of consistency across displays and efficiency. We could provide a bunch of named constants mapping the numbers to more friendly names but again that can be something we can look at in the future.

incubator/reporter/src/output.ts
JasonVMo: The type is either string or write stream, and if it gets to this point we've already handled strings. Still likely good to put in an instance of check.

incubator/reporter/src/reporter.ts
JasonVMo: It leaves some room for something to come in sideways during shutdown but that's a pretty niche edge case and would be a race condition anyways.

incubator/reporter/src/types.ts
JasonVMo: The one behavior left out of that are the colors provided by Inspect. That being said a user can already set colors: false in inspect options if desired.

incubator/reporter/src/reporter.ts
tido64: JS is single threaded unless you start using worker threads (which is a whole world of pain in itself)

tido64 (approved):

PR #3657, opened 2025-06-24
feat(tools-filesystem): implement common functions

tido64:
### Description

Added `ensureDir`, `ensureDirForFile`, `writeJSONFile` and `writeTextFile`.

### Test plan

n/a


packages/cli/package.json
```diff
@@ -51,6 +51,7 @@
     "@rnx-kit/third-party-notices": "^2.0.0",
     "@rnx-kit/tools-android": "^0.2.0",
     "@rnx-kit/tools-apple": "^0.2.1",
+    "@rnx-kit/tools-filesystem": "^0.1.0",
```
Saadnajmi, line 54: Going off an earlier conversation, why don't we use `workspace:*` here?

tido64: Because of this: https://github.com/changesets/changesets/issues/432

packages/tools-filesystem/package.json
```diff
@@ -14,7 +13,6 @@
     "lib/**/*.js",
     "src"
   ],
-  "type": "module",
```
Saadnajmi, line 17: For my own knowledge, does the removal mean this is not an ESM only module?

packages/tools-filesystem/src/index.ts
```diff
@@ -0,0 +1,62 @@
+import * as nodefs from "node:fs";
+import * as path from "node:path";
+
+const DEFAULT_ENCODING = "utf-8";
+const DEFAULT_FILE_MODE = 0o644;
```
Saadnajmi: I know this is probably well known, but could we put a comment on what permissions this allows? I only know 755/777 off the top of my head

packages/tools-filesystem/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+const MKDIR_P_OPTIONS = { recursive: true, mode: 0o755 } as const;
+
+/**
+ * Ensures that a directory exists, creating it if necessary.
+ * @param path The path to the directory to ensure.
+ */
+export function ensureDir(
+  path: string, // Force a newline here so that `fs` can be marked as internal
+  /** @internal */ fs = nodefs
```
Saadnajmi, line 15: I guess the idea is you could pass in something like bun's fs implementation?

Saadnajmi (approved):

packages/tools-filesystem/package.json
tido64: Yes. CJS is assumed if type is unset.

packages/tools-filesystem/src/index.ts
tido64: The way this is calculated is basically:

1. For each of owner/group/other, there are three permissions: read, write and execute (or simply rwx)
2. You flip a bit for each permission, for example: Read/write permission is rw- or 110, which is 6

So 644 means read/write for owner, readonly for group/other. The execute bit is only needed for executables and directories.

packages/tools-filesystem/src/index.ts
tido64: This is for mocking purposes. See tests.

PR #3666, opened 2025-06-27
fix(align-deps): use react-navigation v7 in prests

SunskyXH:
### Description

Currently, `@react-navigation/stack` is limited to version 6.x in the dependency configuration. However, since React Native 0.72, version 7.x is fully supported and compatible.
This PR updates the version constraint to allow 7.x versions, enabling projects to use the latest features and improvements in React Navigation's stack navigator.

If directly updating to 7.x is not acceptable, would it be possible to use a range like `^6.0.0 || ^7.0.0` to support both major versions? This would provide flexibility for projects that need to stay on 6.x while allowing others to upgrade to 7.x.

### Test plan

``` shell
```


SunskyXH:

JasonVMo (commented):
Thanks for the contribution and sorry for the delay. I'm waiting on tido64 to get back from vacation to sign off since he is the primary maintainer of align-deps and I'll defer to him on how we wants to approach this. I'm a bit hesitant to do something that will force an upgrade for older RN profiles, potentially unexpectedly.

tido64:
Thanks for the contribution, @SunskyXH. Sorry for the late reply.

As @JasonVMo pointed out, we want to avoid making breaking changes as possible. I had a quick look and it does look like a few props have been removed and/or renamed. I think if we made this change from 0.76 and say that it's required for New Arch, it's less problematic. Would that work for you?

SunskyXH:
Thanks for your suggestion @tido64 , staring from 0.76 is much reasonable and works for me.
I have updated the code. Please take a look when you have time.

tido64:
Thanks for the update. Could you also update the README? `yarn update-readme` should do the trick.

SunskyXH:
> Thanks for the update. Could you also update the README? `yarn update-readme` should do the trick.

I have updated README, please take a look.

tido64 (approved):
Thanks again for your contribution 😊

PR #3685, opened 2025-07-23
fix(react-native-host): fix 3rd party Fabric components not rendering

tido64:
### Description

Fix third party Fabric components not rendering

### Test plan

See https://github.com/microsoft/react-native-macos/issues/2558

### DIFF
--- a/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
+++ b/packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
@@ -48,6 +48,13 @@
 #import <react/nativemodule/defaults/DefaultTurboModules.h>
 #endif
 
+#if __has_include(<ReactAppDependencyProvider/RCTAppDependencyProvider.h>)
+#import <ReactAppDependencyProvider/RCTAppDependencyProvider.h>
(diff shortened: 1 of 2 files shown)


packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
```diff
@@ -48,6 +48,13 @@
 #import <react/nativemodule/defaults/DefaultTurboModules.h>
 #endif
 
+#if __has_include(<ReactAppDependencyProvider/RCTAppDependencyProvider.h>)
+#import <ReactAppDependencyProvider/RCTAppDependencyProvider.h>
+#define USE_OSS_CODEGEN 1
```
Saadnajmi, line 53: I think this name means more in Meta’s internal build system. Do we need this name specifically?

Saadnajmi (approved):

packages/react-native-host/cocoa/RNXTurboModuleAdapter.mm
tido64: Not really, but it's easier to find references upstream if we need to.

PR #3711, opened 2025-07-30
fix(align-deps): use react-native-screens 4.4.0 for React Native 0.76

balazsgerlei:
### Description

`@react-navigation` has been updated recently to `7.x` (see https://github.com/microsoft/rnx-kit/pull/3666) from _React Native_ `0.76` but that doesn't work with `react-native-screens` `3.x` which it is currently limited to (the concrete definition is `^3.34.0` from `0.75`). Technically any version works from `4.0.0 `up to (but not including) `4.5.0` (if using Fabric) but I didn't want to break the convention of using `^` syntax (and I'm not even sure if something like `>=4.0.0 <4.5.0` would even work, but FYI that's the correct requirement AFAIK).

See README of `react-native-screens` where it defines compatibility: https://github.com/software-mansion/react-native-screens?tab=readme-ov-file#support-for-fabric

### Test plan

```
```


balazsgerlei:
We really need this change as we are on _React Native_ `0.76` and use `align-deps` as part of our CI checks and while it now requires `^7.1.14` for `@react-navigation`, it still forces react-native-screens to `^3.34.0` which doesn't work together, see [https://stackoverflow.com/a/79611866](https://stackoverflow.com/a/79611866)

packages/align-deps/src/presets/microsoft/react-native/profile-0.76.ts
```diff
@@ -143,6 +143,10 @@ export const profile: Profile = {
     name: "react-native-safe-area-context",
     version: "^4.12.0",
   },
+  screens: {
+    name: "react-native-screens",
+    version: "^4.4.0",
```
tido64: As you've said in the description, let's make this so:
```suggestion
    version: ">=4.0 <4.5",
```
Then add `^4.5.0` in `profile-0.77.ts` so that it works with 0.77+

tido64 (commented):
Sorry this wasn't properly tested. Thanks for the fix! I think we should go ahead with your suggestion and limit 0.76 to `>=4.0 <4.5`. We have done this before: https://github.com/microsoft/rnx-kit/blob/33a9ae1d1262eb8be79f7a657858e647f42bfbe7/packages/align-deps/src/presets/microsoft/react-native/profile-0.72.ts#L119-L122

We should also add an entry for 0.77+

packages/align-deps/src/presets/microsoft/react-native/profile-0.76.ts
balazsgerlei: Ah, I see, you are right, I'll change it! I think 0.77 already has the proper entry:
```
screens: {
    name: "react-native-screens",
    version: "^4.5.0",
  },
```

balazsgerlei: Okay, did the changes via rebase.

tido64 (approved):
Thanks again for fixing this ❤️

balazsgerlei:
Hmm... It seems like I messed up some formatting, can you help me with what should I do differently @tido64 ?

tido64:
> Hmm... It seems like I messed up some formatting, can you help me with what should I do differently @tido64 ?

Weird, it looks like some Windows newlines got in? I didn't think that should be happening. Can you run `yarn format` locally and see if that fixes it?

tido64:
Oh, did you manually edit the README or did you run a script to update it? `yarn update-readme` should format it correctly.

tido64 (approved):

balazsgerlei:
> Oh, did you manually edit the README or did you run a script to update it? `yarn update-readme` should format it correctly.

You were right, that was the key, I manually edited it. Thanks for helping with getting this merged!
Now we just have to wait for the release. 😃

tido64:
> You were right, that was the key, I manually edited it. Thanks for helping with getting this merged! Now we just have to wait for the release. 😃

FYI: https://github.com/microsoft/rnx-kit/releases/tag/%40rnx-kit%2Falign-deps%403.2.0

PR #3721, opened 2025-08-01
chore(align-deps): automate preset update

tido64:
### Description

Automatically generate a profile for new React Native release candidates when they become available.

### Test plan

n/a


.github/workflows/align-deps.yml
```diff
@@ -0,0 +1,53 @@
...
+          yarn install --mode update-lockfile
+      - name: Generate changeset
+        id: changeset
+        run: |
+          echo '---' > .changeset/rnx-align-deps.md
+          echo '"@rnx-kit/align-deps": patch' >> .changeset/rnx-align-deps.md
+          echo '---' >> .changeset/rnx-align-deps.md
+          echo '' >> .changeset/rnx-align-deps.md
+          head -n 1 packages/align-deps/update-profile.output.txt >> .changeset/rnx-align-deps.md
+          echo "message=$(head -n 1 packages/align-deps/update-profile.output.txt | tr '[:upper:]' '[:lower:]')" >> $GITHUB_OUTPUT
```
Saadnajmi, line 37: You can't do this with like a changesets CLI command?

Saadnajmi (approved):

.github/workflows/align-deps.yml
tido64: The CLI is interactive-only so I'm assuming it will just stall the pipeline.

PR #3728, opened 2025-08-08
Create AI hint files for Copilot/Claude in the repository and turn on nx mcp server

JasonVMo:
### Description

This creates the AI hint files for both Copilot and the Claude console for anyone using those tools. It also enables the nx mcp information.

### Test plan

Non-coding change, has no impact on packages in the repository.


.github/copilot-instructions.md
```diff
@@ -0,0 +1,271 @@
...
+## Project Overview
+
+`rnx-kit` is a collection of battle-tested React Native development tools
+created by Microsoft engineers for optimizing React Native development at scale.
+This is a **monorepo** using:
+
+- **Nx v20.3.3** (patched) for build orchestration and project management
+- **Yarn v4.6.0** as the package manager with workspaces
+- **TypeScript v5.0+** as the primary language with strict configuration
+- **Node.js ≥18.12** runtime requirement
```
Saadnajmi, line 12: Are the version numbers helpful? I fear they will get out of date. Perhaps we could say they're in the root `package.json`? Not sure that's how these hints work.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,271 @@
...
+yarn build              # Build all packages using Nx
+yarn test               # Build and test all packages
+yarn lint               # Lint all packages with ESLint
+yarn bundle             # Bundle packages for distribution
+yarn format             # Format code with Prettier
+yarn clean              # Clean build artifacts (git clean -dfqx)
+yarn new-package <name> # Create new package from template
+yarn change             # Create Changeset entry for versioning
+yarn rnx-align-deps     # Align dependencies across workspace
+yarn show-affected      # Show Nx affected projects
```
Saadnajmi, line 53: I may copy this standard set for other repos...

.github/copilot-instructions.md
```diff
@@ -0,0 +1,271 @@
...
+### Metro Integration
+
+- Follow Metro plugin architecture
+- Handle both development and production builds
+- Respect Metro's transformer and serializer APIs
+- Metro plugins follow established plugin architecture patterns
+
+### Node.js
+
+- Minimum Node 18.12 support
```
Saadnajmi, line 244: Note about version numbers again

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
```
Saadnajmi, line 3: If this is a copy, is there a way to generate these or enforce they are in sync?

Saadnajmi: Or is that what the addition to `nx.json` is doing?

JasonVMo: Nx does a runtime mcp server that integrates with vscode. In terms of the two files they are copies as each model type needs a dedicated file to keep the token count reasonable.

.github/copilot-instructions.md
JasonVMo: This was added by claude, then integrated by Copilot. I asked each tool to generate the files, then pull information from the other one. I believe the version numbers don't need to be up to date, though it helps. It mainly helps hint which version of the API to reference for usage. So the numbers need to be in the right ballpark range. It is about keeping the token count reasonable. It could pull them from package.json but that would increase the query complexity which then limits the amount of other context that can be provided.

.github/copilot-instructions.md
JasonVMo: See the above note about them. It hints the AI models to use the Node 18.x API without having to do additional lookups. You can easily ask the AI models to update new values when things change but they don't need to match perfectly.

CLAUDE.md
tido64: It looks like you can reference other files: https://docs.anthropic.com/en/docs/claude-code/memory#claude-md-imports

.github/copilot-instructions.md
tido64, line 1: Can we add the information that is useful for humans in `CONTRIBUTING.md` and just reference it in this file?

PR #3732, opened 2025-08-11
ci: use `setup-toolchain` action from `react-native-test-app`

tido64:
### Description

Use `setup-toolchain` action from `react-native-test-app`

### Test plan

CI should pass.


Saadnajmi:
Doesn't it make more sense for rnx-kit to be the source of truth?

tido64:
This action was created before rnx-kit was. It makes more sense for RNTA to keep it as it relates to building native and externals may already be using it.

PR #3736, opened 2025-08-14
fix(metro-config): fix duplicate packages slipping through

tido64:
### Description

Fix duplicate packages slipping through in a Yarn + pnpm setup. The previous method would prevent duplicates of the same version only e.g., both `react-19.0.0-xxxxxxxx` and `react-19.1.0-yyyyyyyy` could still be bundled together.

### Test plan

n/a

### DIFF
--- a/.changeset/shaggy-dragons-fry.md
+++ b/.changeset/shaggy-dragons-fry.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-config": patch
+---
+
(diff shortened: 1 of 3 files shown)


Saadnajmi (approved):
Do we need to somehow check what package manager we're using?

tido64:
> Do we need to somehow check what package manager we're using?

We are, by path: https://github.com/microsoft/rnx-kit/blob/0626a223b38bed8f247464fac3b7a70745415798/packages/metro-config/src/index.js#L108

PR #3742, opened 2025-08-15
feat(console): Add basic ansi color support to the @rnx-kit/console package

JasonVMo:
### Description

This enhances the console package by adding support for the basic set of ansi16 color values. These are named with the same convention that chalk uses (yellowBright rather than brightYellow). The color functions are contained under the colorize object so users can import that and call `colorize.red('my string')` to color text.

Existing functionality in the package should remain unchanged.

### DIFF
--- a/packages/console/src/index.ts
+++ b/packages/console/src/index.ts
@@ -7,21 +7,6 @@ export const supportsColor =
   process.env["NODE_ENV"] !== "test";
 
 /**
- * These are the base color constants for Ansi16 based colors suitable to be passed into the encodeAnsi16 and
- * encodeAnsi16Bright functions.
- */
-export const Ansi16 = {
-  black: 0,
-  red: 1,
-  green: 2,
-  yellow: 3,
-  blue: 4,
-  magenta: 5,
-  cyan: 6,
-  white: 7,
-};
-
-/**
  * Wraps a given string with ANSI escape codes for coloring. Will be a no-op if colors are disabled in process.stdout.
  * @param s The string to colorize.
  * @param start The starting color code, either a number or raw text.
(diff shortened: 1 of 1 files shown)


packages/console/src/index.ts
```diff
@@ -1,33 +1,105 @@
...
+  redBright: (s: string) => encodeAnsi16Bright(s, 1),
+  greenBright: (s: string) => encodeAnsi16Bright(s, 2),
+  yellowBright: (s: string) => encodeAnsi16Bright(s, 3),
+  blueBright: (s: string) => encodeAnsi16Bright(s, 4),
+  magentaBright: (s: string) => encodeAnsi16Bright(s, 5),
+  cyanBright: (s: string) => encodeAnsi16Bright(s, 6),
+  whiteBright: (s: string) => encodeAnsi16Bright(s, 7),
+  bold: (s: string) => encodeColor(s, 1, 22),
+  dim: (s: string) => encodeColor(s, 2, 22),
+};
```
tido64: Why are we adding so many colors? Where are they being used? The whole point with this package was to make it small/fast. If we're adding this many colors, should we consider using chalk instead?

JasonVMo: We effectively had five of these defined already (red, yellow, cyan, dim, bold) but not in a reusable way. These are the standard set of Ansi 16 colors that all libraries have, with the same naming convention as chalk. If we want to add a very lightweight colors package this feels like the minimum viable implementation.

The reason I'm adding them is that in OMR we end up bundling chalk/logging for bootstrap/yarn/node wrappers and this adds size to everything and overhead to every process launch. I'm going to streamline the reporter, fully align it with OMR implementation, and get rid of the chalk dependency there as a next step.

I could add the ansi color functions there but it feels wrong to define some here with a full set there. I've also considered doing something like having the red function have an attached function bright so that you can call red(myText) or red.bright(myText), but while it would make the code ever so slightly shorter, it doesn't really change the overall number of functions.

JasonVMo: If you would prefer I could just define the constants for the color values in this package and add the functions to the reporter package.

tido64: Given that this is a package for logging things to the console, I don't think exporting more colors is a good fit. We have some in there, but mostly for convenience. What worries me is that we'll have a bunch of colors but only use a few. If we know which colors we use in OMR, can we reduce the available colors to just those? It would also be great if we could "abstract away" when they are used, like with `error`, `info`, `warn`.

PR #3746, opened 2025-08-18
chore: bump `react-native` to 0.81

tido64:
### Description

Bump `react-native` to 0.81

### Test plan

n/a


packages/cli/src/start.ts
```diff
@@ -7,9 +7,14 @@ import {
...
-import type { ReportableEvent, Reporter, RunServerOptions } from "metro";
+import type {
+  ReportableEvent,
+  Reporter,
+  RunServerOptions,
+  RunServerResult,
+} from "metro";
 import type { Middleware } from "metro-config";
-import type Server from "metro/src/Server";
+import type Server from "metro/private/Server";
```
Saadnajmi, line 17: somewhat worrying, but alright

tido64: We've always been using Metro's private bits so I think this is fine. As long as they don't start blocking access completely.

PR #3764, opened 2025-09-09
Port forbid-foreach-with-variables-outside-of-function-scope eslint rule from Outlook

Farcemolina:
### Description
Port forbid-foreach-with-variables-outside-of-function-scope from Outlook custom rules, this is to start the progress to share Outlook rules into 1JS. 

### Test plan


Saadnajmi:
Probably better not to mention Microsoft internal repos or links in Github, as an FYI

PR #3766, opened 2025-09-10
chore(align-deps): parse and reconstruct GitHub url

tido64:
### Description

For packages that don't have a homepage url, we fallback to the repository url. But in some cases, the SSH url is provided instead of HTTP.

### Test plan

Before:

```
% yarn update-profile --pr --release-candidate | grep popover
| popover           | [react-native-popover-view](git@github.com:steffeydev/react-native-popover-view.git)                                       | ^6.0.1   | 6.1.0       | 9 months ago    |
```

After:

```
% yarn update-profile --pr --release-candidate | grep popover
| popover           | [react-native-popover-view](https://github.com/steffeydev/react-native-popover-view)                                       | ^6.0.1   | 6.1.0       | 9 months ago    |
```


packages/align-deps/scripts/github.ts
```diff
@@ -87,3 +87,12 @@ export async function fetchPullRequestFeedback(
 
   info(`No feedback found for pull request #${pr.number}`);
 }
+
+export function parseGitHubUrl(url: string | undefined): string | undefined {
```
Saadnajmi, line 91: why not `string?`

tido64: Because `url` is `string | undefined`

PR #3781, opened 2025-09-18
ci(ios): use prebuilt binaries

tido64:
### Description

Use prebuilt binaries when building iOS.

Significantly reduces build times without needing Ccache:

image

### Test plan

CI should pass.


packages/test-app/ios/Podfile
```diff
@@ -9,7 +9,7 @@ workspace 'SampleCrossApp.xcworkspace'
 options = {
   :bridgeless_enabled => true,
   :fabric_enabled => true,
-  :hermes_enabled => false,
+  :hermes_enabled => true,
```
Saadnajmi, line 12: Intentional with this change?

tido64: Yeah, we haven't built jsc since 0.80 ripped it out of core.

PR #3791, opened 2025-09-23
ci: build macOS and Windows

tido64:
### Description

Add macOS and Windows to CI

### Test plan

CI should pass

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -8,8 +8,7 @@
       "{workspaceRoot}/packages/eslint-config/*",
       "{workspaceRoot}/packages/eslint-plugin/*",
       "{workspaceRoot}/packages/jest-preset/*",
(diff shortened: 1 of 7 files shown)


nx.json
```diff
@@ -8,8 +8,7 @@
       "{workspaceRoot}/packages/eslint-config/*",
       "{workspaceRoot}/packages/eslint-plugin/*",
       "{workspaceRoot}/packages/jest-preset/*",
-      "{workspaceRoot}/scripts/tsconfig-esm.json",
-      "{workspaceRoot}/scripts/tsconfig-shared.json",
+      "{workspaceRoot}/packages/tsconfig/*",
```
Saadnajmi, line 11: Curious why this changed?

JasonVMo: I believe this is because we formalized the tsconfigs coming out of that package and there are multiple versions published.

PR #3849, opened 2025-11-06
docs(command example): fix missing space

wneel:
### Description

Hi, I saw there is a missing space on the run command component as you can see below. When copy pasting the space is not included and we can't just copy paste.

So I added the missing space!

Before:
image

After:
image

It did not resolve any issue

### Test plan

It's just a space so i tried to run the project in local and it worked fine :)


tido64 (approved):
Thanks!

PR #3915, opened 2026-01-08
fix(align-deps): align react-native-screens capability with the project README

balazsgerlei:
### Description

Recent releases of `react-native-screens` differ in terms of React Native version compatibility, for example their 4.14.0+ versions only support React Native 0.79.0+ and their 0.19.0+ versions only support React Native 0.81.0+. I updated the capabilities to be aligned with their stated compatibility.

See README of `react-native-screens` where it defines compatibility: https://github.com/software-mansion/react-native-screens?tab=readme-ov-file#support-for-fabric

### Test plan

```
```


tido64 (approved):
Thanks!

PR #3920, opened 2026-01-12
metro-resolver-symlinks support for metro 0.82

acoates-ms:
### Description
Update @rnx-kit/metro-resolver-symlinks to handle the move away from commonjs modules in metro.
Update @rnx-kit/metro-resolver-symlinks to patch the new location of the getSha1 function.


NikoAri (approved):

packages/metro-resolver-symlinks/src/utils/metro.ts
```diff
@@ -25,7 +25,7 @@ function fileExists(path: string): boolean {
 function importMetroModule(path: string) {
   const modulePath = findMetroPath() + path;
   try {
-    return require(modulePath);
+    return require(modulePath).default ?? require(modulePath);
```
tido64:
```suggestion
    const mod = require(modulePath);
    return mod.default ?? mod;
```

packages/metro-resolver-symlinks/src/utils/metro.ts
```diff
@@ -158,31 +158,63 @@ export function patchMetro(options: Options): void {
...
+          e.message.startsWith("Failed to get the SHA-1")
+        ) {
+          // Paths generated by pnpm setups include version numbers and/or hashes
+          if (filePath.includes(".pnpm-store") || filePath.includes(".store")) {
+            return { sha1: filePath };
+          }
+
+          const stat = fs.lstatSync(filePath);
+          return { sha1: filePath + "|" + stat.mtime.toISOString() };
+        }
```
tido64, line 207: Can we dedupe this logic?
```ts
function computeSha1(filePath: string): string {
  // Paths generated by pnpm setups include version numbers and/or hashes
  if (filePath.includes(".pnpm-store") || filePath.includes(".store")) {
    return filePath;
  }

  const stat = fs.lstatSync(filePath);
  return { sha1: filePath + "|" + stat.mtime.toISOString() };
}
```
Then we call `computeSha1()` in `getSha1` and `getOrComputeSha1`.

packages/metro-resolver-symlinks/src/utils/metro.ts
tido64: We should also remove the > 0.81 check.

PR #3959, opened 2026-02-10
Add `@rnx-kit/fork-sync` package

vmoroz:
## Summary

- Adds `@rnx-kit/fork-sync`, an experimental package for syncing forked/vendored dependencies with upstream using Git's native 3-way merge
- Ships two CLI tools: `fork-sync` (sync orchestrator) and `ai-merge` (AI-assisted conflict resolver using Claude or GitHub Copilot)
- Zero runtime dependencies — uses only Node.js built-in modules
- Includes a number of unit tests covering all major modules
- Configuration via `sync-manifest.json`, per-dependency `sync-config.json`, `.syncignore` exclusions, and hierarchical `sync-instructions.md` for AI merge guidance

## What's in the package

| Component | Description |
|-----------|-------------|
| `fork-sync` CLI | Clones upstream, creates work branch with local changes, performs Git 3-way merge, copies result back |
| `ai-merge` CLI | Parses conflict markers into hunks, sends each to AI for resolution with confidence-based validation |
| 13 reusable modules | Git operations, process spawning, parallel async iterators, TTY UI, logging, and more |

## Files added

- `incubator/fork-sync/` — the full package (source, tests, harnesses, docs)
- `.changeset/cold-numbers-talk.md` — changeset for initial minor release
- Minor update to `scripts/src/process.js` for monorepo build support

### DIFF
--- a/incubator/fork-sync/package.json
+++ b/incubator/fork-sync/package.json
@@ -31,8 +31,8 @@
     "fix": "tsc --noEmit && rnx-kit-scripts format && rnx-kit-scripts lint --fix",
     "format": "rnx-kit-scripts format",
     "lint": "rnx-kit-scripts lint",
-    "test": "rnx-kit-scripts test",
-    "test:all": "node --env-file=.env.integration --test test/*.test.ts"
+    "test": "tsc -p tsconfig.test.json && node --test lib-test/test/*.test.js",
+    "test:all": "tsc -p tsconfig.test.json && node --env-file=.env.integration --test lib-test/test/*.test.js"
   },
   "keywords": [
     "git",
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 !.yarn/releases/
 /incubator/*/dist/
 /incubator/*/lib/
+/incubator/*/lib-test/
 /incubator/@react-native-webapis/*/lib/
 /packages/*/*.LICENSE.txt
 /packages/*/*/rnx-build/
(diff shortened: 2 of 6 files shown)


tido64 (approved):

incubator/fork-sync/package.json
```diff
@@ -31,8 +31,8 @@
     "fix": "tsc --noEmit && rnx-kit-scripts format && rnx-kit-scripts lint --fix",
     "format": "rnx-kit-scripts format",
     "lint": "rnx-kit-scripts lint",
-    "test": "rnx-kit-scripts test",
-    "test:all": "node --env-file=.env.integration --test test/*.test.ts"
+    "test": "tsc -p tsconfig.test.json && node --test lib-test/test/*.test.js",
+    "test:all": "tsc -p tsconfig.test.json && node --env-file=.env.integration --test lib-test/test/*.test.js"
```
tido64, line 35: Wait, why doesn't `rnx-kit-scripts test` work here? Why do we need to transpile first?

vmoroz: Please see the PR failures in the initial submission. The tests are using the "using" statements and it seems that the Node.js 22.x does not recognize it. There are no issues when we run them on Node.js 24.x. Thus, I had to convert the TypeScript to JavaScript to satisfy the PR check requirements. The alternative was to avoid the disposable pattern in tests, but I did not like it.

The initial PR tests only had 2 failures because we use it in two test files, but each file has many such cases. It seems that Node.js exits on the first syntax error it finds in a file and did not try other tests.

tido64: Weird that this works locally with the same Node version:

image

I'll leave this as an exercise for another time.

PR #3961, opened 2026-02-10
feat(core-types): Create @rnx-kit/core-types package and consume types from that package

JasonVMo:
### Description

This change creates a new `@rnx-kit/core-types` package for core configuration types, manifest types, lint types, and so on. This avoids pulling in implementations when not needed and generally cleans up the dependency graph. 

There are also several places where using rnx-kit packages via an API from outside rnx-kit weren't exposing types correctly because the types came from a devDependency package (because the implementation wasn't needed) but as a result the types weren't being exported correctly. Having a lightweight types package allows for these to be correctly marked as dependencies causing the types to resolve correctly from outside the repo.

- despite the move, all packages that previously exported types still do so, they just now do it via re-exporting the types from the new package.

#### KitConfig and Supporting Types

The `KitConfig` was coming out of the `@rnx-kit/config` package which was pulling in several metro plugin packages for the types. Some plugin types were defined in the config package, some were defined locally, regardless it created a bit of a dependency snarl. These were the primary types moved into the new package.

- the plugin configurations that were still in their own packages were moved into the core-types
- esbuild and metro are marked as optional peer dependencies which allows those types to be fulfilled when necessary without creating a costly transitive dependency install chain.
- the align deps config was separated out and defined on its own with the ability to template it with extended capabilities. This makes creating custom presets more ergonomic.

#### Package Manifest Types

The `PackageManifest` type has been moved from tools-node into core-types. It has also been greatly improved by:
- adding most of the standard package.json types
- adding a typed entry for the KitConfig (this was impossible due to circular dependency issues before)

I also added a `PackageData` type which is designed as a parameter type for functions that want to take a folder and manifest pairing which can be fulfilled by `PackageInfo` or by other implementations. This will be used in subsequent changes.

#### AllPlatforms

I also moved the `AllPlatforms` type here, and changed it to be based on an array definition ensuring those stay in sync.

### Test plan

This is almost 100% type only (with the exception of the `ALL_PLATFORM_VALUES` array so automated tests + pipelines should be sufficient.

### DIFF
--- a/packages/types-kit-config/package.json
+++ b/packages/types-kit-config/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@rnx-kit/types-kit-config",
   "version": "0.0.1",
-  "description": "core types",
+  "description": "Type definitions for the rnx-kit configuration",
   "homepage": "https://github.com/microsoft/rnx-kit/tree/main/packages/types-kit-config#readme",
   "license": "MIT",
   "author": {
--- a/packages/types-node/package.json
+++ b/packages/types-node/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@rnx-kit/types-node",
   "version": "0.0.1",
-  "description": "node-types",
+  "description": "Type definitions for package manifests and other core node/npm types",
   "homepage": "https://github.com/microsoft/rnx-kit/tree/main/packages/types-node#readme",
   "license": "MIT",
   "author": {
--- a/packages/types-plugin-cyclic-dependencies/tsconfig.json
+++ b/packages/types-plugin-cyclic-dependencies/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "extends": "@rnx-kit/tsconfig/tsconfig.node.json",
   "compilerOptions": {
-    "emitDeclarationOnly": true
+    "noEmit": true
   },
   "include": ["src"]
 }


JasonVMo:
> In my experience, these kinds of catch-all type packages tend to accumulate unrelated or legacy definitions over time. Without clear, enforced criteria for what qualifies as "core", the package can gradually become a dumping ground for anything that’s shared somewhere, including types tied to features that later change or disappear. When there’s also no explicit policy for ownership, review standards, and deprecation/removal, it becomes hard to keep the package coherent and healthy.
> 
> The long-term risk is that the module turns into a tightly coupled dependency hub: changes become high-blast-radius, the package becomes intimidating to modify, and it eventually slows down refactoring because it's easier to add "just one more type" than to untangle things properly.
> 
> Instead, I think we should prefer package-/domain-specific type packages. If another package needs to consume types from `@rnx-kit/config`, we should publish something like `@rnx-kit/config-types` and keep the scope intentionally narrow and owned by the same domain. This keeps dependencies and responsibility clear, avoids a single "everything depends on this" package, and allows types to evolve with the package they represent.
> 
> If we go this route, the guideline could be:
> 
> * No centralized types package.
> * Types should live with their owning package, and only be extracted into a *-types package when there’s a clear consumer need outside that package.
> * Each *-types package should have an explicit scope tied to that domain (e.g., config-related types only).

Makes sense. I've split the types packages into three more specific packages in the latest iteration.

packages/bundle-types/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@rnx-kit/bundle-types",
+  "version": "1.0.0",
```
tido64: *All* new packages should have version 0.0.1. Changesets will bump them to 1.0 if you run `yarn change` and mark these as major changes.

packages/bundle-types/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "directory": "packages/bundle-types"
+  },
+  "engines": {
+    "node": ">=18.12"
+  },
+  "scripts": {
+    "build": "rnx-kit-scripts build",
+    "format": "rnx-kit-scripts format",
+    "lint": "rnx-kit-scripts lint",
+    "test": "rnx-kit-scripts test"
```
tido64: For all new packages: We can remove the test command since there are no tests to run.

packages/types-bundle-config/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "author": {
+    "name": "Microsoft Open Source",
+    "email": "microsoftopensource@users.noreply.github.com"
+  },
+  "files": [
+    "lib/**/*.d.ts",
+    "lib/**/*.js"
+  ],
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
```
tido64, line 17: For all new packages: Can't we export `src/**/*.ts` instead? And should they be ESM modules?

packages/types-bundle-config/README.md
```diff
@@ -0,0 +1,176 @@
...
+import type {
+  BundleConfig,
+  BundleParameters,
+  ServerConfig,
+  AllPlatforms,
+  Plugin,
+} from "@rnx-kit/bundle-types";
+```
+
+## Types
```
tido64, line 34: For all new packages: Can this section be generated by the `update-readme` script?

packages/cli/test/bundle/kit-config.test.ts
```diff
@@ -127,7 +127,7 @@ describe("bundle/kit-config/getCliPlatformBundleConfigs()", () => {
     });
   });
 
-  const testMultiConfig: KitConfig = {
```
tido64, line 130: Why was the type removed?

packages/types-metro-serializer-esbuild/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "@rnx-kit/tsconfig/tsconfig.node.json",
+  "compilerOptions": {
+    "emitDeclarationOnly": true
+  },
```
tido64, line 5: This differs from the config in other packages. Should there be a `tsconfig.types.json`?

packages/config/src/index.ts
```diff
@@ -3,8 +3,18 @@ export type {
...
+} from "@rnx-kit/bundle-types";
+
+export type {
+  Capability,
+  DependencyVersions,
+  GetDependencyVersions,
+  KitConfig,
+  KitType,
+  MetaCapability,
+} from "@rnx-kit/config-types";
```
tido64: Are these exports still necessary? If we can't remove them, can we mark them as deprecated?

packages/metro-plugin-cyclic-dependencies-detector/src/detectCycles.ts
```diff
@@ -1,16 +1,11 @@
+import type { CyclicDetectorOptions } from "@rnx-kit/bundle-types";
```
tido64: This is kinda weird. `CyclicDetectorOptions` belongs to this package and should be exported from here instead.

packages/metro-plugin-duplicates-checker/src/checkForDuplicatePackages.ts
```diff
@@ -1,3 +1,4 @@
+import type { DuplicateDetectorOptions as Options } from "@rnx-kit/bundle-types";
```
tido64: Same here. `DuplicateDetectorOptions` belongs here and should be consumed from here instead (or its types package).

packages/metro-serializer-esbuild/src/index.ts
```diff
@@ -1,7 +1,8 @@
+import type { EsbuildOptions as Options } from "@rnx-kit/bundle-types";
```
tido64: I think esbuild options should also come from this package (or its types package).

packages/types-node/src/index.ts
```diff
@@ -0,0 +1,10 @@
+// export types from manifest.ts
```
tido64: Redundant comment.
```suggestion
```

packages/node-types/src/manifest.ts
```diff
@@ -0,0 +1,133 @@
...
+export type ExportsGroup = {
+  types?: string;
+  import?: string;
+  require?: string;
+  default?: string;
+} & Record<string, string | Record<string, string>>;
+
+export type PackageExports = Record<string, string | ExportsGroup>;
+
+export type PersonEntry = {
```
tido64: This doesn't have to be a person.
```suggestion
export type AuthorEntry = {
```

packages/node-types/src/manifest.ts
```diff
@@ -0,0 +1,133 @@
...
+  name: string;
+  email?: string;
+  url?: string;
+};
+
+export type PackagePerson = string | PersonEntry;
```
tido64: Likewise:
```suggestion
export type PackageAuthor = string | AuthorEntry;
```

packages/types-node/src/manifest.ts
```diff
@@ -0,0 +1,133 @@
...
+  /** main or CJS entry point */
+  main?: string;
+  /** ESM entry point */
+  module?: string;
+  /** Browser entry point */
+  browser?: string;
+  /** TypeScript declaration entry point */
+  types?: string;
+  /** @deprecated Use `types` instead */
+  typings?: string;
```
tido64: We should just remove this.
```suggestion
```

packages/tools-node/package.json
```diff
@@ -53,8 +53,10 @@
     "build": "rnx-kit-scripts build",
     "format": "rnx-kit-scripts format",
     "lint": "rnx-kit-scripts lint",
-    "test": "rnx-kit-scripts test",
-    "update-readme": "rnx-kit-scripts update-api-readme"
```
tido64, line 57: Once you set versions of new packages to 0.0.1, this command should start working again.

packages/cli/test/bundle/kit-config.test.ts
JasonVMo: It's a subtlety of type inferencing. When you set a type here, it is good for enforcing that the constant you are creating follows the right shape, but the type ends up as the specified type, in this case KitConfig. When you don't set a type, the shape is of this particular object shape, that means you get an acceptance check when passing it in to a function that expects a KitConfig, in other words does this shape fulfill the contract. 

But when using it below it is still a kitConfig which means that the types that definitely exist in this object shape will show up as optional because of the config type. So in short, this avoids adding type assertions to the test code.

It wasn't happening before because the types were not resolving reliably and we weren't always getting a full KitConfig definition.

packages/types-metro-serializer-esbuild/tsconfig.json
JasonVMo: This is overly specific. Things like emitOnly or emitDeclarationOnly are still variations on the same config.

packages/types-bundle-config/package.json
JasonVMo: It's more optimized and standard to emit the .d.ts files and reference those with types. I do think including the .ts files is generally good practice, particularly if you have an export that references them.

packages/config/src/index.ts
JasonVMo: We can remove them. I was trying to maintain API consistency but most of our tools aren't called via API anyways. It does make some things easier.

packages/node-types/src/manifest.ts
JasonVMo: This was for consistency with the previous definition in tools-node. It's used for both author and contributors. Maybe ContributorEntry is better?

packages/types-node/src/manifest.ts
JasonVMo: I want to leave it in as it is something that is worth using constraints or package linting to migrate/flag if you have it set.

packages/types-bundle-config/package.json
JasonVMo: For type only packages it doesn't really matter. I kept the one that exports a constant as CJS for the time being as we need to work from root packages down to the leaf packages for the conversion.

packages/tools-node/package.json
JasonVMo: The issue was that our PR check github script doesn't do a build beforehand, as a result the `export type { Mytype } from '@rnx-kit/types-node';` lines would try to reference the index.d.ts files which would cause the script to fail. Once I removed the export forwarding this went away so I re-enabled the update-readme.

packages/types-bundle-config/README.md
JasonVMo: I tried it and the documentation ends up quite a bit worse. We will likely want to look at some other way of refreshing the documentation every so often.

incubator/tools-typescript/package.json
```diff
@@ -43,13 +43,16 @@
...
     "@rnx-kit/typescript-service": "^2.0.0"
   },
   "devDependencies": {
     "@rnx-kit/eslint-config": "*",
     "@rnx-kit/jest-preset": "*",
     "@rnx-kit/scripts": "*",
-    "@rnx-kit/tsconfig": "*"
+    "@rnx-kit/tsconfig": "*",
+    "@rnx-kit/types-kit-config": "^0.0.1",
+    "@rnx-kit/types-node": "^0.0.1"
```
tido64: For dev dependencies, we should use `*` for now until we figure out a way to use `workspace:*`
```suggestion
    "@rnx-kit/types-kit-config": "*",
    "@rnx-kit/types-node": "*"
```

packages/types-bundle-config/package.json
tido64, line 1: I still don't think it is correct to lump all plugins together like this. It is now too easy for someone to "see patterns" and make them interconnected. We should either keep them within the plugin package, or create separate types packages. I don't believe the number of packages in the repo is going to dominate our CI times.

packages/types-kit-config/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "@rnx-kit/tsconfig/tsconfig.node.json",
+  "compilerOptions": {
+    "emitDeclarationOnly": true
```
tido64: I still don't understand this. If this is a types only package, why do we need to emit anything at all? Can we not just consume the TypeScript?

packages/types-node/src/manifest.ts
tido64: We can still check whether `typings` is present in the manifest regardless. We should remove it so that people can't use it.

CLAUDE.md
tido64, line 1: This should be a separate PR. Let's not scope creep.

packages/types-kit-config/tsconfig.json
JasonVMo: It's a slight build optimization as typically .ts files are slower to parse and reference than .d.ts fiels and it is more common practice to use .d.ts files. That being said these files are types only so I don't know that it really matters in this case. I went ahead and switched it.

packages/types-kit-config/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@rnx-kit/types-kit-config",
+  "version": "0.0.1",
+  "description": "core types",
```
tido64: Can we get a better description here?

packages/types-bundle-config/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "name": "Microsoft Open Source",
+    "email": "microsoftopensource@users.noreply.github.com"
+  },
+  "files": [
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "src"
+  ],
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
```
tido64, line 17: Can this be TypeScript only?

packages/types-node/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@rnx-kit/types-node",
+  "version": "0.0.1",
+  "description": "node-types",
```
tido64: Better description?

packages/types-bundle-config/package.json
JasonVMo: This one exports a constant array from which the AllPlatforms type is derived, so typescript only won't work here.

tido64: But then it's no longer a types only package?

JasonVMo: This is a bit of a gray area I've encountered. In cases where you have discrete union types, and where you want the ability to validate the array, turning a string into the union type, the best approach is to to something like:
```ts
const VALID_VALUES = ["value1", "value2", "value3"] as const;
export type MyValidValues = (typeof VALID_VALUES)[number];
```
This is equivalent to `export type MyValidValues = "value1" | "value2" | "value3";` but with the added benefit of a single source of truth. With this you can easily write a type assertion or conversion function.

You can define a constant array elsewhere of the values, which will cause a build-break if a value is removed, but will not break when a value is added.

In short I think this is the best pattern for defining these kinds of types when you will want to validate user-input against the type. The package can be treated as type only by most consumers, but for code that wants to do type validation it can consume the constants as well. These constants are effectively the backing for the types so I think this is still consistent with a package called types.

PR #3969, opened 2026-02-16
chore: opt-in to TypeScript native preview

tido64:
### Description

Use TypeScript native preview wherever we can. `metro-config` is currently the only package that falls back to `tsc` because `tsgo` doesn't correctly deal with the combination of CJS and `noEmit`.

### Test plan

CI should pass.

### DIFF
--- a/incubator/rn-changelog-generator/tsconfig.json
+++ b/incubator/rn-changelog-generator/tsconfig.json
@@ -2,8 +2,7 @@
   "extends": "@rnx-kit/tsconfig/tsconfig.node.json",
   "compilerOptions": {
     "alwaysStrict": false,
-    "rootDir": "src",
-    "lib": ["ES2019.Object"]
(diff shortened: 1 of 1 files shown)


incubator/rn-changelog-generator/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "extends": "@rnx-kit/tsconfig/tsconfig.node.json",
   "compilerOptions": {
-    "lib": ["ES2019.Object"],
-    "alwaysStrict": false
+    "alwaysStrict": false,
+    "rootDir": "src",
+    "lib": ["ES2019.Object"]
```
Saadnajmi: ?

tido64: Inherited from RN upstream. Probably no longer needed.

PR #3970, opened 2026-02-17
feat(eslint-plugin): add rule for types only modules

tido64:
### Description

Added rule for types only modules.

### Test plan

Tests added, and enabled repo-wide.


tido64:
> I'm not as familiar with what eslint provides, but I wonder if there is a way we can allow the const array pattern we were discussing. In other words allow const arrays that are referenced by types. I guess the array could also be put in a .js file.

I'm pretty sure there is a way to express this but maybe we'll just allow literal constants or something like that. I'll take a look in a separate PR.

JasonVMo (approved):

PR #3977, opened 2026-02-18
feat(tools-filesystem): Add a richer set of filesystem helpers to tools-filesystem

JasonVMo:
### Description

This enhances the set of filesystem helpers in tools-filesystem.

#### Changes to existing functions
- renamed `ensureDir` and `ensureDirForFile` to follow standard sync naming conventions.
- restructured the files to create an index file and dedicated files for types of functionality.
- enhanced the reworked the mockFS to provide much more functionality including promise based mocks.

#### New functionality
- asynchronous versions of the helper functions. Pairings follow fs naming (e.g. ensureDirSync + ensureDir)
- helpers for reading files with UTF encoding
- helpers for parsing JSON with BOM stripping, and serializing with trailing "\n" addition which is always omitted by JSON.serialize.
- helpers for reading/writing JSON files which use these helpers.
- optimized file compares which do cheap compares (size, ino/dev) then content comparisons. Sync and async versions provided.
- `FSEntry` class which provides a wrapper around a path, progressively caching information and coalescing promise based filesystem calls. This is useful to either implement a caching layer, or to be handed out by a caching layer.
- ability to use the mockFS to test the new functionality.
- tests for all the new functionality.

### Test plan

Automated tests.

### DIFF
--- a/packages/cli/test/copy-assets/assembleAarBundle.test.ts
+++ b/packages/cli/test/copy-assets/assembleAarBundle.test.ts
@@ -1,8 +1,18 @@
 import { readTextFileSync as readText } from "@rnx-kit/tools-filesystem";
 import { getMockFSFiles, mockFS } from "@rnx-kit/tools-filesystem/mocks";
 import * as child_process from "node:child_process";
+import { platform } from "node:os";
 import * as path from "node:path";
 import { assembleAarBundle } from "../../src/copy-assets.ts";
+
+const toPosix =
+  platform() === "win32"
+    ? (input: string) => {
+        const posixPath = input.replaceAll(path.win32.sep, path.posix.sep);
+        // This regex replaces the drive letter and colon at the start of the string
+        return posixPath.replace(/^[a-zA-Z]:/, "");
+      }
+    : (input: string) => input;
 
 jest.mock("node:child_process");
 jest.unmock("@rnx-kit/console");
@@ -151,11 +161,11 @@ describe("copy-assets/assembleAarBundle()", () => {
       })
     );
     expect(Object.entries(files)).toEqual([
-      [gradleWrapper, ""],
-      [authBuildGradle, path.basename(authBuildGradle)],
-      [authBuildArtifact, path.basename(authBuildArtifact)],
-      [authManifest, dummyManifest],
-      [rnManifest, dummyManifest],
+      [toPosix(gradleWrapper), ""],
+      [toPosix(authBuildGradle), path.basename(authBuildGradle)],
+      [toPosix(authBuildArtifact), path.basename(authBuildArtifact)],
+      [toPosix(authManifest), dummyManifest],
+      [toPosix(rnManifest), dummyManifest],
       [
         expect.stringMatching(
           /[/\\]node_modules[/\\].rnx-gradle-build[/\\]rnx-kit_react-native-auth[/\\]build.gradle$/
@@ -218,12 +228,12 @@ describe("copy-assets/assembleAarBundle()", () => {
       })
     );
     expect(Object.entries(files)).toEqual([
-      [gradleWrapper, ""],
-      [authBuildGradle, path.basename(authBuildGradle)],
-      [authBuildArtifact, path.basename(authBuildArtifact)],
-      [authSettingsGradle, path.basename(authSettingsGradle)],
-      [authManifest, dummyManifest],
-      [rnManifest, dummyManifest],
+      [toPosix(gradleWrapper), ""],
+      [toPosix(authBuildGradle), path.basename(authBuildGradle)],
+      [toPosix(authBuildArtifact), path.basename(authBuildArtifact)],
+      [toPosix(authSettingsGradle), path.basename(authSettingsGradle)],
+      [toPosix(authManifest), dummyManifest],
+      [toPosix(rnManifest), dummyManifest],
       [
         expect.stringMatching(
           /dist[/\\]aar[/\\]rnx-kit_react-native-auth-0.0.0-dev.aar$/


packages/tools-filesystem/src/mockfs/index.ts
tido64, line 1: Could we use [`memfs`](https://github.com/streamich/memfs) instead? The previous mock was simple and I didn't want to bring in another dependency. If this is a full blown implementation, we're better off using an existing solution.

packages/tools-filesystem/src/compare.ts
```diff
@@ -0,0 +1,171 @@
...
+export function isSameFileFromStats(
+  stats1: nodefs.BigIntStats,
+  stats2: nodefs.BigIntStats
+): boolean | undefined {
+  // Check if both stats refer to the same file, handles hardlinks and other ways to reference the
+  // same file on disk. This is a definitive check that does not require further checks. Note that these values can
+  // overflow non-BigInt stats on certain filesystems, so we need to use BigInt stats for this check.
+  if (
+    stats1.ino !== undefined &&
+    stats1.dev !== undefined &&
```
tido64: Shorter and would also work with `null`:
```suggestion
    stats1.ino &&
    stats1.dev &&
```

packages/tools-filesystem/src/compare.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  // do a quick stats comparison first to preempt more expensive checks if possible
+  const sameFromStats = isSameFileFromStats(f1.stats, f2.stats);
+  if (sameFromStats !== undefined) {
+    return sameFromStats;
+  }
+
+  // at this point, we know the sizes are the same, so we can use either file's size
+  const size = f1.size;
+  if (f1.contentLoaded || f2.contentLoaded || size <= CHUNK_SIZE) {
```
tido64: `size` isn't used elsewhere, might as well inline it:
```suggestion
  if (f1.contentLoaded || f2.contentLoaded || f1.size <= CHUNK_SIZE) {
```

packages/tools-filesystem/src/compare.ts
```diff
@@ -0,0 +1,171 @@
...
+  } finally {
+    if (fileA !== undefined) {
+      fs.closeSync(fileA);
+    }
+    if (fileB !== undefined) {
+      fs.closeSync(fileB);
+    }
+  }
+
+  return true;
```
tido64, line 106: Won't we be returning `true` here if an error is thrown during read?

packages/tools-filesystem/src/compare.ts
```diff
@@ -0,0 +1,171 @@
...
+      if (Buffer.compare(chunkA, chunkB) !== 0) {
+        return false;
+      }
+    }
+  } finally {
+    // ensure streams are closed
+    streamA.destroy();
+    streamB.destroy();
+  }
+}
```
tido64, line 174: This operation never throws?

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+};
+
+/**
+ * Helper function to convert a string path to an FSEntry if it is not already an FSEntry
+ * @param path The path to convert to an FSEntry
+ * @returns An FSEntry instance representing the path
+ */
+export function toFSEntry(
+  path: string | FSEntry,
+  /** @internal */ fsModule?: typeof fs
```
tido64: We should follow this pattern from previous files:
```suggestion
  /** @internal */ fsModule = nodefs
```

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+  protected _stats?: fs.BigIntStats;
+  protected _statsPromise?: Promise<fs.BigIntStats | undefined>;
+  protected _exists?: boolean;
+  protected _content?: string;
+  protected _contentPromise?: Promise<string>;
+  protected _needsWrite?: boolean;
+  protected _needsDirEnsure?: boolean;
+  protected _fs: typeof fs;
+
+  constructor(path: string, fsModule?: typeof fs) {
```
tido64:
```suggestion
  constructor(path: string, fsModule = nodefs) {
```

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
+import fs from "node:fs";
```
tido64: This should be `nodefs` to avoid using `fs.<function>`
```suggestion
import nodefs from "node:fs";
```

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+  get isFile(): boolean {
+    return this.stats.isFile();
+  }
+
+  /**
+   * get the size of the file
+   * @returns the size of the file in bytes as a number
+   */
+  get size(): number {
+    return Number(this.stats.size);
```
tido64, line 138: Won't this truncate the size if `this.stats.size` is a `BigInt`? Wouldn't that invalidate big file comparisons?

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+   * @param options - options for writing the file
+   */
+  writeContentsSync(options?: WriteOptions): void {
+    const content = this.getContentToWrite(options);
+    if (content !== undefined) {
+      if (this._needsDirEnsure) {
+        ensureDirForFileSync(this.path, this._fs);
+        this._needsDirEnsure = false;
+      }
+      this._fs.writeFileSync(this.path, content, WITH_UTF8_ENCODING);
```
tido64, line 197: Is this always text? Could it be binary?

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+    return this.writeContents();
+  }
+
+  /**
+   * synchronously write out the file
+   * @param options - options for writing the file
+   */
+  writeContentsSync(options?: WriteOptions): void {
+    const content = this.getContentToWrite(options);
+    if (content !== undefined) {
```
tido64: Handle `null`:
```suggestion
    if (content != undefined) {
```
Or shorter:
```suggestion
    if (content != null) {
```

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+  }
+
+  /**
+   * asynchronously write the content of the file
+   * @param options - options for writing the file
+   */
+  async writeContents(options?: WriteOptions): Promise<void> {
+    const content = this.getContentToWrite(options);
+    if (content !== undefined) {
```
tido64: Handle `null`:
```suggestion
    if (content != null) {
```

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  /**
+   * Get the content to write to the file, optionally adding a newline at the end.
+   * @param options - options for writing the file
+   * @returns the content to write, or undefined if no write is needed
+   */
+  private getContentToWrite(options?: WriteOptions): string | undefined {
+    const { force, newline } = options ?? {};
+    if (this._content === undefined || (!force && !this._needsWrite)) {
```
tido64:
```suggestion
    if (this._content == null || (!force && !this._needsWrite)) {
```

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+    return newline && !this._content.endsWith("\n")
+      ? this._content + "\n"
+      : this._content;
+  }
+
+  /**
+   * asynchronously check if the file exists, does not throw
+   */
+  async getExists(): Promise<boolean> {
+    return (this._exists ??= (await this.getStatsAsync()) !== undefined);
```
tido64:
```suggestion
    return (this._exists ??= (await this.getStatsAsync()) != null);
```

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  /**
+   * asynchronously get the content of the file, throws if the file does not exist
+   */
+  async getContent(): Promise<string> {
+    return (
+      this._content ??
+      (this._contentPromise ??= this._fs.promises
+        .readFile(this.path, WITH_UTF8_ENCODING)
```
tido64: Are we assuming that we will never read anything but text files?

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,303 @@
...
+      })
+    );
+  }
+
+  /**
+   * Throw an error if stats are undefined. Because this was potentially obtained via the non-throwing
+   * stats method for an existence check, synthesize the ENOENT error in this case.
+   */
+  private requireStats(stats: fs.BigIntStats | undefined): fs.BigIntStats {
+    if (stats === undefined) {
```
tido64:
```suggestion
    if (stats == null) {
```

packages/tools-filesystem/src/fileio.ts
```diff
@@ -0,0 +1,91 @@
...
+import { parseJson, serializeJson } from "./json.ts";
+
+/**
+ * Synchronously reads the content of a file as a UTF-8 string.
+ *
+ * @param filePath path to the file to read
+ * @param fs filesystem module to use (defaults to Node's fs)
+ * @returns the content of the file as a UTF-8 string
+ */
+export function readFileSync(
```
tido64: Should this (and its async variant) be `readTextFile*` to match `writeTextFile` below? As is, they're not generic file read operations.

packages/tools-filesystem/src/fileio.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Asynchronously reads the content of a file as JSON, stripping a UTF-8 BOM if present. This is a convenience method that combines
+ * readFile with JSON.parse, and is useful for reading JSON files without having to worry about BOMs causing parse failures.
+ *
+ * @param filePath path to the JSON file to read
+ * @param fs the filesystem module to use (defaults to Node's fs)
+ * @returns the parsed JSON object, optionally cast to type T
+ */
+export async function readJson<T = ReturnType<typeof JSON.parse>>(
```
tido64: Can we make the `Json` part of the function names consistent? Node seems to favor `JSON` and we should probably follow suite to make things look the same everywhere.
```suggestion
export async function readJSON<T = ReturnType<typeof JSON.parse>>(
```

packages/tools-filesystem/src/mockfs/index.ts
JasonVMo: I'll look into that. Want me to do it as part of this PR or as a separate one?

packages/tools-filesystem/src/entry.ts
JasonVMo: Good point. For the size comparison in the compare we use the size from stats, but for the iteration we use the size from the file object which could get truncated for very very large files. I was thinking about this a bit and I think we likely want to make explicit accessors for each type. If people don't realize the size is a bigint they can write code that does unexpected things without realizing it.

packages/tools-filesystem/src/entry.ts
JasonVMo: I think the second one is the one we want.

packages/tools-filesystem/src/entry.ts
JasonVMo: The convenience methods in here are generally for text files, both reading and writing. For binary/buffer types we need to handle the file similarly to what the compare functions do. That can be still done externally (not baked into the class) but caching with read streams is dangerous as they are stateful (unlike a string) and I didn't want to complicate things.

I'd prefer to have separate helpers for dealing with stream/buffer reads and cleanup but didn't want scope creep here.

packages/tools-filesystem/src/entry.ts
JasonVMo: The assumption is that content is text based. You can still open the file as a buffer if desired, but with the exception of file compares, hashing, and file copies, text is what we want when opening a file.

packages/tools-filesystem/src/fileio.ts
JasonVMo: Makes sense.

packages/tools-filesystem/src/compare.ts
JasonVMo: Added a catch block that returns false. Good catch.

packages/tools-filesystem/src/compare.ts
JasonVMo: No. I think because it is a cleanup utility often handled in blocks like this.

packages/tools-filesystem/src/entry.ts
JasonVMo: Made the change I mentioned. It will be in the next push.

packages/tools-filesystem/src/mockfs/index.ts
JasonVMo: I went ahead and switched it. This required some test updates as the files object is no longer mutated when passed in. An equivalent can be retrieved but paths will be absolute rather than relative so many of the comparison functions had to be reworked. Directories are handled differently as well.

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,310 @@
...
+import { ensureDirForFile, ensureDirForFileSync } from "./dirs.ts";
+import { createEnoentError } from "./errors.ts";
+import { readTextFile, readTextFileSync } from "./fileio.ts";
+import { parseJSON, serializeJSON } from "./json.ts";
+
+export type WriteOptions = {
+  /** whether to force writing the file even if the content is not marked as dirty (defaults to false) */
+  force?: boolean;
+
+  /** whether to end with a newline (defaults to false) */
```
tido64: POSIX requires text files to always end with a newline. This should default to true.

packages/tools-filesystem/src/entry.ts
```diff
@@ -0,0 +1,310 @@
...
+    this.writeContentsSync();
+  }
+
+  /**
+   * Asynchronously write out data as JSON, replacing the content of the file with the serialized
+   * JSON string, writing to disk asynchronously, and marking the content as not dirty after writing.
+   * @param data the data to serialize to JSON and write to the file
+   * @returns a promise that resolves when the write operation is complete
+   */
+  async writeJson(data: unknown): Promise<void> {
```
tido64: These should also have JSON in all caps.

tido64 (approved):

PR #4023, opened 2026-03-12
fix(babel-plugin-import-path-remapper): remap packages with exports field

sverrejoh:
## Summary

`findMainSourceFile` previously bailed out entirely when a package's `package.json` contained an `exports` field ([line 49](https://github.com/microsoft/rnx-kit/blob/main/packages/babel-plugin-import-path-remapper/src/index.js#L49)):

```js
if (exports) {
  // Skip packages that declare entry points
  return;
}
```

This prevented the `lib/` → `src/` remapping for any package that uses `exports`, even when the entry point still follows the standard `lib/index.js` → `src/index.ts` pattern. In the 1JS monorepo this caused Metro to load pre-built files (which have production localization transforms applied) instead of source files during development, breaking string resolution for React Native.

### Changes

- Extract the main entry point from `exports["."]` (checking `default`, `require`, `import` conditions) when `main` is not present
- When `main` IS present (the common case), behavior is unchanged — `main` is used as before
- `customRemap` callback is now reachable for packages with `exports`
- Added test fixture and 3 test cases for packages with `exports`

### Before

```
import { Foo } from "@scope/pkg-with-exports";
// → stays as "@scope/pkg-with-exports" (remapper skips)
// → Metro loads lib/ (pre-built, trimmed)
```

### After

```
import { Foo } from "@scope/pkg-with-exports";
// → remapped to "@scope/pkg-with-exports/src/index.ts"
// → Metro loads src/ (source)
```


tido64:

PR #4044, opened 2026-03-24
feat(tools-react-native): Add mergeTransformerConfigs function to @rnx-kit/tools-react-native

JasonVMo:
### Description

This adds a utility for merging transformer configurations to tools-react-native to help with transformer merging. I added this as part of a new submodule called metro-utils for general metro utility functions.

I considered various other places for this. The main package it would make sense in is @rnx-kit/metro-config but that isn't a utility package used by other packages in react-native and has issues if you try to include it as a dependency of the cli. Given the already existing utilities for resolving packages from metro it made the most sense to add it to tools-react-native.

The function itself is straightforward and will perform a flat merge (which is what metro does) except if there are multiple getTransformOptions implementations in which case they will be wrapped with the results merged together in order of precedence. This function is used to initialize the babel transformers, it is called once with the results being serialized via json to the metro transformers.

Having the merge function be baked into the CLI is required so that any transform options being set by a transformer configuration don't get stomped by the esbuild transformer which is always added at the end. The esbuild transform options will take precedence still, but things like custom options can still be used in this case.


packages/tools-react-native/src/metro-utils.ts
```diff
@@ -0,0 +1,77 @@
+import type { GetTransformOptions, TransformerConfigT } from "metro-config";
+
+/**
+ * Type guard to check if a value is a plain object (i.e., a record). This is used to ensure that we only attempt to
+ * recursively merge plain objects in the `simpleObjectMerge` function.
+ */
+function isRecord(value: unknown): value is Record<string, unknown> {
```
tido64, line 7: Should this be parameterized?
```suggestion
function isRecord<K extends string | number = string, V = unknown>(value: unknown): value is Record<K, V> {
```

packages/tools-react-native/src/metro-utils.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Simple merge helper that recursively merges plain objects. Note that array merges are not supported,
+ * as the behavior isn't deterministic (e.g., should we concatenate arrays, or override them?). If a property is an array in
+ * multiple configs, the value from the last config will win.
+ */
+function simpleObjectMerge(
+  ...options: Record<string, unknown>[]
+): Record<string, unknown> {
+  /** @type {Record<string, unknown>} */
```
tido64: Stray JSDoc
```suggestion
```

packages/tools-react-native/src/metro-utils.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @param configs one or more transformer config objects to merge. Later configs take precedence over earlier ones.
+ * @returns transformer configuration suitable for use by Metro
+ */
+export function mergeTransformerConfigs(
+  ...configs: Partial<TransformerConfigT>[]
+): Partial<TransformerConfigT> {
+  // collect the getTransformOptions functions from all configs, and if there are multiple, we'll create a wrapper function for them
+  const getTransformOptionsFns = configs
+    .map((config) => config?.getTransformOptions)
+    .filter((fn) => typeof fn === "function");
```
tido64: Whenever you see `.map` followed by `.filter`, please rewrite it as `.reduce` or a for-loop instead:
```suggestion
  const getTransformOptionsFns = configs
    .reduce<TransformerConfigT["getTransformOptions"][]>((result, config) => {
      const getTransformOptions = config?.getTransformOptions;
      if (typeof getTransformOptions === "function") {
          result.push(getTransformOptions);
      }
      return result;
    }, []);
```
JavaScript's implementation is eager and will cause additional unnecessary allocations.

packages/tools-react-native/test/metro-utils.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  });
+
+  it("merges scalar properties from multiple configs", () => {
+    const result = mergeTransformerConfigs(
+      { minifierPath: "/minifier-a", hermesParser: false },
+      { minifierPath: "/minifier-b", enableBabelRuntime: true }
+    );
+    equal(result.minifierPath, "/minifier-b");
+    equal(result.hermesParser, false);
+    equal(result.enableBabelRuntime, true);
```
tido64: Can we use `deepEqual` to ensure that nothing else gets added?

packages/tools-react-native/test/metro-utils.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  });
+
+  it("later configs override earlier ones", () => {
+    const result = mergeTransformerConfigs(
+      { hermesParser: false },
+      { hermesParser: true }
+    );
+    equal(result.hermesParser, true);
```
tido64: Can we use `deepEqual` to ensure that nothing else gets added?

packages/tools-react-native/test/metro-utils.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      [],
+      { dev: true, hot: true, platform: "ios" },
+      async () => []
+    );
+
+    deepEqual(options, {
+      transform: {
+        experimentalImportSupport: true,
+        inlineRequires: false,
+        nonInlinedRequires: ["react"],
```
tido64, line 121: Can you add a test for how arrays should be handled?

packages/tools-react-native/src/metro-utils.ts
JasonVMo: Not for this usage. Because the value is of type unknown you won't get any inference for free. It's possible the record type should be the more general Record<string | number | symbol, unknown> but for this local function this is sufficient.

tido64 (approved):

PR #4079, opened 2026-04-09
feat(tools-performance): Add @rnx-kit/tools-performance package

JasonVMo:
### Description

This adds a framework for doing lightweight performance tracing, some basic reporting, and enabling tracing at a global level.

#### Usage

In packages, to add instrumentation:
```ts
import { getTrace } from "@rnx-kit/tools-performance";

async function myFunction() {
  // will return a logging tracer if "transform" is enabled, a passthrough if not
  const trace = getTrace("transform");
  // trace a sync closure
  const result1 = trace("func1", () => doSomething(p1, p2));
  // trace a sync function without creating a closure
  const result2 = trace("process result, doSomethingElse, result1, options);
  // trace an async function and return
  const final = await trace("call async", finalCall, result2);
  return final;
}
```

Right now, to enable this for metro bundling a user would add in their metro.config.js:
```ts
import { trackPerformance } from "@rnx-kit/tools-performance";

// these would all work
trackPerformance("transform"); // only turn on this one area
trackPerformance(["transform", "metro"]); // turn on a set of areas
trackPerformance(/* or true */); // turn on all performance tracing
```

#### Future things to add (or consider)

- [ ] instrument various parts of our infrastructure
- [ ] add an implementation of unstable_perfLoggerFactory that hooks into this
- [ ] have our metro config automatically add the unstable_perfLoggerFactory if "metro" is enabled
- [ ] consider adding a cli option that will enable perf logging for our tools automatically

#### Other notes

- text is styled using the formatting from node:utils
- simple table formatting in the same format as console.table is added, except with support for formatted text.


incubator/tools-performance/src/perf.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export function reportPerfData(peekOnly = false) {
+  defaultManager?.report(peekOnly);
+}
+
+/**
+ * Check if tracking is enabled for a specific category. If category is undefined, checks if all tracking is enabled.
+ */
+export function isTrackingEnabled(category?: string) {
+  return defaultManager?.isEnabled(category) ?? false;
```
tido64:
```suggestion
  return Boolean(defaultManager?.isEnabled(category));
```

incubator/tools-performance/src/trace.ts
```diff
@@ -0,0 +1,104 @@
...
+    const start = performance.now();
+    const result = callFunction<TFunc>(fn, args);
+    if (isPromiseLike(result)) {
+      return result.then((res: ReturnType<TFunc>) => {
+        record(tag, performance.now() - start);
+        return res;
+      });
+    }
+    record(tag, performance.now() - start);
+    return result;
```
tido64: Can we use `performance.mark` and `performance.measure` instead of manually recording time? https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure#examples

Another added benefit is getting tracking for free: https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntries

incubator/tools-performance/README.md
```diff
@@ -0,0 +1,251 @@
+<!-- We recommend an empty change log entry for a new package: `yarn change --empty` -->
+
```
tido64: Remove
```suggestion
```

incubator/tools-performance/src/report.ts
```diff
@@ -0,0 +1,194 @@
...
+        const isError = colKey === "errors";
+        const style = isError ? "redBright" : undefined;
+        result[colKey] = formatNumber(value, style, isError);
+      } else {
+        result[colKey] = styleColumn(colKey, entry, maxOpWidth);
+      }
+    }
+    return result;
+  });
+  // use console.table for nice formatting
```
tido64: Do you plan to implement this in this PR or later?

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,154 @@
...
+    if (!callbacks) {
+      callbacks = [];
+      process.on("exit", () => {
+        for (const cb of callbacks!) {
+          cb();
+        }
+      });
+    }
+    callbacks.push(callback);
+  }
```
tido64: Avoid force nonnull assertion. Rewrite to help TypeScript infer nullability:
```suggestion
type OnExit = () => void;

export const exitHandler = (() => {
  let callbacks: OnExit[] | undefined = undefined;

  function add(callback: OnExit) {
    if (!callbacks) {
      const arr: OnExit[] = [];
      process.on("exit", () => {
        for (const cb of arr) {
          cb();
        }
      });
      callbacks = arr;
    }
    callbacks.push(callback);
  }
```

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  function remove(callback: () => void) {
+    if (callbacks) {
+      callbacks = callbacks.filter((cb) => cb !== callback);
+    }
```
tido64: Should we be using `Set<OnExit>` instead to ensure uniqueness and gain access to `.delete()`?

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,154 @@
...
+  enable(category: true | string | string[]) {
+    if (category === true) {
+      this.enabled.add(this.unscoped);
+    } else if (typeof category === "string") {
+      this.enabled.add(category);
+    } else {
+      for (const cat of category) {
+        this.enabled.add(cat);
+      }
+    }
```
tido64: We should probably also check that `category` is an actual array:
```suggestion
    } else if (Array.isArray(category)) {
      for (const cat of category) {
        this.enabled.add(cat);
      }
    } else {
      throw new Error(`invalid category: ${category}`);
    }
```

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,154 @@
...
+    return this.enabled.has(category ?? this.unscoped);
+  }
+
+  getTrace(category?: string): TraceFunction {
+    return this.getCategory(category)?.trace ?? nullTrace;
+  }
+
+  getRecorder(category?: string): TraceRecorder {
+    return this.getCategory(category)?.record ?? nullRecord;
+  }
```
tido64: These should return `undefined` since we're already handling nullable in the module-level `getRecorder`/`getTrace` methods.

incubator/tools-performance/src/trace.ts
JasonVMo: I'm reworking this a bit to be cleaner and will support both. The perf markers are heavier than timing based ones and have memory bloat issues. Instrumenting babel plugins with them would run out of memory. It only takes a slight adjustment to make either work.

incubator/tools-performance/src/report.ts
JasonVMo: Old comment. Will clean it up in the next iteration.

incubator/tools-performance/src/tracker.ts
JasonVMo: Great idea

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,221 @@
+import { styleText } from "util";
```
tido64:
```suggestion
import { styleText } from "node:util";
```

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,221 @@
...
+        }
+      });
+    }
+    this.exitHandlers.add(callback);
+  }
+
+  static removeExitHandler(callback: () => void) {
+    if (this.exitHandlers) {
+      this.exitHandlers.delete(callback);
+    }
```
tido64:
```suggestion
    this.exitHandlers?.delete(callback);
```

incubator/tools-performance/src/domain.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * End event for the performance namespace. This will finish the start events and do some cleanup. If running
+   * with the node strategy the marks for this namespace will be cleared.
+   * @param processExit is this happening as part of process exit.
+   */
+  stop(processExit = false) {
+    // first attempt to close the initial start event if it exists
+    if (this.startVal !== undefined) {
```
tido64: This supports both null/undefined and is shorter:
```suggestion
    if (this.startVal != null) {
```

incubator/tools-performance/src/domain.ts
```diff
@@ -0,0 +1,171 @@
...
+  private trace: TraceFunction;
+
+  private coerceTag(tag: string) {
+    return (this.tagMap[tag] ??= `${this.name}:${tag}`);
+  }
+
+  private timingRecorder(tag: string, startTime?: number) {
+    const timeNow = performance.now();
+    tag = this.coerceTag(tag);
+    if (startTime === undefined) {
```
tido64:
```suggestion
    if (startTime == null) {
```

incubator/tools-performance/src/domain.ts
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      this.recordTime(tag, timeNow - startTime);
+      this.lastTime = timeNow;
+    }
+    return timeNow;
+  }
+
+  private markingRecorder(tag: string, startMark?: string) {
+    tag = this.coerceTag(tag);
+    if (startMark === undefined) {
```
tido64:
```suggestion
    if (startMark == null) {
```

incubator/tools-performance/src/table.ts
```diff
@@ -0,0 +1,290 @@
...
+): CellEntry {
+  const { format, style, maxWidth } = config;
+  let text = format ? format(value) : undefined;
+  let key: string | number;
+  if (text === undefined && typeof value === "number") {
+    const { digits, intlFormatter } = config;
+    if (intlFormatter) {
+      text = intlFormatter.format(value);
+    } else {
+      text = digits !== undefined ? value.toFixed(digits) : String(value);
```
tido64:
```suggestion
      text = digits != null ? value.toFixed(digits) : String(value);
```

incubator/tools-performance/src/table.ts
```diff
@@ -0,0 +1,290 @@
...
+    }
+    key = value;
+  } else {
+    text ??= String(value);
+    key = text;
+  }
+  if (maxWidth !== undefined && text.length > maxWidth) {
```
tido64:
```suggestion
  if (maxWidth != null && text.length > maxWidth) {
```

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,221 @@
...
+    const timings = this.timings;
+    const current = timings.get(tag);
+    const entry = current ?? {
+      name: tag,
+      session: performance.now() - PerfTracker.startTime,
+      calls: 0,
+      completions: 0,
+      total: 0,
+    };
+    if (duration !== undefined) {
```
tido64:
```suggestion
    if (duration != null) {
```

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,221 @@
...
+    }));
+    if (maxNameWidth && cols.includes("name")) {
+      columnConfigs[cols.indexOf("name")].maxWidth = maxNameWidth;
+    }
+    // filter the sort to include columns in the report
+    const sort: number[] = [];
+    if (reportSort && reportSort.length > 0) {
+      for (const sortCol of reportSort) {
+        const index = cols.indexOf(sortCol);
+        if (index !== -1) {
```
tido64: Ensure positive index:
```suggestion
        if (index >= 0) {
```

incubator/tools-performance/src/tracker.ts
```diff
@@ -0,0 +1,221 @@
...
+          sort.push(index);
+        }
+      }
+    }
+    // maps the column keys to entries
+    const rows = Array.from(this.timings.values()).map(
+      (entry: OperationData) => {
+        return cols.map((col: PerfReportColumn) => getCellValue(entry, col));
+      }
+    );
```
tido64, line 171: Can we use `.forEach` instead of `Array.from` to avoid the temporary array?

tido64 (approved):
Mostly minor issues/nits

incubator/tools-performance/src/tracker.ts
JasonVMo: `Array.from` is actually pretty optimal in that it will create an appropriately sized array up front then fill in the values directly. There's no intermediate array. It's a good equivalent to Array.map usable with Maps.

tido64 (approved):

tido64 (approved):

PR #4087, opened 2026-04-13
feat(tools-babel): Add tools-babel package for babel configuration and parsing support

JasonVMo:
**NOTE: this currently includes the tools-performance change as it depends on that change. Ignore those files here.**

### Description

This adds a tools project for dealing with babel with regards to react-native transformations. This builds upon some of the work from the metro-transformer-oxc package as well as some of the work I was looking at for using native tools for transformation.

This package has tools that deal with babel configuration and parsing.

#### Babel Configuration

Several utilities are provided here:

- A central implementation of getting the babel configuration in a way compatible with what is in `@react-native/metro-babel-transformer`, which is the standard upstream babel transformer for react-native.
- Helpers for parsing babel's plugin and preset types.
- An implementation of the plugin visitor factory for instrumenting plugins.
- A routine to update babel configs with a visitor pattern that is will only make changes if something is updated. It can be used to update plugins or remove them entirely.
- A filter function for filtering plugins by key.

#### Babel Parsing

Parsing is one of the most time consuming and memory intensive parts of the transformation process. There are faster parsers but they generally produce estree ASTs rather than babel-ASTs. The current metro-transformer-oxc uses a utility estree-to-babel to convert the AST, which mostly works, but when timing the performance it ends up being almost as slow as babel parsing.

The parsing in here:

- can parse with oxc, hermes-parser, or babel
- has a reworked and heavily tuned AST converted that mutates in place, greatly reducing memory churn and making this drastically faster than hermes-parser end-to-end, which is itself substantially faster than babel.
- uses a huge set of test fixtures (now in the @rnx-kit/test-fixtures package) to validate and test AST compatibility both at a raw level, as well as with generation and transformation.

The tests are fairly wordy for output but this is useful for allowing tools like Claude to gain information about what is happening and what the differences are.


incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+
+// ── Imperative conversion functions ─────────────────────────────────
+// These handle complex structural transforms that don't fit a table.
+
+function convertLiteral(node: MutableNode): void {
+  if (node.type !== "Literal") return;
+
+  const { value, raw } = node;
+
+  if (raw !== undefined) {
```
tido64, line 134:
```suggestion
  if (raw != null) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  if (lowestOptional < 0) return;
+
+  for (let i = 0; i <= lowestOptional; i++) {
+    const n = chain[i];
+    if (n.type === "MemberExpression") {
+      n.type = "OptionalMemberExpression";
+    } else {
+      n.type = "OptionalCallExpression";
+    }
+    if (!n.optional) n.optional = false;
```
tido64, line 200:
```suggestion
    if (n.optional == null) n.optional = false;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+      chain.push(node);
+      if (node.optional) lowestOptional = chain.length - 1;
+      node = node.type === "MemberExpression" ? node.object : node.callee;
+    } else {
+      break;
+    }
+  }
+
+  if (lowestOptional < 0) return;
+
```
tido64: Nit: Technically, this isn't necessary as the for-loop already performs the same check
```suggestion
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  if (!expr) return;
+  markOptionalChain(expr);
+  inlineReplaceWithExpression(node);
+}
+
+function convertMethodDefinition(node: MutableNode): void {
+  const { key, value } = node;
+  if (!value) return;
+
+  const keyType = key && key.type;
```
tido64:
```suggestion
  const keyType = key?.type;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  node.async = value.async;
+  node.id = value.id ?? null;
+  delete node.value;
+  delete node.expression;
+
+  if (isPrivate) {
+    node.type = "ClassPrivateMethod";
+    if (keyType === "PrivateIdentifier") convertPrivateIdentifierToName(key);
+    if (node.kind === "get" || node.kind === "set") {
+      if (node.computed === undefined) node.computed = false;
```
tido64:
```suggestion
      if (node.computed == null) node.computed = false;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+    } else {
+      delete node.computed;
+    }
+  } else if (node.body === null) {
```
tido64, line 235:
```suggestion
  } else if (node.body == null) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+    node.type = "TSDeclareMethod";
+    if (node.computed === undefined) node.computed = false;
```
tido64:
```suggestion
    if (node.computed == null) node.computed = false;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  } else {
+    node.type = "ClassMethod";
+    if (node.computed === undefined) node.computed = false;
```
tido64:
```suggestion
    if (node.computed == null) node.computed = false;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+function convertPropertyDefinition(node: MutableNode): void {
+  const keyType = node.key && node.key.type;
+  if (keyType === "PrivateIdentifier" || keyType === "PrivateName") {
+    node.type = "ClassPrivateProperty";
+    if (keyType === "PrivateIdentifier")
+      convertPrivateIdentifierToName(node.key);
+    delete node.computed;
+  } else {
+    node.type = "ClassProperty";
+    if (node.computed === undefined) node.computed = false;
```
tido64:
```suggestion
    if (node.computed == null) node.computed = false;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  }
+}
+
+function convertTSAbstractPropertyDefinition(node: MutableNode): void {
+  node.type = "ClassProperty";
+  node.abstract = true;
+  if (node.computed === undefined) node.computed = false;
```
tido64:
```suggestion
  if (node.computed == null) node.computed = false;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+
+  if (i > 0) {
+    node.body = body.slice(i);
+  }
+}
+
+function convertProperty(node: MutableNode): void {
+  const isObjectMethod =
+    node.value &&
+    node.value.type === "FunctionExpression" &&
```
tido64:
```suggestion
    node.value?.type === "FunctionExpression" &&
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  if (!node.exportKind) node.exportKind = "value";
+}
+
+function convertExportAllWithExported(node: MutableNode): void {
+  if (!node.exported) return;
+
+  const exported = node.exported;
+  if (exported.type === "Literal" && typeof exported.value === "string") {
+    exported.type = "StringLiteral";
+    if (exported.raw !== undefined) {
```
tido64, line 396:
```suggestion
    if (exported.raw != null) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+      end: exported.end,
+    },
+  ];
+  node.declaration = null;
+  delete node.exported;
+}
+
+function convertTSInterfaceHeritage(node: MutableNode): void {
+  let expr = node.expression;
+  while (expr && expr.type === "MemberExpression") {
```
tido64:
```suggestion
  while (expr?.type === "MemberExpression") {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+      : undefined,
+  };
+  node.arguments = args;
+  delete node.source;
+  delete node.options;
+  delete node.phase;
+}
+
+function convertEnumDeclaration(node: MutableNode): void {
+  if (node.body && node.body.members && !node.members) {
```
tido64:
```suggestion
  if (node.body?.members && !node.members) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+
+// ── Universal cleanup (hot path — runs on every node) ───────────────
+
+/**
+ * Strip fields that OXC adds on many node types but Babel omits.
+ * Only truly universal checks belong here — type-specific cleanup
+ * is handled by the per-node-type ops table below.
+ */
+function cleanupOxcExtras(n: MutableNode): void {
+  if (n.decorators && n.decorators.length === 0) delete n.decorators;
```
tido64:
```suggestion
  if (n.decorators?.length === 0) delete n.decorators;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+};
+
+function applyOps(n: MutableNode, ops: NodeOps): void {
+  if (ops.type) n.type = ops.type;
+  if (ops.delete) {
+    for (const f of ops.delete) delete n[f];
+  }
+  if (ops.deleteIfEmpty) {
+    for (const f of ops.deleteIfEmpty) {
+      if (n[f] && n[f].length === 0) delete n[f];
```
tido64:
```suggestion
      if (n[f]?.length === 0) delete n[f];
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+    }
+  }
+  if (ops.deleteIfFalse) {
+    for (const f of ops.deleteIfFalse) {
+      if (n[f] === false) delete n[f];
+    }
+  }
+  if (ops.rename) {
+    for (const [from, to] of ops.rename) {
+      if (n[from] !== undefined && n[to] === undefined) {
```
tido64, line 541:
```suggestion
      if (n[from] != null && n[to] == null) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+        n[to] = n[from];
+        delete n[from];
+      }
+    }
+  }
+  if (ops.defaults) {
+    for (const [f, v] of ops.defaults) {
+      if (n[f] === undefined) n[f] = v;
```
tido64, line 549:
```suggestion
      if (n[f] == null) n[f] = v;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+      }
+    },
+  },
+
+  // ── Properties ──
+  Property: {
+    custom: (n) => {
+      const isMethodLike =
+        n.value &&
+        n.value.type === "FunctionExpression" &&
```
tido64:
```suggestion
        n.value?.type === "FunctionExpression" &&
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+
+  // ── Import/Export specifiers ──
+  ImportSpecifier: { custom: normalizeSpecifierImportKind },
+  ImportDefaultSpecifier: { custom: normalizeSpecifierImportKind },
+  ImportNamespaceSpecifier: { custom: normalizeSpecifierImportKind },
+  ExportSpecifier: {
+    custom: (n) => {
+      if (n.importKind === "value") delete n.importKind;
+      const exported = n.exported;
+      if (exported && exported.raw !== undefined) {
```
tido64, line 625:
```suggestion
      if (exported && exported.raw != null) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+
+  // ── Classes ──
+  ClassDeclaration: { deleteIfEmpty: ["implements"] },
+  ClassExpression: { deleteIfEmpty: ["implements"] },
+
+  // ── JSX ──
+  JSXOpeningFragment: { delete: ["attributes", "selfClosing"] },
+  JSXText: {
+    custom: (n) => {
+      if (n.raw !== undefined) {
```
tido64, line 665:
```suggestion
      if (n.raw != null) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  "leadingComments",
+  "trailingComments",
+  "innerComments",
+  "extra",
+  "directives",
+]);
+
+function isAstNode(val: unknown): val is MutableNode {
+  return (
+    val !== null &&
```
tido64:
```suggestion
    val != null &&
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  } else if (
+    type === "ClassMethod" ||
+    type === "ClassPrivateMethod" ||
+    type === "ObjectMethod" ||
+    type === "TSDeclareMethod"
+  ) {
+    firstChild = node.key;
+  }
+
+  if (!firstChild || !firstChild.leadingComments) return;
```
tido64:
```suggestion
  if (!firstChild?.leadingComments) return;
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+  // - Comments before `(` stay on id as trailing
+  // - Comments after `(` should be leading on first param
+  // First, remove from firstParam.leading any that are before `(`
+  if (firstParam.leadingComments) {
+    const keep: MutableNode[] = [];
+    for (const c of firstParam.leadingComments) {
+      if (c.start <= openParen) {
+        if (id) {
+          // Before `(` — trailing on id/callee/key
+          if (!id.trailingComments || !id.trailingComments.includes(c)) {
```
tido64:
```suggestion
          if (!id.trailingComments?.includes(c)) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+      } else {
+        keep.push(c);
+      }
+    }
+    if (keep.length > 0) firstParam.leadingComments = keep;
+    else delete firstParam.leadingComments;
+  }
+
+  // Move id trailing comments that are after `(` to first param leading
+  if (id && id.trailingComments) {
```
tido64:
```suggestion
  if (id?.trailingComments) {
```

incubator/tools-babel/src/estree.ts
```diff
@@ -0,0 +1,1640 @@
...
+    const keepTrailing: MutableNode[] = [];
+    for (const c of id.trailingComments) {
+      if (c.start > openParen) {
+        if (
+          !firstParam.leadingComments ||
+          !firstParam.leadingComments.includes(c)
+        ) {
```
tido64:
```suggestion
        if (!firstParam.leadingComments?.includes(c)) {
```

incubator/tools-babel/src/options.ts
```diff
@@ -0,0 +1,118 @@
...
+export function initTransformerContext<
+  T extends TransformerContext = TransformerContext,
+>(filename: string, settings: Partial<TransformerSettings>): T | undefined {
+  const {
+    parseFlowDefault = true,
+    parseFlowWorkspace = false,
+    parseExtAliases,
+    parseExtDefault = "js",
+  } = settings;
+  const isNodeModule = filename.includes("node_modules");
```
tido64: We should also include separators to avoid false positives
```suggestion
  const isNodeModule = /[/\\]node_modules[/\\]/.test(filename);
```

incubator/tools-babel/src/parse.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (errors.length > 0) {
+    if (!isFlowError(errors)) {
+      for (const e of errors) {
+        console.error(e.codeframe);
+      }
+      throw new Error(`Failed to parse '${filename}' because of above errors`);
+    }
+  } else {
+    const isTypeScript =
+      context.srcSyntax === "ts" || context.srcSyntax === "tsx";
```
tido64: We should move this before `oxcResult` so we don't have to check twice.

incubator/tools-babel/src/plugins.ts
```diff
@@ -0,0 +1,204 @@
...
+ */
+type ArrayUpdates<T> = Record<number, T | null>;
+
+function getUpdatedArray<T>(base: T[], updates: ArrayUpdates<T>): T[] {
+  if (Object.keys(updates).length === 0) {
+    return base;
+  }
+  const result: T[] = [];
+  for (let i = 0; i < base.length; i++) {
+    const entry = i in updates ? updates[i] : base[i];
```
tido64: Should we be using `Object.hasOwn` here to avoid issues related to inheritance?
```suggestion
    const entry = Object.hasOwn(updates, i) ? updates[i] : base[i];
```

incubator/tools-babel/src/types.ts
```diff
@@ -0,0 +1,145 @@
...
+  configCallerMixins?: Record<string, string>;
+
+  /** key values for plugins that should be disabled in the config */
+  configDisabledPlugins?: Set<string>;
+
+  /** disable the oxc parser */
+  parseDisableOxc?: boolean;
+
+  /** disable the hermes parser */
+  parseDisableHermes?: boolean;
```
tido64: These are awkwardly named, can we change them?
```suggestion
  /** disable the oxc parser */
  parseOxcDisabled?: boolean;

  /** disable the hermes parser */
  parseHermesDisabled?: boolean;
```
Or even invert the names?

package.json
```diff
@@ -106,7 +106,8 @@
           "src/index.{js,ts}"
         ],
         "project": [
-          "src/**/*.{js,ts,tsx}"
+          "src/**/*.{js,ts,tsx}",
+          "test/**/*.{js,ts,tsx}"
```
tido64: Is this one necessary? It seems to create more warnings:

Image

incubator/tools-babel/src/estree.ts
JasonVMo: I'm keeping the raws as some things actually use null, and it isn't quite equivalent.

incubator/tools-babel/src/estree.ts
JasonVMo: This would change the check from any falsy to only undefined/null.

incubator/tools-babel/src/estree.ts
JasonVMo: This is checking the null value specifically so best to not add undefined, which means something different.

incubator/tools-babel/src/estree.ts
JasonVMo: Raw is a field where null may have a specific meaning.

incubator/tools-babel/src/estree.ts
JasonVMo: This check is if the key isn't set which is undefined and not null.

incubator/tools-babel/src/estree.ts
JasonVMo: Checking key existence, intentional

tido64 (approved):

PR #4090, opened 2026-04-14
feat(tools-performance): Add subdomains and a metro perfLoggerFactory implementation

JasonVMo:
### Description

There are a few main parts to this change:

#### Subdomain Support

Nothing fancy here, subdomains are just domains that can be enabled by themselves or via the root key. Because the enabled check should be very fast, this just adds the subdomains directly to the enabled set. It also creates a mapping table that allows that to happen in any order. This also adds a `registerSubdomain` function to allow for callers to set up that association. Note that this will do nothing if perf tracking is not enabled. See the metro implementation for an example.

#### Metro `unstable_perfLoggerFactory` Implementation

This adds a create function to create a perf logger factory. It is mainly focused on the wider datasets (tracked by subdomains) and point to point tracking. This uses a new bit of functionality on `PerfDomains` for starting and finishing an event with a given name. This hooks into whatever strategy is being used (node or timing) and creates single timing entry for the two points.

The perf logger factory will return empty factories if it is not enabled.

#### Changed the packages from "module" to "commonjs"

While packages on recent versions of node can cleanly interop with ESM packages, given that metro.config.js files are typically CJS and aren't necessarily on the latest node, switching to CJS makes this easier. Making it explicitly "commonjs" rather than removing it removes warnings when running typescript test files directly about ambiguous module systems.


incubator/tools-performance/src/domain.ts
```diff
@@ -52,6 +52,21 @@ export class PerfDomain {
...
+   * @returns A function that stops the event when called
+   */
+  startEvent(tag: string, frequency?: EventFrequency): () => void {
+    if (!this.enabled(frequency)) {
+      return nullFunction;
+    }
+    const startVal = this.record(tag);
+    return () => {
+      this.record(tag, startVal);
+    };
```
tido64:
```suggestion
    return () => this.record(tag, startVal);
```

incubator/tools-performance/src/metro.ts
```diff
@@ -0,0 +1,89 @@
...
+} from "metro-config";
+import { type PerfDomain } from "./domain.ts";
+import { getDomain, registerSubdomain } from "./perf.ts";
+import { nullFunction } from "./trace.ts";
+
+function createEmptyLogger(): PerfLogger {
+  return {
+    point: nullFunction,
+    annotate: nullFunction,
+    subSpan: () => createEmptyLogger(),
```
tido64:
```suggestion
    subSpan: createEmptyLogger,
```

incubator/tools-performance/src/metro.ts
```diff
@@ -0,0 +1,89 @@
...
+
+function createLogger(subdomainName: string, domain?: PerfDomain): PerfLogger {
+  if (!domain) {
+    return createEmptyLogger();
+  }
+  const openEvents: Record<string, () => void> = {};
+  return {
+    point(name: string, _opts?: PerfLoggerPointOptions) {
+      if (name.endsWith("_start")) {
+        const eventKey = name.slice(0, -6);
```
tido64: Avoid magic numbers:
```suggestion
      const startSuffix = "_start";
      if (name.endsWith(startSuffix)) {
        const eventKey = name.slice(0, -startSuffix.length);
```

tido64: We should do the same for `"_end"`.

incubator/tools-performance/src/trace.ts
```diff
@@ -31,8 +31,9 @@ export function nullPassthrough<T>(value: T): T {
   return value;
 }
 
-/** no-op function matching the recordTime signature */
-export function nullRecordTime(_tag: string, _duration?: number): void {
+/** no-op function taking any number of parameters and doing nothing */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function nullFunction(..._args: any[]): void {
```
tido64: Shouldn't this just be `unknown`? We're not using it anyway so it really doesn't matter.
```suggestion
export function nullFunction(..._args: unknown[]): void {
```

incubator/tools-performance/src/tracker.ts
```diff
@@ -212,10 +232,10 @@ function styleName(name: string): string {
...
 
 const COL_OPTIONS: Record<PerfReportColumn, ColumnOptions> = {
-  name: { label: "operation", align: "left", style: styleName, maxWidth: 50 },
+  name: { label: "operation", align: "left", style: styleName, maxWidth: 70 },
   session: { label: "session", ...NUM_COL_OPTIONS },
   calls: { label: "calls", ...NUM_COL_OPTIONS },
-  total: { label: "total", ...NUM_COL_OPTIONS },
-  avg: { label: "avg", ...NUM_COL_OPTIONS },
+  total: { label: "total (ms)", ...NUM_COL_OPTIONS, digits: 1 },
+  avg: { label: "avg (ms)", ...NUM_COL_OPTIONS, digits: 1 },
```
tido64, line 239: What are `digits`? Number of zeros?

incubator/tools-performance/src/trace.ts
JasonVMo: Let me check to see whether it works with unknown. I have found for generic acceptance, that unknown ends up more restrictive. Given that this is usage unknown may work in practice.

incubator/tools-performance/src/tracker.ts
JasonVMo: Yes. It makes the difference between columns with ms output easier to distinguish vs. call/error count columns. Also, for timing values ms precision is too low, particularly for tiebreaking.

incubator/tools-performance/src/trace.ts
JasonVMo: Ok, appears to work in this usage. Switched it.

incubator/tools-performance/src/metro.ts
JasonVMo: switched to one time allocated constants for this.

tido64 (approved):

PR #4110, opened 2026-04-24
feat(tools-packages): Add package validation helpers to tools-packages

JasonVMo:
### Description

I started this code while doing some package linting work in the FURN repository where I wanted to shift some of the work to yarn constraints. It had what I thought was a fairly good pattern for enforcing rules on package.json files and realized this could be useful as general helpers for authoring rules against package.json files (and other JSON files).

This has a few main parts.

#### `PackageContext`

This is a raw bundle of readonly manifest, name, and root path. It is essentially the core of the PackageInfo type but the creation functions for it are uncached, whereas PackageInfo are cached and contain information about whether or not the package is a workspace.

#### `JSONValidator`

This can be created in fix mode or in check mode. By use of the `enforce` routine the manifest will be updated (if in fix) or emit errors (if in check). When the finish routine is called the manifest will be written out if there are changes. This can be created against any JSON file and is not package.json specific.

#### `PackageValidationContext`

This a a validating context which can be created against a folder/manifest in the repo, or from a `Yarn.Constraints.Workspace` allowing validation rules to be written such that they can run in `yarn constraints` calls or in other contexts. The dependency on `@yarnpkg/types` where this type comes from is an optional peer dependency as it only needs to be filled in when you are using constraints where you should already be typing your constraints via that package.

### Test plan

Added automated tests for the new functionality.

### DIFF
--- a/packages/tools-packages/README.md
+++ b/packages/tools-packages/README.md
@@ -3,23 +3,31 @@
 [![Build](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml/badge.svg)](https://github.com/microsoft/rnx-kit/actions/workflows/build.yml)
 [![npm version](https://img.shields.io/npm/v/@rnx-kit/tools-packages)](https://www.npmjs.com/package/@rnx-kit/tools-packages)
 
-This package has utilities for loading base information about packages,
-retrieved in a `PackageInfo` type, with a layer of caching that happens
-automatically, as well as the ability to store additional custom values in the
-retrieved `PackageInfo`
+This package provides utilities for working with npm packages from Node.js
+tooling. It covers three areas:
+
+- **Package info** — load and cache `PackageInfo` for a package by path or by
+  workspace name, with helpers for attaching custom data to the cached entry
+  via symbol-keyed accessors.
+- **Package contexts** — build a `PackageContext` from a root directory, and
+  optionally extend it into a `PackageValidationContext` that can validate or
+  fix the package's `package.json`.
+- **JSON validator** — a standalone `JSONValidator` for any JSON document that
+  reports differences (or applies them as fixes) at given paths, with optional
+  file-write-on-finish behavior. The same surface drives both standalone
+  validation and the Yarn constraints adapter.
 
 ## Motivation
 
-While loading package.json is pretty quick, this can quickly end up being a
-redundant operation as there different packages in rnx-kit all need different
-information from the file. This adds a simple caching layer for retrieving
-packages so work is not done multiple times.
+While loading `package.json` is pretty quick, it quickly becomes redundant when
+multiple tools in rnx-kit each need to read the same file. This package adds a
+caching layer so the work is done once.
 
-The packages can also have custom accessors defined that allow storing of
-additional data in the `PackageInfo` and because of that, associated with that
-package in the cache. This might be loading the `KitConfig` parsing and
-validating a tsconfig.json file. This package doesn't need to care what is being
-stored, other packages can add their custom accessors as needed.
+Tools also frequently need to enforce the same rules on `package.json` —
+sometimes as a CI check (report-only) and sometimes as a fix-up step. The
+`JSONValidator` API lets a single piece of validation code run in either mode,
+and the `createYarnWorkspaceContext` adapter lets the same code be wired into
+Yarn constraints without changes.
 
 ## Installation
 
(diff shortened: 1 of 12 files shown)


packages/tools-packages/src/json.ts
```diff
@@ -0,0 +1,370 @@
...
+) {
+  const parent = walkPath(path, context, true);
+  if (!parent) {
+    context.error(valueMessage(path, value, undefined));
+  } else {
+    const currentValue = parent[path[path.length - 1]];
+    if (!compareValues(currentValue, value)) {
+      if (context.fix) {
+        context.changed();
+        parent[path[path.length - 1]] = value;
```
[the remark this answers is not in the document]
JasonVMo: Fixed

packages/tools-packages/src/json.ts
```diff
@@ -0,0 +1,370 @@
...
+  ensureExists: boolean
+): Record<string, unknown> | undefined {
+  let current: Record<string, unknown> = context.json;
+  // walk to the second to last segment, the last one is the key we want to set/unset
+  for (let i = 0; i < path.length - 1; i++) {
+    const segment = path[i];
+    if (!isRecord(current[segment])) {
+      if (ensureExists && context.fix) {
+        context.changed();
+        current[segment] = {};
```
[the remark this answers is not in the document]
JasonVMo: Should be fixed now

packages/tools-packages/README.md
```diff
@@ -35,23 +43,114 @@ npm add --save-dev @rnx-kit/tools-packages
...
+```
+
+When `fix` is true and a `jsonFilePath` is set (the validation context infers
+this from the package root), `finish()` writes the updated `package.json` back
+to disk.
+
+### Yarn constraints
+
+The same validation code can run inside a Yarn constraints file by adapting
+the workspace object Yarn provides:
```
tido64, line 84: I'm not seeing the benefits of this when Yarn already solves the exact same problem that we've already implemented in rnx-kit. https://github.com/microsoft/rnx-kit/blob/18bf2c4c307826d8d075351d1b8aa93387566177/yarn.config.cjs#L16-L43

I think it would make more sense if the intention is to bring Yarn Constraints to other package managers or into other workloads, and that the syntax is basically the same.

JasonVMo: There are two questions here. The usefulness and the API. In terms of usefulness, yarn constraints are limited in applicability. There are a number of places where you want to run package validation logic, one of which is yarn constraints. This pattern allows you to author that code once and then have it be usable in multiple contexts. Align-deps is an example. A follow up PR here would be to have align-deps use this structure for package.json modifications, thus allowing it to be run via API in constraints if desired. Another point of the usefulness is that the yarn constraints model only works for package.json files. In FURN I use this same pattern for tsconfig.json validation as well.

The other question is API surface. The set/unset/error pattern is pretty clean, but I'm less a fan of using set as it is an extremely generic name and has different semantics than what actually happens. I chose the `undefined` to delete mainly because it allows better generic patterns (like loop handling) rather than having to branch the logic between deletes and adds.

tido64: That makes sense to me. It does sound like we want a new JSON validation package though. I don't think it fits in `tools-package` if it's supposed to be able to handle *any* `.json` file, not just `package.json`, which is the main purpose of this package.

packages/tools-packages/src/accessors.ts
```diff
@@ -11,34 +16,67 @@ import type { GetPackageValue, PackageInfo } from "./types.ts";
...
+  return createObjectValueAccessors<PackageContext, T>(friendlyName);
+}
+
+/**
+ * Create has/get/set accessors using a newly created symbol key that can look up values in any object
+ * in a way that is guaranteed to be unique and not collide with any other properties on the object.
+ * @param friendlyName name used to create a symbol key for the object
+ * @returns a set of accessors for the symbol key
+ */
+export function createObjectValueAccessors<TObj extends object, TVal>(
```
tido64, line 66: Can we use `Record<symbol, T>` instead of `object` here to avoid casting? I've managed to reduce it to 1 casting with this:
```diff
   friendlyName: string,
   initialize: (pkgInfo: PackageInfo) => T
 ): GetPackageValue<T> {
+  return createValueLoader<T, PackageInfo<T>>(friendlyName, initialize);
+}
+
+/**
+ * Helper function to create a typed accessor function for getting and storing information
+ * in any object. This can be whatever you want, the key is only created and stored in
+ * the generated function so there are no collisions.
+ *
+ * @param friendlyName name used to create a symbol key for the package info
+ * @param initialize function used to initialize the value stored in the key
+ * @returns a function to retrieve the value from the object, if unset the initialize function is called
+ */
+export function createValueLoader<T, TObj extends Record<symbol, T>>(
+  friendlyName: string,
+  initialize: (obj: TObj) => T
+): (obj: TObj) => T {
   const symbolKey = Symbol(friendlyName);
-  return (pkgInfo: PackageInfo) => {
-    if (!(symbolKey in pkgInfo)) {
-      pkgInfo[symbolKey] = initialize(pkgInfo);
+  return (obj: TObj) => {
+    if (!Object.hasOwn(obj, symbolKey)) {
+      (obj as Record<symbol, T>)[symbolKey] = initialize(obj);
     }
-    return pkgInfo[symbolKey] as T;
+    return obj[symbolKey];
   };
 }
```
And the following changes in `types.ts`:
```diff
diff --git a/packages/tools-packages/src/types.ts b/packages/tools-packages/src/types.ts
index cbcdb16cf..eb67240fd 100644
--- a/packages/tools-packages/src/types.ts
+++ b/packages/tools-packages/src/types.ts
@@ -1,6 +1,6 @@
 import type { PackageManifest } from "@rnx-kit/types-node";

-export type PackageInfo = {
+export type PackageInfo<T = unknown> = {
   /** name of the package */
   name: string;

@@ -18,13 +18,13 @@ export type PackageInfo = {
    * other packages to leverage any caching happening for PackageInfo entries to store additional
    * information and ensure it is only loaded once.
    */
-  [key: symbol]: unknown;
+  [key: symbol]: T;
 };

 /**
  * Typed accessors for retrieving values from the package info
  */
-export type GetPackageValue<T> = (pkgInfo: PackageInfo) => T;
+export type GetPackageValue<T> = (pkgInfo: PackageInfo<T>) => T;

 /**
  * Set of accessor functions that can be retrieved for a specific symbol
```
The final cast can't be removed because of TS2862 Type 'T' is generic and can only be indexed for reading. Which I guess is a limitation in TypeScript itself.

packages/tools-packages/src/types.ts
```diff
@@ -1,22 +1,105 @@
 import type { PackageManifest } from "@rnx-kit/types-node";
 
-export type PackageInfo = {
+/**
+ * Basic information about a package
+ */
+export type PackageContext<
```
tido64: Why is the name change necessary?

packages/tools-packages/src/types.ts
```diff
@@ -1,22 +1,105 @@
...
+ * Examples:
+ * - "dependencies.react" is equivalent to ["dependencies", "react"]
+ * - ["exports", ".", "import"] is ambiguous if using a string, but unambiguous as an array
+ */
+export type JSONValuePath = string | string[];
+
+/** JSON primitive types */
+export type JSONPrimitive = string | number | boolean | null;
+/** JSON value type */
+export type JSONValue = JSONPrimitive | Record<string, unknown> | JSONValue[];
```
tido64, line 33: We don't have to use `unknown` here. This is what we use in RNTA:
```ts
export type JSONValue =
  | string
  | number
  | boolean
  | JSONArray
  | JSONObject
  | null;

export type JSONArray = JSONValue[];
export type JSONObject = { [key: string | symbol]: JSONValue };
```
Note that we don't use `Record` because TypeScript doesn't like circular references.

packages/tools-packages/src/json.ts
tido64, line 1: This (along with the validation code) doesn't fit in `tools-package`. Consider moving to a separate package (not necessarily with a `tools-` prefix).

packages/tools-packages/src/json.ts
```diff
@@ -0,0 +1,393 @@
...
+  fix,
+  reportError,
+  reportPrefix,
+  jsonFilePath,
+}: JSONValidatorOptions): ResolvedOptions {
+  return {
+    fix: fix ?? defaultOptions.fix ?? false,
+    reportError: reportError ?? defaultOptions.reportError ?? console.error,
+    reportPrefix: reportPrefix ?? defaultOptions.reportPrefix,
+    jsonFilePath: jsonFilePath,
```
tido64, line 86:
```suggestion
    jsonFilePath,
```

packages/tools-packages/src/json.ts
```diff
@@ -0,0 +1,393 @@
...
+ */
+export const getJSONPathSegments = (() => {
+  const blocked = new Set(["__proto__", "constructor", "prototype"]);
+  return (path: JSONValuePath): string[] => {
+    const segments = Array.isArray(path) ? path : path.split(".");
+    for (const segment of segments) {
+      if (blocked.has(segment)) {
+        throw new Error(
+          `Blocked JSON path segment: ${segment} in "${segments.join(".")}"`
+        );
```
tido64, line 141: Can this be configured to delete/ignore instead of throwing similar to Node's [`--disable-proto`](https://nodejs.org/api/cli.html#disable-protomode)?

packages/tools-packages/src/json.ts
```diff
@@ -0,0 +1,393 @@
...
+  JSONValidator,
+  "error" | "changed" | "finish"
+> & {
+  json: Record<string, JSONValue>;
+  fix: boolean;
+};
+
+const defaultOptions: Omit<JSONValidatorOptions, "jsonFilePath"> = {
+  fix: false,
+};
```
tido64, line 59: It looks like we already handle `fix` being unset so we can make this smaller:
```suggestion
const defaultOptions: Omit<JSONValidatorOptions, "jsonFilePath"> = {};
```

JasonVMo:
Splitting this into two PRs. The first one is: https://github.com/microsoft/rnx-kit/pull/4124

PR #4115, opened 2026-04-29
feat(tools-formatting): Add a tree formatter to tools-formatting

JasonVMo:
### Description

This adds a simple console tree formatter to the package which outputs a header and a set of sub-messages associated with that header.

It also generalizes some of the ascii art patterns to have unicode/ascii defaults (for both tables and trees) as well as allowing direct customization.

### Test plan

Tests updated as part of this change.

### DIFF
--- a/incubator/tools-formatting/src/const.ts
+++ b/incubator/tools-formatting/src/const.ts
@@ -1,3 +1,35 @@
+import type { TableViewParts, TreeViewParts } from "./types.ts";
+
 export const ELLIPSIS = "...";
 export const SRC_DIRS = ["src", "lib", "dist", "bin"];
 export const SEPARATORS = ["/", "\\"];
+
+type StyleKeys = "default" | "ascii";
+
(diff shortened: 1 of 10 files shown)


incubator/tools-formatting/src/const.ts
```diff
@@ -1,3 +1,35 @@
...
+    fill: "─",
+  },
+  ascii: {
+    top: ["+", "+", "+"],
+    mid: ["+", "+", "+"],
+    bottom: ["+", "+", "+"],
+    row: ["|", "|", "|"],
+    fill: "-",
+  },
+};
```
tido64, line 35: Nit: I think these could be defined as an ordered flat array but this is also fine.

tido64 (approved):

incubator/tools-formatting/src/const.ts
JasonVMo: I may look at that in a separate commit. I'm worried about having 9 or 10 entries which feels a bit unwieldy to fill in.

PR #4121, opened 2026-05-04
fix(third-party-notices): ignore submodule manifests

### LINKED ISSUE #4118: `@rnx-kit/third-party-notices` crashes on subpath `package.json` markers (Yarn 4 pnpm linker, dual-format ESM/CJS packages)
## Package & version

`@rnx-kit/third-party-notices@2.0.2` (latest at time of filing).

## Repro

- Yarn 4 with `nodeLinker: pnpm` (workspace layout: `node_modules/.store/<pkg>/package/...`).
- A workspace whose dependency graph includes modern dual-format ESM/CJS packages — e.g. `socket.io-client@4.x`, `@socket.io/component-emitter`
- Run a Metro bundle that invokes `third-party-notices` (e.g. via `@office-iss/sdx-build-tools`).

## Root cause

Modern dual-format packages ship subpath `package.json` markers to flag a subdirectory's module format, e.g.:

```json
// .../socket.io-client/build/esm/package.json
{ "name": "socket.io-client", "version": "4.8.3", "type": "module" }
```

These are **not** package boundaries — they lack `license`, `dependencies`, etc. Under the Yarn 4 pnpm linker, source paths resolved by Metro often point *into* such subdirectories, so `splitSourcePath` in `lib/write-third-party-notices.js` calls `findPackageDir(p)` and lands on the marker rather than the real package root.

## Three crash sites (all from the same root cause)

1. **`lib/write-third-party-notices.js:165`** — `parseModule` does `moduleName.startsWith(...)`, but `moduleName` is `undefined` when the marker has no `name`.
2. **`lib/extractors.js:45`** — `module.version.match(...)` throws because the marker has no `version`.
3. **`lib/output/text.js:17`** — `throw new Error(\"No license information found...\")` fires because the marker has no `license`, even though the real package root one directory up does.

## Proposed fix

In `splitSourcePath`, walk up past subpath markers when the discovered `package.json` lacks license/dependencies (heuristic for \"this is a subpath marker, not a real package boundary\"):

```js
let pkgRoot = findPackageDir(p);
while (pkgRoot) {
  const manifest = readPackage(pkgRoot);
  if (manifest && (manifest.license || manifest.licenses ||
      manifest.dependencies || manifest.peerDependencies)) {
    return [manifest.name, pkgRoot];
  }
  const parent = path.dirname(pkgRoot);
  if (parent === pkgRoot) break;
  const next = findPackageDir(parent);
  if (!next || next === pkgRoot) break;
  pkgRoot = next;
}
```

Validated locally in our internal monorepo — with the walk-up, the bundle build passes cleanly with no `ignoreModules` workarounds; affected packages resolve to their real roots and emit correct license entries.

Filing as an issue (not a PR) so maintainers can choose their preferred approach — the snippet above is just a starting point.

tido64:
### Description

Fixed submodule manifests being picked up as the package manifest.

Resolves #4118.

### Test plan

See #4118 for repro.

### DIFF
--- a/packages/third-party-notices/src/write-third-party-notices.ts
+++ b/packages/third-party-notices/src/write-third-party-notices.ts
@@ -108,6 +108,18 @@ export async function getCurrentPackageId(
   return undefined;
 }
 
+function isPackageManifest(manifest: ReturnType<typeof readPackage>): boolean {
+  return Boolean(
+    manifest.license ||
+    manifest.licenses ||
+    manifest.files ||
+    manifest.exports ||
+    manifest.dependencies ||
+    manifest.peerDependencies ||
+    manifest.devDependencies
+  );
+}
+
 // helper functions
 export function normalizePath(p: string, currentPackageId?: string): string {
   let result = p.replaceAll("webpack:///", "");
@@ -159,10 +171,20 @@ export function splitSourcePath(rootPath: string, p: string): string[] {
 
   // This is most likely an absolute file path e.g.,
   // `/~/node_modules/.store/react-native-virtual-3e97acc5aa/package/index.js`
-  const pkgRoot = findPackageDir(p);
-  if (pkgRoot) {
+  let pkgRoot = findPackageDir(p);
+  while (pkgRoot) {
     const manifest = readPackage(pkgRoot);
-    return [manifest.name, pkgRoot];
+    // Retry if this package manifest seems to be a submodule manifest
+    if (isPackageManifest(manifest)) {
+      return [manifest.name, pkgRoot];
+    }
+
+    const parentDir = path.dirname(pkgRoot);
+    if (parentDir === pkgRoot) {
+      break;
+    }
+
+    pkgRoot = findPackageDir(parentDir);
   }
 
   throw new Error(`Could not parse module: ${p}`);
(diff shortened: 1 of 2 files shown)


packages/third-party-notices/src/write-third-party-notices.ts
```diff
@@ -159,10 +171,20 @@ export function splitSourcePath(rootPath: string, p: string): string[] {
 
   // This is most likely an absolute file path e.g.,
   // `/~/node_modules/.store/react-native-virtual-3e97acc5aa/package/index.js`
-  const pkgRoot = findPackageDir(p);
-  if (pkgRoot) {
+  let pkgRoot = findPackageDir(p);
+  while (pkgRoot) {
```
JasonVMo, line 175: Should there be some tracking of the last one found in the failure case? If the isPackageManifest check above fails for some reason it will skip over that manifest, then end up walking to the root case and throw an exception.

JasonVMo: This does seem more theoretical than practical so it is a bit of a nit.

tido64: Because we couldn't find a valid package manifest, I think it's correct to throw. But maybe we should include all the paths we looked at? That would make it easier to debug.

Saadnajmi (approved):

packages/third-party-notices/src/write-third-party-notices.ts
tido64: Resolving for now and addressing this later.

PR #4124, opened 2026-05-05
feat(lint-json): Move the JSON validation code to its own package

JasonVMo:
### Description

This creates a new package for json linting with the validation code from the tools-packages PR. This should also address the PR feedback in the other PR.

The proto blocking I left as an exception as that is an error in the test code not a random user error and I prefer it to silent failures.

### Test plan

Automated tests


.changeset/eager-parents-battle.md
tido64, line 1: Are we not publishing this package yet?

incubator/lint-json/src/validator.ts
```diff
@@ -0,0 +1,175 @@
...
+function setValue(
+  path: string[],
+  value: JSONValue,
+  context: JSONEditingContext
+) {
+  const parent = walkPath(path, context, true);
+  if (!parent) {
+    context.error(valueMessage(path, value, undefined));
+  } else {
+    const currentValue = parent[path[path.length - 1]];
```
tido64: The current index can be stored for reuse further down
```suggestion
    const index = path[path.length - 1];
    const currentValue = parent[index];
```

incubator/lint-json/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "directory": "incubator/lint-json"
+  },
+  "files": [
+    "lib/**/*.d.ts",
+    "lib/**/*.js"
+  ],
+  "type": "module",
+  "sideEffects": false,
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
```
tido64, line 23: Can you also provide an exports map?

tido64 (approved):
In a future iteration, we should provide a way to say "one of these fields must be set". This is required to replace our `tsconfig.json` validator.

incubator/lint-json/src/validator.ts
```diff
@@ -0,0 +1,175 @@
...
+ * and finalize the validation process, optionally writing changes back to disk if fixes
+ * are enabled and a file path is provided.
+ * @param jsonPath the path to the JSON file being validated; used for default error headers and writing on fix
+ * @param json loaded JSON object to validate; if not provided the file at `jsonPath` will be read from disk
+ * @param userOptions options controlling fix behavior, error reporting, and file writing
+ * @returns a JSONValidator instance for validating and optionally fixing the provided JSON object
+ */
+export function createJSONValidator(
+  jsonPath: string,
+  json: Record<string, JSONValue> | undefined,
```
tido64: Can we replace all instances of `Record<string, JSONValue>`?
```suggestion
  json: JSONObject | undefined,
```

tido64 (approved):

PR #4128, opened 2026-05-06
feat(lint-package): Create @rnx-kit/lint-package

JasonVMo:
### Description

This creates a new experimental package @rnx-kit/lint-package with a `PackageValidationContext` class that can be initialized via a path or via a `Yarn.Constraints.Workspace` for running in constraints mode.

This brings together a bunch of core functionality for validating package configuration in a single place. A framework for adding and running rules will come in the future for lint-style operations, and this new class will be the context those rules will run against.

In the meantime this is still a useful rollup for various configuration or linting type scripts and covers some use cases present in both FURN and OMR. 

Other notes:

- The class is exposed directly so that it can be extended if so desired.
- The manifest type can be overridden for repo-specific extensions.


.changeset/real-crabs-enter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rnx-kit/lint-json": patch
```
tido64: Should this be minor since you've made a breaking change to `dirty()`?

incubator/lint-package/test/context.test.ts
```diff
@@ -0,0 +1,395 @@
...
+import { PackageValidationContext } from "../src/context.ts";
+
+function makeTempPackage(
+  manifest: PackageManifest,
+  files: string[] = []
+): {
+  root: string;
+  cleanup: () => void;
+} {
+  const root = fs.mkdtempSync(path.join(os.tmpdir(), "lint-package-"));
```
tido64: I don't suppose we can use memfs instead of writing to disk?

yarn.lock
```diff
@@ -15975,16 +15995,17 @@ __metadata:
   linkType: hard
 
 "react-native-test-app@npm:^5.0.9, react-native-test-app@npm:^5.1.2":
-  version: 5.1.4
-  resolution: "react-native-test-app@npm:5.1.4"
+  version: 5.1.2
```
tido64: I think you need to reset and regenerate this file. There's a bunch of downgrades in here.

incubator/lint-package/src/context.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+  get kitConfig(): KitConfig {
+    return (this._kitConfig ??=
+      getKitConfigFromPackageManifest(this.manifest, this.root) ?? {});
+  }
+
+  /**
+   * protected and private properties and helpers
+   */
+  protected static JS_CONFIG_EXTENSIONS = [".js", ".cjs", ".mjs", ".ts"];
```
tido64: `.cts` and `.mts` are also valid extensions

incubator/lint-package/src/context.ts
```diff
@@ -0,0 +1,214 @@
...
+  protected readonly _validator: JSONValidator;
+  protected _delegates?: Record<string, JSONValidator | null>;
+  private _fileExists?: Record<string, boolean>;
+  private _attached?: Record<symbol, unknown>;
+  private _kitConfig?: KitConfig;
```
tido64, line 114: What is this used for?

JasonVMo: Loading the kitConfig, including extends resolution. It's cached but will be reloaded if it has changed locally. (cleared via the dirty function changing the rnx-kit key.

incubator/lint-package/test/context.test.ts
JasonVMo: I'll look at piping through the tools-filesystem pattern for both packages.

tido64 (approved):

PR #4144, opened 2026-05-10
feat(react-native-host): extend RNXHostConfig with optional host hooks (1/3)

Saadnajmi:
### Description

Adds three optional lifecycle hooks to `RNXHostConfig`. All three are opt-in via `respondsToSelector:` — consumers that don't implement them see no behavior change.

#### `host:didLoadInstanceWithError:` and `hostWillUnloadInstance:`

`ReactNativeHost` subscribes to `RCTJavaScriptDidLoadNotification`, `RCTJavaScriptDidFailToLoadNotification`, and `RCTBridgeWillBeInvalidatedNotification` when the config implements the
 corresponding selectors, and forwards. `-dealloc` removes the observers.

Works in both bridge and bridgeless modes — the same notifications fire in both.

#### `host:didInitializeRuntime:` (Objective-C++ only)

Fires inside the bridgeless `RCTHost` runtime-initialization lambda, after host bindings install (TurboModules, native logger, native component registry) but **before** the user JS
bundle loads. Useful for evaluating pre-user JS on the runtime — e.g., bundles that must be available before app code runs via `runtime.evaluateJavaScript`.

Bridgeless mode only. Wired via an internal `_RNXForwardingRCTHostDelegate` passed as `RCTHost`'s `hostDelegate` (was `nil`). Retained as an ivar because `RCTHost` stores host delegates
 weakly. C++-typed (`jsi::Runtime &`), so only visible to Objective-C++ consumers.

### Test plan

Tested internally

### DIFF
--- a/packages/react-native-host/cocoa/RNXHostConfig.h
+++ b/packages/react-native-host/cocoa/RNXHostConfig.h
@@ -3,6 +3,12 @@
 #import <React/RCTBridgeDelegate.h>
 #import <React/RCTLog.h>
 
+#ifdef __cplusplus
+#include <jsi/jsi.h>
+#endif
+
+@class ReactNativeHost;
+
 NS_ASSUME_NONNULL_BEGIN
 
 /// Configuration object for ``ReactNativeHost``.
@@ -29,6 +35,20 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Handles a fatal error.
 - (void)onFatalError:(NSError *)error;
+
+/// Called after the JS instance has finished loading. ``error`` is ``nil``
+/// on success.
+- (void)host:(ReactNativeHost *)host didLoadInstanceWithError:(nullable NSError *)error
+    __attribute__((__swift_name__("host(_:didLoadInstanceWithError:)")));
+
+/// Called when the instance is about to be unloaded.
+- (void)hostWillUnloadInstance:(ReactNativeHost *)host;
+
+#ifdef __cplusplus
+/// Called after host bindings install but before the user JS bundle loads.
+/// Use to evaluate pre-user JS on the runtime. Bridgeless mode only.
+- (void)host:(ReactNativeHost *)host didInitializeRuntime:(facebook::jsi::Runtime &)runtime;
+#endif  // __cplusplus
 
 // MARK: - RCTBridgeDelegate deprecated details (for backwards compatibility) [>=0.84]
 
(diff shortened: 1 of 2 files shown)


packages/react-native-host/cocoa/RNXHostConfig.h
```diff
@@ -30,6 +36,20 @@ NS_ASSUME_NONNULL_BEGIN
 /// Handles a fatal error.
 - (void)onFatalError:(NSError *)error;
 
+/// Called after the JS instance has finished loading. ``error`` is ``nil``
+/// on success.
+- (void)host:(ReactNativeHost *)host didLoadInstanceWithError:(nullable NSError *)error
```
Saadnajmi, line 41: I'm a little iffy with the optional error parameter.. the signature makes it feel like there is always an error despite the comment. Alternative is to also expose `failedtoLoadInstanceWithError`. Thoughts?

tido64: Shouldn't this be `- (void)didLoadInstanceWithHost:(ReactNativeHost *)host error: (nullable NSError **)error`, where the user passes a ref to an NSError if they're interested.

packages/react-native-host/cocoa/ReactNativeHost.mm
```diff
@@ -35,6 +35,41 @@ @interface ReactNativeHost () <RCTComponentViewFactoryComponentProvider>
 @end
 #endif  // USE_CODEGEN_PROVIDER
 
+#if USE_BRIDGELESS
```
tido64, line 38: Should this block be in a private header instead?

packages/react-native-host/cocoa/ReactNativeHost.mm
```diff
@@ -99,11 +137,63 @@ - (instancetype)initWithConfig:(id<RNXHostConfig>)config launchOptions:(NSDictio
...
+                       name:RCTJavaScriptDidFailToLoadNotification
+                     object:nil];
+        }
+        if ([config respondsToSelector:@selector(hostWillUnloadInstance:)]) {
+            [[NSNotificationCenter defaultCenter]
+                addObserver:self
+                   selector:@selector(_rnxInstanceWillUnload:)
+                       name:RCTBridgeWillBeInvalidatedNotification
+                     object:nil];
+        }
```
tido64, line 158: Shouldn't all of this be contained within `_RNXForwardingRCTHostDelegate`?

packages/react-native-host/cocoa/RNXHostConfig.h
```diff
@@ -30,6 +36,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+    __attribute__((__swift_name__("host(_:didLoadInstanceWithError:)")));
+
+/// Called when the instance is about to be unloaded.
+- (void)hostWillUnloadInstance:(ReactNativeHost *)host;
+
+#ifdef __cplusplus
```
tido64, line 47: Is this allowed? Won't we end up with two definitions of this protocol?

tido64:

PR #4145, opened 2026-05-11
feat(react-native-host): expose auxiliary RCTTurboModuleManagerDelegate via RNXHostConfig (2/3)

Saadnajmi:
### Description

Depends on #4144 

Adds an optional `turboModuleManagerDelegate` property to `RNXHostConfig` that lets consumers participate in TurboModule construction without replacing `RNXTurboModuleAdapter`.

#### Motivation

`RNXTurboModuleAdapter` is rnx-kit's `RCTTurboModuleManagerDelegate`. Today it owns the full TurboModule construction path:

- `getModuleClassFromName:` falls back to `RCTCoreModulesClassProvider`
- `getModuleInstanceFromClass:` falls back to `RCTAppSetupDefaultModuleFromClass`

For modules that need *per-host state* — e.g., an `RCTExceptionsManager` initialized with a per-host delegate, or an `RCTDevSettings` with a custom data source — there's no extension
point. In bridge mode this case is handled via `RCTBridgeDelegate.extraModulesForBridge:` (the bridge prefers pre-constructed instances vended there). In bridgeless mode, that hook
isn't called — modules go through `RCTTurboModuleManager.delegate` instead.

#### Design

`RNXHostConfig` adds:

```objc
@property (nonatomic, readonly, weak, nullable) id<RCTTurboModuleManagerDelegate> turboModuleManagerDelegate;
```

`RNXTurboModuleAdapter` adds a `hostConfig` weak reference and, for each `RCTTurboModuleManagerDelegate` method it implements, consults the auxiliary via `respondsToSelector:` first;
returning `nil` / `Nil` falls through to the existing default behavior.

```objc
- (Class)getModuleClassFromName:(char const *)name {
    id<RCTTurboModuleManagerDelegate> aux = _hostConfig.turboModuleManagerDelegate;
    if ([aux respondsToSelector:_cmd]) {
        Class cls = [aux getModuleClassFromName:name];
        if (cls != Nil) return cls;
    }
    return RCTCoreModulesClassProvider(name);
}
```

Same shape for `getModuleInstanceFromClass:`.

#### Properties

- **Symmetric with bridge-mode `extraModulesForBridge:`**: the per-host injection point that bridge consumers use.
- **Backward-compatible**: consumers that don't implement `turboModuleManagerDelegate` see no behavior change.
- **Generic, not per-method**: any current or future `RCTTurboModuleManagerDelegate` method (`getModuleProvider:`, `getTurboModule:jsInvoker:`, etc.) the consumer implements is routed
automatically — no per-hook plumbing required in rnx-kit.
- **Zero overhead** when the auxiliary delegate doesn't respond — single `respondsToSelector:` check.

#### Example

```objc
@interface MyHostConfig : NSObject <RNXHostConfig, RCTTurboModuleManagerDelegate>
@end

@implementation MyHostConfig {
    MyExceptionsManagerDelegate *_exceptionsDelegate;
}

- (id<RCTTurboModuleManagerDelegate>)turboModuleManagerDelegate {
    return self;
}

// Required RCTTurboModuleManagerDelegate methods — return nil to defer to rnx-kit's defaults.
- (Class)getModuleClassFromName:(const char *)name { return Nil; }

- (id<RCTTurboModule>)getModuleInstanceFromClass:(Class)cls {
    if (cls == [RCTExceptionsManager class]) {
        return [[RCTExceptionsManager alloc] initWithDelegate:_exceptionsDelegate];
    }
    return nil;
}
@end
```

### Test plan

Tested internally

### DIFF
--- a/packages/react-native-host/cocoa/RNXHostConfig.h
+++ b/packages/react-native-host/cocoa/RNXHostConfig.h
@@ -7,6 +7,10 @@
 #include <jsi/jsi.h>
 #endif
 
+#if __has_include(<ReactCommon/RCTTurboModuleManager.h>)
+#import <ReactCommon/RCTTurboModuleManager.h>
+#endif
+
 @class ReactNativeHost;
 
 NS_ASSUME_NONNULL_BEGIN
@@ -24,6 +28,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Returns whether the bridge should be released when the app is backgrounded.
 @property (nonatomic, readonly) BOOL shouldReleaseBridgeWhenBackgrounded;
+
+#if __has_include(<ReactCommon/RCTTurboModuleManager.h>)
+/// Auxiliary ``RCTTurboModuleManagerDelegate`` consulted before
+/// ``RNXTurboModuleAdapter``'s defaults. For each delegate method the
+/// adapter implements, the auxiliary is called via ``respondsToSelector:``
+/// first; returning ``nil`` / ``Nil`` falls through to default behavior.
+@property (nonatomic, readonly, weak, nullable) id<RCTTurboModuleManagerDelegate> turboModuleManagerDelegate;
+#endif
 
 /// Logs a message.
 - (void)logWithLevel:(RCTLogLevel)level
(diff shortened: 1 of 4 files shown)


tido64:

PR #4146, opened 2026-05-11
feat(react-native-host): allow opting out of React-RCTAppDelegate dependency (3/3)

Saadnajmi:
### Description

Depends on #4144
Depends on #4145

Adds `RNX_USE_RCT_APP_DELEGATE_HELPERS` (default `1`, preserving current behavior). Set to `0` to drop the React-RCTAppDelegate / ReactAppDependencyProvider framework dependency from
`RNXTurboModuleAdapter`.

#### Motivation

`React-RCTAppDelegate`'s `RCTAppSetupDefaultModuleFromClass` and `RCTAppSetupDefaultJsExecutorFactory` helpers assume a single-host-per-app shape — process-wide module providers and
executor factories. Multi-host integrations (where each host has its own modules and lifecycle) want neither. Today the only way to use `RNXTurboModuleAdapter` is to link against
React-RCTAppDelegate even if it's actively the wrong design for the consumer.

#### Design

When `RNX_USE_RCT_APP_DELEGATE_HELPERS=0`:

- `getModuleInstanceFromClass:` skips `RCTAppSetupDefaultModuleFromClass` and returns `nil`. Consumers vending custom-initialized modules do so via
`RNXHostConfig.turboModuleManagerDelegate.getModuleInstanceFromClass:` (added in the previous PR). Falling through to `nil` makes `RCTTurboModuleManager` use its default `[cls new]` for
 modules the consumer doesn't override — same behavior the manager has for unknown classes.

- `jsExecutorFactoryForBridge:` returns `nullptr`. Bridgeless integrations (`RCTHost`-driven) construct the JS runtime via `RCTHostJSEngineProvider` and never reach this path.
Bridge-mode consumers wanting this adapter must either keep `HELPERS=1` or supply their own `RCTBridgeDelegate.jsExecutorFactoryForBridge:`.

The macro auto-defaults to `1` when the React-RCTAppDelegate headers are reachable (via `__has_include`), so behavior is unchanged for existing consumers. Multi-host consumers `#define
RNX_USE_RCT_APP_DELEGATE_HELPERS 0` in their build settings to opt out.

#### Why this is useful

Without the previous PR in this stack (the auxiliary `turboModuleManagerDelegate` hook), opting out of the helpers leaves the consumer with nil-returning defaults and no escape hatch —
pointless. With the auxiliary delegate in place, consumers have a real alternative: implement `getModuleInstanceFromClass:` themselves to vend whatever modules they need, and
`RNX_USE_RCT_APP_DELEGATE_HELPERS=0` drops the framework dependency cleanly.

### Test plan

Tested internally

### DIFF
--- a/packages/react-native-host/cocoa/RNXHostConfig.h
+++ b/packages/react-native-host/cocoa/RNXHostConfig.h
@@ -7,6 +7,10 @@
 #include <jsi/jsi.h>
 #endif
 
+#if __has_include(<ReactCommon/RCTTurboModuleManager.h>)
+#import <ReactCommon/RCTTurboModuleManager.h>
+#endif
+
 @class ReactNativeHost;
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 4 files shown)


tido64:

PR #4150, opened 2026-05-13
Add CI log grouping for align-deps output

### LINKED ISSUE #2906: align-deps: add support for grouping log lines on CI
- GitHub Actions: https://github.com/actions/toolkit/blob/main/docs/commands.md#group-and-ungroup-log-lines
- Azure DevOps: ?

Example: https://github.com/yarnpkg/berry/blob/a64075561a6476aa79d0fa1012ecf6b1633a88f2/packages/yarnpkg-core/sources/StreamReport.ts#L31-L37

vivekjm:
## Summary

Adds CI log grouping around `align-deps` output so runs over larger workspaces are easier to scan in GitHub Actions and Azure DevOps.

This keeps local output unchanged and only opens a group when a package actually writes log output, avoiding empty groups for packages with no changes.

Fixes #2906.

## Testing

- `git diff --check`
- `yarn workspace @rnx-kit/align-deps format`
- `PATH="/Users/vivek/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" yarn workspace @rnx-kit/align-deps lint`
- `PATH="/Users/vivek/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" yarn workspace @rnx-kit/align-deps build --dependencies`
- `PATH="/Users/vivek/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" yarn workspace @rnx-kit/align-deps test`

### DIFF
--- a/packages/align-deps/src/ci.ts
+++ b/packages/align-deps/src/ci.ts
@@ -1,34 +1,79 @@
-type LogGroup = {
-  end: string;
-  start: string;
+type GroupMarkers = {
+  beginGroup: (title: string) => void;
+  endGroup: () => void;
 };
 
-type Env = NodeJS.ProcessEnv;
+type Output = Pick<typeof console, "error" | "log" | "warn">;
 
-function escapeGitHubCommand(value: string): string {
-  return value
-    .replace(/%/g, "%25")
-    .replace(/\r/g, "%0D")
-    .replace(/\n/g, "%0A");
+function noop() {
+  return undefined;
 }
 
-export function logGroupFor(
-  title: string,
-  env: Env = process.env
-): LogGroup | undefined {
+export function getGroupMarkers(
+  log: typeof console.log = console.log,
+  env = process.env
+): GroupMarkers {
   if (env["GITHUB_ACTIONS"] === "true") {
     return {
-      start: `::group::${escapeGitHubCommand(title)}`,
-      end: "::endgroup::",
+      beginGroup: (title) => log(`::group::${encodeURI(title)}`),
+      endGroup: () => log("::endgroup::"),
     };
   }
 
   if (env["TF_BUILD"] === "True") {
     return {
-      start: `##[group]${title}`,
-      end: "##[endgroup]",
+      beginGroup: (title) => log(`##[group]${title}`),
+      endGroup: () => log("##[endgroup]"),
     };
   }
 
-  return undefined;
+  return {
+    beginGroup: noop,
+    endGroup: noop,
+  };
+}
+
+export function withLogGroup<T>(
+  title: string,
+  fn: () => T,
+  output: Output = console,
+  env = process.env
+): T {
+  const originalLog = output.log;
+  const originalWarn = output.warn;
+  const originalError = output.error;
+  const { beginGroup, endGroup } = getGroupMarkers(originalLog, env);
+  let groupStarted = false;
+
+  const begin = () => {
+    if (!groupStarted) {
+      beginGroup(title);
+      groupStarted = true;
+    }
+  };
+
+  output.log = (...args) => {
+    begin();
+    originalLog(...args);
+  };
+  output.warn = (...args) => {
+    begin();
+    originalWarn(...args);
+  };
+  output.error = (...args) => {
+    begin();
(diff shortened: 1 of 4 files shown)


packages/align-deps/src/ci.ts
```diff
@@ -0,0 +1,34 @@
...
+function escapeGitHubCommand(value: string): string {
+  return value
+    .replace(/%/g, "%25")
+    .replace(/\r/g, "%0D")
+    .replace(/\n/g, "%0A");
+}
+
+export function logGroupFor(
+  title: string,
+  env: Env = process.env
```
tido64: Nit: Type can be inferred
```suggestion
  env = process.env
```

packages/align-deps/src/ci.ts
```diff
@@ -0,0 +1,34 @@
...
+): LogGroup | undefined {
+  if (env["GITHUB_ACTIONS"] === "true") {
+    return {
+      start: `::group::${escapeGitHubCommand(title)}`,
```
tido64: We should use `encodeURI` instead:
```suggestion
      start: `::group::${encodeURI(title)}`,
```

packages/align-deps/src/cli.ts
```diff
@@ -266,7 +267,11 @@ export async function cli({ packages, ...args }: Args): Promise<void> {
   // disk only when everything is in order for the target package. Packages with
   // invalid or missing configurations are skipped.
   const errors = manifests.reduce((errors, manifest) => {
+    const logGroup = logGroupFor(manifest);
```
tido64: Can we always return start/end functions instead?
```js
  const { beginGroup, endGroup } = getGroupMarkers();
  const errors = manifests.reduce((errors, manifest) => {
    try {
      beginGroup(manifest);
      ...
    } finally {
      endGroup(manifest);
    }
```

packages/align-deps/src/cli.ts
```diff
@@ -266,7 +267,11 @@ export async function cli({ packages, ...args }: Args): Promise<void> {
...
     try {
+      if (logGroup) {
+        console.log(logGroup.start);
+      }
```
tido64: I don't think this is correct. This will *always* start a group even though we don't know if anything will be logged yet. When no changes are needed, nothing should be logged.

tido64 (commented):
Thanks for your contribution! I've left some comments.

packages/align-deps/src/cli.ts
vivekjm: Agreed. I changed the implementation so the group starts lazily on the first log/warn/error for that manifest. If nothing is logged, no group markers are emitted.

packages/align-deps/src/ci.ts
vivekjm: Good call, thanks. I removed the explicit env type and let TypeScript infer it from the default value.

packages/align-deps/src/cli.ts
vivekjm: Updated this to expose beginGroup/endGroup through getGroupMarkers. I also moved the call site behind a small helper so the CLI loop stays readable.

packages/align-deps/src/ci.ts
vivekjm: Done. I switched the GitHub Actions group title handling to use encodeURI here.

packages/align-deps/src/ci.ts
```diff
@@ -0,0 +1,79 @@
...
+    };
+  }
+
+  return {
+    beginGroup: noop,
+    endGroup: noop,
+  };
+}
+
+export function withLogGroup<T>(
```
tido64, line 36: I think the `withLogGroup` pattern is a good one. However, I don't think that we should be overwriting `global.console` like this. I created a draft of what I had in mind here: #4156

If you don't mind, we can close this one in favour of that one. I'll give you full credit for it, of course.

vivekjm: yes please !! feel free to close it

tido64:
Superseded by #4156

PR #4153, opened 2026-05-14
feat(tools-formatting): Support github and azure output formatting for annotations and groups

JasonVMo:
### Description

This adds functionality in tools-formatting to support github actions and azure pipelines console output.

#### Reporters
It does this by introducing the concept of reporters, similar to many other tools (like testing frameworks), which controls how certain types of output is formatted.
- four built-in reporter types are supported: azure, github, console, and file
- console and file are identical for now, except that file disables color formatting
- `ReporterRegistry` is a singleton class which can be overridden if additional reporter types need to be supported.
- reporters handle annotations, file annotations, and groups
- for console/file: annotations just provide standard error message formatting, groups utilize tree-views for output

#### Core functions
The library exposes `formatMessage`, `formatFileMessage`, and `formatGroup` which can have the reporter specified, or will default to whatever is appropriate for the current environment.

#### To-dos
- [ ] Integrate the table formatting. This will involve handling writing MD tables to github summary files, and support for MD as a format in general.
- [ ] Consume the library in tools that produce grouped output (like align-deps)


Saadnajmi:
Any chance this could be shared with Lage?

JasonVMo:
> Any chance this could be shared with Lage?

Yes, totally. This package is a zero-dependency formatting package that can be used anywhere for structured formatting.

Saadnajmi:
Is there an example of output? Hard to judge otherwise

incubator/tools-formatting/src/paths.ts
```diff
@@ -36,3 +37,19 @@ export function shortenPath(path: string, segments = 3): string {
...
+ * Normalize a path for display in messages so that links work correctly in GitHub and Azure DevOpts. This
+ * also makes them readable in normal logs as well as it normalizes them to forward slashes and makes them
+ * relative to the repo root if a root is provided.
+ * @param file The file path to normalize
+ * @param root The root directory to make the file path relative to, will default to process.cwd which may be less correct
+ * @returns The normalized file path
+ */
+export function normalizePath(
+  file: string,
+  root: string = process.cwd()
```
tido64: I think we should only make paths relative if root is set. Otherwise, we should preserve the input.

incubator/tools-formatting/src/reporters.ts
```diff
@@ -0,0 +1,282 @@
...
+   * first and call `super.createReporter(type, options)` for anything the
+   * subclass doesn't handle so the standard built-ins still resolve.
+   *
+   * @param type Reporter name. Built-in values are `"github"`, `"azure"`,
+   *             `"console"`, and `"file"`.
+   * @param options Optional overrides applied to the resulting reporter
+   *             (name, color, ASCII-only).
+   * @throws when `type` is not a recognized name.
+   */
+  createReporter(type: string, options?: ReporterPropOverrides): Reporter {
```
tido64: I think this is a misnomer. It should be `createFormatter` instead. All references to "reporter" should be "formatter".

incubator/tools-formatting/src/index.ts
tido64, line 1: Can we remove all the empty newlines so that oxfmt can sort this file?

incubator/tools-formatting/src/github.ts
```diff
@@ -0,0 +1,110 @@
...
+  return [groupStart, ...children, groupEnd].join("\n");
+}
+
+/**
+ * Escape data for the message portion of a GitHub Actions workflow command.
+ * Per the workflow-command spec, message data only needs `%`, CR, and LF
+ * encoded; `:` and `,` are safe to leave alone.
+ */
+function escapeGitHubData(s: string): string {
+  return s.replace(/%/g, "%25").replace(/\r/g, "%0D").replace(/\n/g, "%0A");
```
tido64: Use `replaceAll` instead of a regex.

incubator/tools-formatting/src/github.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Escape values used in workflow-command property positions (e.g. `file=...`).
+ * Adds `:` and `,` to the data escape set since those characters are the
+ * property delimiters.
+ */
+function escapeGitHubProp(s: string): string {
+  return escapeGitHubData(s).replace(/:/g, "%3A").replace(/,/g, "%2C");
```
tido64: Use `replaceAll` instead of a regex.

incubator/tools-formatting/src/azure.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * Escape Azure DevOps logging-command "data" (the portion after the `]`).
+ * Uses `%AZP25` for `%` so the agent can distinguish encoded text from
+ * user-supplied `%` characters, and percent-encodes CR / LF so multi-line
+ * messages don't break the single-line command format.
+ */
+function escapeAzureData(s: string): string {
+  return s.replace(/%/g, "%AZP25").replace(/\r/g, "%0D").replace(/\n/g, "%0A");
```
tido64: Use `replaceAll` instead of a regex.

incubator/tools-formatting/src/azure.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * Escape Azure DevOps logging-command "property" values (the parts between
+ * `[task.logissue ...]`). In addition to the data escapes, also escapes the
+ * property separator `;` and the command terminator `]`.
+ */
+function escapeAzureProp(s: string): string {
+  return escapeAzureData(s).replace(/;/g, "%3B").replace(/\]/g, "%5D");
```
tido64: Use `replaceAll` instead of a regex.

incubator/tools-formatting/src/azure.ts
```diff
@@ -0,0 +1,118 @@
...
+    return formatAzureFileMessage(severity, fileMessage, base);
+  }
+
+  function formatGroup(header: string, children: string[]): string {
+    const groupStart = `##[group]${escapeAzureData(header)}`;
+    const groupEnd = "##[endgroup]";
+    return [groupStart, ...children, groupEnd].join("\n");
+  }
+
+  return Object.assign(base, { formatMessage, formatFileMessage, formatGroup });
```
tido64: Can we define this object like we do in the GitHub one without `Object.assign`?

incubator/tools-formatting/src/core.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Format grouped output for display using standard formatting rules. For normal console output this will format the
+ * group in a tree-like structure. When running under github or azure this will attempt to use collapsible groups in the UI.
+ * @param header the header to display for the group
+ * @param children the messages to include in the group
+ * @param reporter the reporter to use for formatting, omit to use the default reporter
+ * @returns the formatted group of messages
+ */
+export function formatGroup(
+  header: string,
+  children: string[],
```
tido64, line 48: This function feels unnatural to use. In an app, you want to be able to add the begin/end markers somewhere and everything in-between should be able to call a reporter freely.

In #4156, I've sort of solved this with a `withGroupReporter(header, (reporter) => { ... })` and passing the reporter down. The wrapper just ensures that begin/end markers are printed as necessary.

I'm also not sure any of the CI specifics belong in a formatting package. They are much closer to reporting than formatting. It's a red flag that the code refers to "reporter" more than formatting.

incubator/tools-formatting/src/reporters.ts
```diff
@@ -0,0 +1,282 @@
...
+  function formatFileMessage(
+    severity: Severity,
+    fileMessage: FileMessage
+  ): string {
+    return formatConsoleFileMessage(severity, fileMessage, base);
+  }
+
+  function formatGroup(header: string, children: string[]): string {
+    return formatAsTree(header, children, base);
+  }
```
tido64: This seems wrong. Log lines can belong to a group, but not all of them need to be displayed as a tree.

incubator/tools-formatting/src/reporters.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  return Object.assign(base, { formatMessage, formatFileMessage, formatGroup });
```
tido64: This is unnecessary. We can declare the object like below:
```ts
export function createConsoleOrFileReporter(
  type: Extract<BuiltinReporter, "console" | "file">,
  options?: ReporterPropOverrides
): Reporter {
  const reporter = {
    name: options?.name ?? type,
    noColors: options?.noColors ?? type === "file",
    asciiOnly: Boolean(options?.asciiOnly),
    formatMessage(severity: Severity, message: string): string {
      return formatConsoleMessage(severity, message, reporter);
    },
    formatFileMessage(
      severity: Severity,
      fileMessage: FileMessage
    ): string {
      return formatConsoleFileMessage(severity, fileMessage, reporter);
    },
    formatGroup(header: string, children: string[]): string {
      return formatAsTree(header, children, reporter);
    }
  };
  return reporter;
}
```

incubator/tools-formatting/src/reporters.ts
```diff
@@ -0,0 +1,282 @@
...
+ * @param options Optional overrides for name / color / ASCII behavior.
+ * @returns A reporter that formats output for plain-text streams.
+ */
+export function createConsoleOrFileReporter(
+  type: Extract<BuiltinReporter, "console" | "file">,
+  options?: ReporterPropOverrides
+): Reporter {
+  const name = options?.name ?? type;
+  const noColors = options?.noColors ?? type === "file";
+  const asciiOnly = options?.asciiOnly ?? false;
```
tido64: Avoid `... ?? false`:
```suggestion
  const asciiOnly = Boolean(options?.asciiOnly);
```

incubator/tools-formatting/src/paths.ts
JasonVMo: I had it that way originally, though passing the root is a bit awkward in many cases. In CI tooling for Azure/GitHub you need the right format for paths to behave like links. I'll switch it back.

tido64:

PR #4154, opened 2026-05-16
fix(tools-node): limit package dependency search scope

### LINKED ISSUE #801: Limit the scope of file and directory searching when resolving modules
### What happened?

`findModuleFile`: When combining `searchDir` and `modulePath`, don't let the result "escape" the confines of `searchDir`. If that happens, return `undefined` indicating a module resolution failure.

`findPackageDependencyDir` (local to resolve.ts): Add a `stopDir` parameter to limit the scope of the search to the current repository. Don't let the search go all the way to the root of the file system. Other resolvers don't do this because it violates the algorithm, but it's reasonable for us since we know `node_modules` will only legitimately be in the repo based on the way npm, yarn, pnpm, and other major package managers work (even if node_modules is a bunch of symlinks).

### Affected Package

@rnx-kit/typescript-react-native-resolver

### Version

*

- [X] Android
- [X] iOS
- [X] macOS
- [X] Windows

### System Information

```true
N/A
```

### Steps to Reproduce

N/A

### Code of Conduct

vivekjm:
### Description

Resolves #801.

This adds a `stopDir` option to `findPackageDependencyDir` so callers can keep dependency lookup bounded to a repository root instead of walking all the way up the filesystem. The stop directory itself remains part of the search, so root-level `node_modules` is still found.

A regression test covers both cases: a dependency above `stopDir` is ignored, while a dependency at `stopDir/node_modules` is still resolved.

### Test plan

- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn workspace @rnx-kit/tools-node test`
- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn workspace @rnx-kit/tools-node build`
- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn workspace @rnx-kit/tools-node lint`
- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn change:check`

Note: the default local Node v20.15.0 cannot run the repo test command because it lacks `fs.globSync`; the commands above were run with local Node v22.22.0.

### DIFF
--- a/packages/tools-node/src/package.ts
+++ b/packages/tools-node/src/package.ts
@@ -171,6 +171,11 @@ export type FindPackageDependencyOptions = {
   startDir?: string;
 
   /**
+   * Optional directory where the search should stop.
+   */
+  stopAt?: string;
+
+  /**
    * Optional flag controlling whether symlinks can be found. Defaults to `true`.
    * When `false`, and the package dependency directory is a symlink, it will not
    * be found.
@@ -201,10 +206,12 @@ export function findPackageDependencyDir(
 ): string | undefined {
   const pkgName =
     typeof ref === "string" ? ref : path.join(ref.scope ?? "", ref.name);
+  const startDir = options?.startDir ?? process.cwd();
   const packageDir = findUp(
     path.join("node_modules", pkgName),
     {
-      startDir: options?.startDir,
+      startDir,
+      stopAt: options?.stopAt,
       type: "directory",
       allowSymlinks: options?.allowSymlinks,
     },
--- a/packages/tools-node/src/package.ts
+++ b/packages/tools-node/src/package.ts
@@ -206,11 +206,10 @@ export function findPackageDependencyDir(
 ): string | undefined {
   const pkgName =
     typeof ref === "string" ? ref : path.join(ref.scope ?? "", ref.name);
-  const startDir = options?.startDir ?? process.cwd();
   const packageDir = findUp(
     path.join("node_modules", pkgName),
     {
-      startDir,
+      startDir: options?.startDir,
       stopAt: options?.stopAt,
       type: "directory",
       allowSymlinks: options?.allowSymlinks,
--- a/.changeset/curly-needles-rest.md
+++ b/.changeset/curly-needles-rest.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/tools-node": patch
+---
+
+Add a `stopAt` option to limit `findPackageDependencyDir` searches.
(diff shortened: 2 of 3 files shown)


packages/tools-node/src/package.ts
```diff
@@ -170,6 +170,12 @@ export type FindPackageDependencyOptions = {
    */
   startDir?: string;
 
+  /**
+   * Optional directory where the search should stop. The directory itself is
+   * included in the search.
+   */
+  stopDir?: string;
```
tido64: This should be the path where the search halts if no matches are found before reaching this point. The option should be passed directly to `findUp`.

Maybe rename it to `stopAt` to match the upstream option name.

vivekjm: Updated. renamed the option to `stopAt`, pass it directly through to `findUp`, and adjusted the regression coverage to match `findUp`’s halt semantics.

packages/tools-node/src/package.ts
```diff
@@ -201,10 +206,12 @@ export function findPackageDependencyDir(
 ): string | undefined {
   const pkgName =
     typeof ref === "string" ? ref : path.join(ref.scope ?? "", ref.name);
+  const startDir = options?.startDir ?? process.cwd();
   const packageDir = findUp(
     path.join("node_modules", pkgName),
     {
-      startDir: options?.startDir,
+      startDir,
```
tido64: Can we revert this?

vivekjm:
Updated in 5918878cd. I reverted the incidental `startDir` defaulting change and left the `stopAt` pass-through/regression coverage intact.

Test note: I ran `corepack yarn install --immutable`, then attempted `corepack yarn workspace @rnx-kit/tools-node test`; the local run is blocked because this machine has Node v20.15.0 and the repo test command calls `fs.globSync`, which is unavailable there.

tido64:
Thanks! It looks like CI is failing on formatting. Can you run `yarn format` and push the changes?

> Test note: I ran `corepack yarn install --immutable`, then attempted `corepack yarn workspace @rnx-kit/tools-node test`; the local run is blocked because this machine has Node v20.15.0 and the repo test command calls `fs.globSync`, which is unavailable there.

Is there a particular reason for why you can't use Node v22 or greater?

vivekjm:
Thanks. I ran `yarn format` and pushed the formatting-only follow-up in `4d71149b`.

I also reran the package test with Node 22.22.0 on PATH:

- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" corepack yarn workspace @rnx-kit/tools-node test`

That passed: 36 tests, 3 suites. The earlier local failure was because my default shell was still on Node 20.15.0, where the repo's test runner hits `fs.globSync` before it gets to the package tests.

tido64 (approved):
Thanks!

PR #4168, opened 2026-05-20
feat(tools-git): introduce a package for common Git operations

tido64:
### Description

Not sure how useful this is. We copied these functions from [`react-native-test-app`](https://github.com/microsoft/react-native-test-app/blob/trunk/scripts/affected.mts). At the very least, this will allow us to reuse the same code.

### Test plan

n/a

### DIFF
--- a/incubator/tools-git/package.json
+++ b/incubator/tools-git/package.json
@@ -1,0 +1,49 @@
+{
+  "name": "@rnx-kit/tools-git",
+  "version": "0.0.1",
+  "description": "A collection of functions for common Git operations",
+  "homepage": "https://github.com/microsoft/rnx-kit/tree/main/incubator/tools-git#readme",
+  "license": "MIT",
+  "author": {
+    "name": "Microsoft Open Source",
+    "email": "microsoftopensource@users.noreply.github.com"
+  },
+  "repository": {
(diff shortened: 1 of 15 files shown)


incubator/tools-git/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "lint": "rnx-kit-scripts lint",
+    "test": "rnx-kit-scripts test",
+    "update-readme": "rnx-kit-scripts update-api-readme"
+  },
+  "devDependencies": {
+    "@rnx-kit/scripts": "*",
+    "@rnx-kit/tsconfig": "*"
+  },
+  "engines": {
+    "node": ">=20.18"
```
JasonVMo, line 43: Why not node 22+?

JasonVMo (approved):
Looks good from what I can tell. You may want to look at what Vladimir added in the fork-sync package recently as that code could be moved to this package as well. One comment on the node requirement but that is just something to consider.

incubator/tools-git/package.json
tido64: It's the current default in the repo and we don't use anything that requires 22+. The only new feature I can think of in 22 is glob, but we don't use that here.

PR #4182, opened 2026-05-30
fix(align-deps): add React Native type profiles

### LINKED ISSUE #269: align-deps profiles: add a few more dependencies to the profiles
Just realised, we don't have in the various profiles the @types/ packages defined (ex. @types/react-native, etc) to match. We should add those and also add "react-test-renderer", since it follows the same rules as react/react-dom.

vivekjm:
## Summary

- add a `types/react-native` align-deps capability for React Native 0.61 through 0.70
- make React Native core profiles resolve `@types/react-native` as a dev-only dependency for those versions
- stop resolving the community type package starting with the 0.71 profile, where React Native ships its own TypeScript declarations
- update capability, manifest, set-version, and snapshot coverage

Fixes #269.

## Test Plan

- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn workspace @rnx-kit/align-deps test test/capabilities.test.ts`
- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn workspace @rnx-kit/align-deps test test/manifest.test.ts`
- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn workspace @rnx-kit/align-deps test`
- `PATH="$HOME/.nvm/versions/node/v22.22.0/bin:$PATH" yarn workspace @rnx-kit/align-deps build`

Note: `yarn change:check --since origin/main` could not run in this checkout because the local Yarn dependency map is missing `oxlint@npm:1.61.0`; I added the required Changesets entry manually.


JasonVMo (approved):
One note is whether we should add @types/react support as well.

tido64:
I don't think we've decided whether having types in the profiles is a thing we want to maintain. For one, it means that we'll have to provide types for all packages, not just React and React Native. Second, do we want to put types behind a flag for users who don't want them? For now, I'm going to close this PR until we figure out how we want to approach this.

PR #4230, opened 2026-06-24
chore: bump TypeScript to 7.0 RC

tido64:
### Description

Bumps TypeScript to 7.0 RC.

TypeDoc does not support 7.0 yet so I had to create a wrapper for it using 6.0.

### Test plan

CI should pass.


tido64:
~~It looks like we'll have to wait for TypeDoc to support 7.0~~ Created a wrapper with TypeScript 6.0

JasonVMo (approved):

PR #4295, opened 2026-08-12
fix(cli): load modules only on demand

tido64:
### Description

Improve `@rnx-kit/cli` start-up performance

### Test plan

n/a


JasonVMo (approved):
Nice, this also avoids issues like when a platform CLI makes calls in initialization that aren't valid on a different platform.

PR #4298, opened 2026-08-14
fix(metro-plugin-cyclic-dependencies-detector): improve stack tracing

tido64:
### Description

Improved performance on deep dependency graphs

### Test plan

n/a

### DIFF
--- a/.changeset/icy-things-share.md
+++ b/.changeset/icy-things-share.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/metro-plugin-cyclic-dependencies-detector": patch
+---
+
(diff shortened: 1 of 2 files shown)


JasonVMo (approved):

JasonVMo:
One thing to validate is whether the module could have appeared in the stack multiple times. The set semantics and array.pop semantics are different.

tido64:
> One thing to validate is whether the module could have appeared in the stack multiple times. The set semantics and array.pop semantics are different.

We call `.has()` before we `.add()`. If it appears twice in the stack, we've detected a cycle and return. This shouldn't be possible.

PR #4299, opened 2026-08-14
fix(metro-plugin-duplicates-checker): improve banned/ignored checks

tido64:
### Description

Improved banned/ignored packages checks

### Test plan

n/a


JasonVMo (approved):
Yikes! Good improvement as I'd have to assume that this is super high frequency.

PR #4302, opened 2026-08-17
fix(align-deps): cache read manifests

tido64:
### Description

Cache read package manifests to avoid re-reading and parsing the same `package.json` files while iterating over dependencies.

Number of packages read in rnx-kit:

- **Before:** 931
- **After:** 503

### Test plan

n/a

### DIFF
--- a/packages/align-deps/src/cache.ts
+++ b/packages/align-deps/src/cache.ts
@@ -1,0 +1,45 @@
+const MAX_CACHE_ENTRIES = 10_000;
+
+export class BoundedCache<K, V> {
+  private map: Map<K, V>;
+  private maxEntries: number;
+
+  constructor(maxEntries = MAX_CACHE_ENTRIES) {
+    this.map = new Map<K, V>();
+    this.maxEntries = maxEntries;
+  }
+
+  get(key: K): V | undefined {
+    const value = this.map.get(key);
+    if (value !== undefined) {
+      this.update(key, value);
+    }
+    return value;
+  }
+
+  has(key: K): boolean {
+    return this.map.has(key);
+  }
+
+  set(key: K, value: V): void {
+    this.update(key, value);
+    if (this.map.size > this.maxEntries) {
+      const lruKey = this.leastRecentlyUsed();
+      if (lruKey !== undefined) {
+        this.map.delete(lruKey);
+      }
+    }
+  }
+
+  private leastRecentlyUsed(): K | undefined {
+    // Least recently used items bubble towards the front
+    return this.map.keys().next().value;
+  }
+
+  private update(key: K, value: V): void {
+    // By deleting and re-inserting the value, we ensure the entry remains at
+    // the end and treated as most recently used
+    this.map.delete(key);
+    this.map.set(key, value);
+  }
+}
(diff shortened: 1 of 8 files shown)


packages/align-deps/src/cache.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  private leastRecentlyUsed(): K | undefined {
+    // Least recently used items bubble towards the front
+    return this.map.keys().next().value;
+  }
+
+  private update(key: K, value: V): void {
+    // By deleting and re-inserting the value, we ensure the entry remains at
+    // the end and treated as most recently used
+    this.map.delete(key);
```
JasonVMo, line 42: Isn't this pretty inefficient? It effectively mutates memory on get, which is likely minor if the cost of loading is big but in many use cases may thrash memory.

packages/align-deps/src/dependencies.ts
```diff
@@ -26,6 +29,17 @@ function isDevOnlyCapability(
   return profiles.some((profile) => profile[capability]?.devOnly);
 }
 
+function readManifestFromDir(packageDir: string): PackageManifest {
+  const cached = manifestCache.get(packageDir);
+  if (cached) {
+    return cached;
+  }
+
+  const manifest = readPackage(packageDir);
```
JasonVMo, line 38: One thought for package.json files. If we use require("<filepath>.json") it will be in the module cache already. It may be worth investigating this.

tido64: The issue with `require()` the `package.json` is that it may fail if there's an exports map and `./package.json` isn't specified. We want to make sure we can migrate to ESM soon to avoid the ESM → CJS lexer overhead.

packages/align-deps/src/cache.ts
tido64: I think it's very minor. I would assume that map allocates some memory upfront to avoid allocating/deallocating with every insert. At least it doesn't show up during profiling.

packages/align-deps/src/dependencies.ts
JasonVMo: Makes sense

packages/align-deps/src/cache.ts
JasonVMo: I'm less familiar with the internal align-deps workflows but do we need to cache the full package.json?
- is a sentinal cache, effectively marking resolved paths as 'done' sufficient?
- would it be better to cache processed results? i.e. cache the resolved and flattened transitive graph for each direct workspace package dependency?

tido64: We can't just mark a resolved path as "done" because of diamond dependencies. But I think it should be sufficient to just cache the dependencies and `KitConfig`. It's smaller and we probably won't need `BoundedCache`.

tido64: Implemented in a separate PR to avoid merge conflicts: https://github.com/microsoft/rnx-kit/pull/4309

It's stacked on top of this one so once all PRs are approved, they'll land together in `main`.

JasonVMo (approved):

PR #4314, opened 2026-08-23
fix(tools-typescript): don't let BatchWriter.finish() wait forever

giaBaoJS:
`writeFinished()` emits `done` the moment the last queued write settles. If nothing is listening at that point the event is dropped, and a later `finish()` starts listening for an event that will never come again. Any caller that writes some files, awaits unrelated work, and then calls `finish()` without queueing anything new hangs there.

`finish()` only guarded against the empty queue, so it re-emitted `done` when nothing had been written but not when everything had. Compare `queued` against `written` instead, which covers both cases: `written` only reaches `queued` once every write has settled.

### DIFF
--- a/.changeset/quiet-donkeys-repeat.md
+++ b/.changeset/quiet-donkeys-repeat.md
@@ -1,0 +1,5 @@
+---
+"@rnx-kit/tools-typescript": patch
+---
+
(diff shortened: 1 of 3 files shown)


tido64 (approved):
Thanks ❤️

PR #4331, opened 2026-08-26
test(third-party-notices): add integration tests

### LINKED ISSUE #283: third-party-notices: Add full integration test
The code is unittested, but does not have a full integration test.

Reason at the time was that an integrioatn test requires a repo and either pulling in packages from npm during the test and running webpack before running this logic... Or we have to check in a webpack file but that would then not be a 'live' integration test in case webpack gets updated...

THis item track adding the integration test

tido64:
### Description

Resolves #283.

### Test plan

n/a


JasonVMo (approved):
I'm not a fan of manually passing the fs through everywhere in order to mock it as it is so easy to have a gap in the flow. That being said I'm not aware of a cleaner way of doing so.

PR #4335, opened 2026-08-30
fix(align-deps): handle whitespace in version ranges

giaBaoJS:
### Description

`dropPatchFromVersion()` visits each comparator in a range by splitting the
range on `" "`, then joining the parts back together. A range like
`>= 0.72.0` is valid semver (`semver.validRange(">= 0.72.0")` returns
`">=0.72.0"`, and `dropPatchFromVersion()` itself accepts it via the
`semverValidRange` guard on line 28), but splitting it on whitespace yields
`[">=", "0.72.0"]`. The comparator is then treated as a version of its own,
and the function returns `">=.0 0.72"`, which is not a valid range. Runs of
more than one space hit the same path through the resulting empty token.

| input | before | after |
| :-- | :-- | :-- |
| `>= 0.72.0` | `>=.0 0.72` | `>=0.72` |
| `^ 1.2.3` | `^.0 1.2` | `^1.2` |
| `> 0.68.0 <= 0.70.0` | `>.0 0.68 <=.0 0.70` | `>0.68 <=0.70` |
| `>0.68.0  <=0.70.0` | `>0.68 .0 <=0.70` | `>0.68 <=0.70` |

This is reachable from `align-deps --init`. `initializeConfig()` passes the
`react-native` version straight from the manifest into
`dropPatchFromVersion()`, and the result goes to `filterPreset()`, which
rejects the mangled range:

```
Error: Invalid version range in requirement: react-native@>=.0 0.64
    at parseRequirements (packages/align-deps/src/preset.ts:57:13)
    at filterPreset (packages/align-deps/src/preset.ts:74:16)
    at initializeConfig (packages/align-deps/src/commands/initialize.ts:64:18)
```

So a package declaring `"react-native": ">= 0.64.1"` cannot be initialized at
all.

The fix strips whitespace between a comparator and its version before
splitting, since `>= 1.0.0` and `>=1.0.0` are the same range, and splits on
`/\s+/` so repeated whitespace no longer produces empty tokens. Hyphen ranges
such as `1.0 - 2.0` are unaffected because `-` is not a comparator.

### Test plan

- `yarn test` in `packages/align-deps`: 195 passed, 0 failed (189 passed
  before this change; the 6 new tests are the 5 added cases in
  `dropPatchFromVersion()` plus one in `initializeConfig()`).
- Reverting only `src/helpers.ts` while keeping the tests turns all 6 red.
  The `initializeConfig()` case fails with the `Invalid version range in
  requirement: react-native@>=.0 0.64` error quoted above, and the
  `dropPatchFromVersion()` cases fail on the mangled strings in the table.
- `yarn lint` and `yarn format` are clean, as are `yarn knip`,
  `yarn constraints`, `yarn dedupe --check` and `node scripts/lint-tsconfig.ts`
  at the repo root.
- `packages/cli` is the only affected downstream package; its 66 tests still
  pass.


tido64 (approved):
Thanks!

PR #4368, opened 2026-09-15
fix(tools-workspaces): read pnpm-workspace.yaml before parsing it

giaBaoJS:
### Description

`@rnx-kit/tools-workspaces` returns zero packages for every pnpm workspace.

Minimal repro, from a pnpm monorepo root with `packages: ['packages/*']` and
five packages under `packages/`:

```js
import { findWorkspacePackages } from "@rnx-kit/tools-workspaces";
console.log(await findWorkspacePackages()); // []
```

`findWorkspaceRoot()` still returns the right directory, so the workspace is
detected; only its packages go missing.

Why it happens: `js-yaml`'s `load()` takes YAML *content*, but
[`packages/tools-workspaces/src/pnpm.ts#L13`](https://github.com/microsoft/rnx-kit/blob/1658f64/packages/tools-workspaces/src/pnpm.ts#L13)
and
[`#L25`](https://github.com/microsoft/rnx-kit/blob/1658f64/packages/tools-workspaces/src/pnpm.ts#L25)
hand it the *path* to `pnpm-workspace.yaml`. That path is valid YAML: it parses
to the plain scalar string `"/path/to/pnpm-workspace.yaml"`, so `packages` is
`undefined`, `findPackages` hits its `Array.isArray` fallback and returns `[]`.
The file read was lost in #4275 when `read-yaml-file` (which read and parsed)
was swapped for `js-yaml` (which only parses).

This change reads the file and passes its contents to `yaml.load()`, restoring
the pre-#4275 behaviour.

Downstream effect, since the value is consumed in three places:

- `align-deps --workspace` (`packages/align-deps/src/cli.ts#L157`) checks no
  packages at all in a pnpm repo.
- `WorkspacesInfo.isWorkspace()` returns `false` for every package, because
  `getPackageFilters()` is `undefined` and the fallback package list is empty.
  `@rnx-kit/tools-packages` uses this to decide whether a package is a
  workspace, and `@rnx-kit/lint-lockfile` uses it to pick rules.

### Test plan

`yarn test` in `packages/tools-workspaces`: 44 passing before, 46 after.

The two existing pnpm tests could not catch this. They iterate
`for (let i = 0; i < result.length; ++i)`, so an empty result passes
vacuously. This adds a length assertion to both, which fails on `main`
(`0 !== 5`) and passes with the fix, plus two tests for a
`pnpm-workspace.yaml` that carries settings but no `packages` key (valid
since pnpm v10).

Verified by reverting only `src/pnpm.ts` with the new tests in place: both
length assertions go red with `0 !== 5`. Three mutants were also checked, each
caught: dropping the read in the async path only, dropping it in the sync path
only, and globbing from the parent of the workspace root.

Consumers re-run against the change, all green: `align-deps` 243, `metro-config`
25, `lint-lockfile` 10, `tools-packages` 2. `yarn lint`, `yarn build` and
`yarn format` are clean in `packages/tools-workspaces`.

### Out of scope

A `pnpm-workspace.yaml` that is empty or contains only comments still throws,
now as `YAMLException: expected a document, but the input is empty` from
js-yaml v5. That is not a regression from #4275: with `read-yaml-file` and
js-yaml v4 the same input returned `undefined` and threw a `TypeError` on
destructuring. Handling it needs a decision about what an empty workspace
config should mean, so I left it alone.

The other two `yaml.load()` call sites in the repo,
`packages/align-deps/src/commands/exportCatalogs.ts#L54` and
`incubator/lint-lockfile/src/yarn/modern.ts#L20`, already pass file contents
and are unaffected.

The vacuous `for (let i = 0; i < result.length; ++i)` loop is used in the yarn,
lerna, rush and bun tests too. Those code paths work, so I only added the
length assertion to the pnpm tests rather than touching unrelated files.


tido64 (commented):
Thanks for the fix. Can you also add tests that do have a `packages` field?

giaBaoJS:
Added in d5499f3.

New fixture `test/__fixtures__/pnpm-with-settings` has a `packages` field next to
another setting, and excludes one package with a negated pattern:

```yaml
onlyBuiltDependencies:
  - esbuild
packages:
  - 'packages/*'
  - '!packages/t-800'
```

There are five package directories under it, so the expected result is four.
Two tests run it, one through `findWorkspacePackages` and one through
`findWorkspacePackagesSync`, which is the path that goes via
`getPackageFilters`. `yarn test` in `packages/tools-workspaces` is 46 to 48.

Counterfactual: reverting only `src/pnpm.ts` to the pre-fix version makes both
new tests fail with `actual: 0, expected: 4`, alongside the two existing pnpm
tests failing with `actual: 0, expected: 5`.

I also checked the tests notice a subtly wrong fix, not just a missing one.
Four mutants, each caught:

1. read dropped in `findWorkspacePackages` only: async tests red, sync green.
2. read dropped in `getPackageFilters` only: sync tests red, async green.
3. `getPackageFilters` filtering out `!` patterns: sync settings test red,
   `actual: 5, expected: 4`.
4. same filter applied in `findWorkspacePackages`: async settings test red,
   `actual: 5, expected: 4`.

Consumers re-run and still green: align-deps 243, metro-config 25,
lint-lockfile 10, tools-packages 2. Lint, build and format are clean.

No changeset added for this, since the existing one already covers the fix.

tido64 (approved):
Thanks again

PR #4370, opened 2026-09-16
feat(align-deps): add `--why`

### LINKED ISSUE #2454: align-deps: add a toggle to show why a dependency got added
This command/flag is modeled after [`yarn why`](https://yarnpkg.com/cli/why). Example usage:

```sh
% rnx-align-deps --why @react-native-community/netinfo
├─ some-package
│  └─ @react-native-community/netinfo (via 'storage')
│
└─ another-package
   └─ @react-native-community/netinfo (via 'storage')
```

`align-deps` scans the current package (either in the current directory, or the directory specified by the user) for dependencies that require the specified package, including the corresponding capability.

`--why` is its own command, on the same line with `--init`, `--export-catalogs`, `--set-version`, and is therefore directly incompatible with them. Similarly, it does not invoke a check.

tido64:
### Description

Added `--why` for displaying why a package was included

Resolves #2454

### Test plan

```
% cd packages/align-deps
% yarn bundle
% node lib/index.js --why react-native ../test-app/
└─ @rnx-kit/test-app
   └─ react-native (via 'core-android', 'core-ios', 'core/testing')
```


packages/align-deps/src/commands/why.ts
```diff
@@ -0,0 +1,136 @@
...
+      const entries = reasons.entries();
+      for (let i = 0; i < size; ++i) {
+        const { value } = entries.next();
+        if (!value) {
+          continue;
+        }
+
+        const [module, capabilities] = value;
+        const last = i === size - 1;
+        console.log(`${last ? "└" : "├"}─ ${module}`);
```
JasonVMo, line 124: This could use the tree reporter in tools-formatting.

packages/align-deps/src/commands/why.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export function makeWhyCommand(name: string, options: Options): Command {
+  return (manifestPath: string) => {
+    const inputConfig = loadConfig(manifestPath, options);
+    if (isError(inputConfig)) {
+      return inputConfig;
+    }
+
+    // Transform legacy config in memory only, even with --migrate-config.
```
JasonVMo, line 75: It feels like most of the code in here could be consolidated as I think it is duplicating code that exists elsewhere in align-deps. Maybe create a wrapper class around a package with load on demand getters for config, capabilities, and so on?

packages/align-deps/src/commands/why.ts
tido64: It's two lines though. I think we're fine with not pulling in another library.

packages/align-deps/src/commands/why.ts
tido64: Besides the config loading/migration logic, most of this code is actually different. Deduplicating means that I will have to make things more complex (less maintainable). I'm thinking of removing the legacy config support next so this will also disappear.

JasonVMo (approved):

PR #4375, opened 2026-09-22
feat(tools-formatting): Add text metrics parsing for terminal output

JasonVMo:
### Description

This adds support for determining the visible width and line-count of strings that will be sent ot the terminal, this will be incorporated into changes to table formatting as a next step. There are two primary routines added: `getTextMetrics` and `getTextOutput`.

#### `getTextMetrics(text?: string | string[])`

This is the main driver for figuring out the visible max line width and line count for terminal output.
- this does not call `vtstripControlCharacters` internally which would do a full string copy
- it handles ansi control characters for colors and formatting
- it handles emoji, including with color sequences
- it handles wide characters like East Asian characters which often take up 2 columns.
- it ignores control characters that aren't rendered
- it handles CRLF and LF line endings correctly
- Note that node.js has much of this internally but it is not exposed

#### `getTextOutput(text?: string | string[])`

This gathers the text metrics and breaks any strings with CRLF/LF sequences and returns an array of lines.
- this handles strings, or arrays of strings. The arrays of strings can themselves have carriage returns within

### Test plan

Extensive test cases added


incubator/tools-formatting/src/const.ts
```diff
@@ -33,3 +33,62 @@ export const TABLE_STYLES: Record<StyleKeys, TableViewParts> = {
     fill: "-",
   },
 };
+
+// Wide/fullwidth ranges not already covered by zero-width or emoji handling:
+// https://www.unicode.org/Public/16.0.0/ucd/EastAsianWidth.txt
+export const WIDE_CHAR_RANGES: readonly (readonly [number, number])[] = [
```
tido64, line 39: Will we need to update this list when characters are added/removed from Unicode?

tido64 (approved):

incubator/tools-formatting/src/const.ts
JasonVMo: Potentially. Though the user-bad is really that columns and output won't line up correctly so it isn't catastrophic. It is also pretty well-standardized information so it is easy to update if needed.

